PR #7, opened 2021-02-02
Add a test for `Dir::read`; run `cargo test` on ubuntu and macos

pchickey:


sunfishcode (approved):
Reading directories is "interesting" enough that this is a good test to have. Also, thanks for adding the `cargo test` rule to CI.

PR #12, opened 2021-06-12
Support OpenBSD.

cfallin:
This basically adds OpenBSD alongside NetBSD where it appears in a target-specific configuration directive, as suggested by @sunfishcode. Verified to pass tests on OpenBSD 6.9 / x86_64. A++ easy port, would do again!

### DIFF
--- a/src/fs/makedev.rs
+++ b/src/fs/makedev.rs
@@ -1,5 +1,10 @@
 /// `makedev(maj, min)`
-#[cfg(not(any(target_os = "android", target_os = "emscripten", target_os = "wasi")))]
+#[cfg(not(any(
+    target_os = "android",
+    target_os = "emscripten",
+    target_os = "wasi",
+    target_os = "openbsd"
+)))]
 #[inline]
 pub fn makedev(maj: u32, min: u32) -> u64 {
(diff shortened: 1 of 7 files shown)


src/fs/makedev.rs
```diff
@@ -1,5 +1,10 @@
 /// `makedev(maj, min)`
-#[cfg(not(any(target_os = "android", target_os = "emscripten", target_os = "wasi")))]
+#[cfg(not(any(
+    target_os = "android",
+    target_os = "emscripten",
+    target_os = "wasi",
+    target_os = "openbsd"
+)))]
```
sunfishcode, line 7: The existing code admittedly isn't entirely consistent about this, but since the entire `makedev.rs` module is not implemented for openbsd, I'd prefer to put the conditional on the `mod makedev;` line in src/fs/mod.rs, instead of adding conditionals to everything it defines.

sunfishcode: Ah, you already added the conditions to mod.rs. I have a big PR which conflicts with this one, so I've now taken the liberty of just fixing this, by reverting the changes to makedev.rs, so that I can land this and sort out the merge conflicts on my side.

PR #18, opened 2021-07-14
Add support for using `dup2` to replace stdin/stdout file descriptors.

sunfishcode:
One of the common use cases for `dup2` is to use it to replace stdin and
stdout with new file descriptors. Add support for this, and add a test
that does this.

This changes the signature of `dup2` to take a `BorrowedFd` to replace
rather than taking an `IntoFd` and returning an `OwnedFd`. This means
that it's no longer necessary to own the "new" fd value to replace it
with a dup of another fd, which may be more than users may be expecting
to deal with when they implement `AsFd`.

It's not yet clear that this is the right way to go.


sunfishcode:
It feels like letting users dup2 unrelated file descriptors over BorrowedFd is too much. Closing this in favor of #19.

PR #28, opened 2021-08-22
Disable the std feature of itoa

bjorn3:
This is a tiny step towards no_std compatibility.

### DIFF
--- a/src/path/dec_int.rs
+++ b/src/path/dec_int.rs
@@ -1,6 +1,6 @@
 use crate::io::AsRawFd;
 use io_lifetimes::AsFd;
-use itoa::{write, Integer};
+use itoa::{fmt, Integer};
 use std::ffi::OsStr;
 use std::ops::Deref;
 #[cfg(unix)]
@@ -38,7 +38,7 @@ impl DecInt {
             buf: [0; 20],
             len: 0,
         };
-        me.len = write(&mut me.buf[..], i).unwrap();
+        fmt(&mut me, i).unwrap();
         me
     }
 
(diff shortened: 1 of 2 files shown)


src/path/dec_int.rs
```diff
@@ -55,6 +55,17 @@ impl DecInt {
     }
 }
 
+impl core::fmt::Write for DecInt {
+    fn write_str(&mut self, s: &str) -> core::fmt::Result {
```
sunfishcode, line 59: Could you add `#[inline]` here so that this code can fully inline if called from another crate?

src/path/dec_int.rs
```diff
@@ -55,6 +55,17 @@ impl DecInt {
...
+        if self.len + s.len() > 20 {
+            return Err(core::fmt::Error);
+        }
+        self.buf[self.len..self.len + s.len()].copy_from_slice(s.as_bytes());
```
sunfishcode, line 63: Instead of manual bounds-checking, would it work to call [`self.buf.get[self.len..self.len+s.len()]`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get), and do a `match` on it to handle the error?

src/path/dec_int.rs
```diff
@@ -55,6 +55,17 @@ impl DecInt {
     }
 }
 
+impl core::fmt::Write for DecInt {
```
sunfishcode, line 58: If I understand correctly, this impl is visible in the public API; it's untidy to have public things we don't want users to use, and we'd have to think about whether `self.len + s.len()` could overflow. Would it work to declare a new private type which implements `fmt::Write`, and is wrapped by `DecInt`?

bjorn3: Done

PR #35, opened 2021-08-31
Some more OpenBSD fixes; now passes tests on OpenBSD 7.0-current.

cfallin:
This fixes up some last tidbits that are needed to compile and pass `cargo test` on OpenBSD 7.0-current with Rust 1.54.0.

Thanks for the work on the `rsix` end of this! I don't expect this to avoid any breakage in the future, unless there's a way to run OpenBSD on CI (there is https://github.com/cross-platform-actions/action which uses the nested-virt-on-macOS-runner trick I mentioned today but that is... not production-grade). But at least it will have been known to work at this one point :-)

### DIFF
--- a/src/imp/libc/fs/dir.rs
+++ b/src/imp/libc/fs/dir.rs
@@ -10,7 +10,14 @@ use io_lifetimes::{AsFd, BorrowedFd, IntoFd};
     target_os = "linux",
     target_os = "openbsd",
 )))]
-use libc::{dirent as libc_dirent, readdir as libc_readdir};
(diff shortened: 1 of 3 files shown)


sunfishcode:
Thanks!

A cross-platform-actions CI looks like more work than I myself am prepared for here, though I'd happily accept a PR if anyone wanted to set this up.

I did find a way that we can at least test `cargo check`, using build-std, which ought to catch the majority of the regressions. The PR is https://github.com/bytecodealliance/rsix/pull/36

PR #36, opened 2021-08-31
Add a CI rule to cross-check on x86_64-unknown-openbsd.

sunfishcode:


sunfishcode:
OpenBSD is Tier 3 and not supported in rustup, but we may be able to do a `cargo check` using build-std without too much fiddling here, which should hopefully catch a lot of the easy regressions.

PR #47, opened 2021-09-03
[WIP] Implement `mlockall` and `munlockall`

vrmiguel:


src/imp/libc/syscalls.rs
```diff
@@ -1331,6 +1332,17 @@ pub(crate) unsafe fn mlock_with(
     ret(libc::mlock(addr, length))
 }
 
+/// # Safety
+///
+/// MT-Safe | AS-Safe | AC-Safe
```
sunfishcode, line 1337: These are interesting :-).

rsix doesn't currently include these POSIX safety markers for its functions. Here's my rough mental model of what Rust programmers should assume:
 - All functions are MT-Safe unless they're marked `unsafe` *and* they specifically document thread-unsafety in their `Safety` comment. And I don't think there's anything rsix will want to expose that's MT-Unsafe.
 - No functions are AS-Safe or AC-Safe. Rust doesn't consider AS-Unsafe or AC-Unsafe to be `unsafe`, and in general Rust code doesn't interact with these concepts.

That said, if anyone has use cases that need AS-Safe or AC-Safe, I'd be interested to hear about them and think about what rsix should do in those situations.

sunfishcode:
Hi @vrmiguel, are you still working on this?

sunfishcode:
Closing for now; please feel free to re-open!

PR #53, opened 2021-09-11
Fix build errors on s390x

uweigand:
- Disable MAP_SYNC until https://github.com/rust-lang/libc/pull/2395

- Handle non-standard statfs member types on s390x.


sunfishcode:
Thanks!

PR #54, opened 2021-09-12
Implement `linux_raw` ARM support

Urgau:
This pull-request implement `linux_raw` ARM support with both inline and outline asm and add it as `arm-linux` in GHA with `armv7-unknown-linux-gnueabihf` for the rust target.

The outline assembly use the same structure as the `riscv64` with one main `rsix_reordered_syscall` and intermediate functions that put `nr` at the last place.

The inline and outline assembly works both with ARMv5 onwards. I can make the outline assembly works with ARMv4T if wanted but I didn't do it because the inline assembly generated by Rust have ARMv5 as minimum.

This PR also adjust `pread64`, `pwrite64` and `ftruncate64` because they have an extra `u32` padding after the file descriptor.


src/imp/linux_raw/arch/outline/arm.S
```diff
@@ -0,0 +1,92 @@
...
+    .p2align    4
+rsix_reordered_syscall:
+rsix_reordered_syscall0:
+rsix_reordered_syscall1:
+rsix_reordered_syscall2:
+rsix_reordered_syscall3:
+rsix_reordered_syscall4:
+rsix_reordered_syscall5:
+rsix_reordered_syscall6:
+    .cfi_startproc
```
sunfishcode, line 18: It looks like `clang` uses `.fnstart` and `.fnend` instead of `.cfi_startproc`/`.cfi_endproc` on ARM. ([AS docs](https://sourceware.org/binutils/docs/as/ARM-Directives.html#ARM-Directives).) Do you know if that'd be appropriate here?

It also adds `.cantunwind` to simple functions; syscalls shouldn't unwind, so should we add that here too?

sunfishcode (commented):
Wow, this is really cool! I'm curious how you found the process of adding a new architecture. Were the reoredered_syscall parts confusing? Or other parts? :sweat_smile: 

Could you also add `arm` as supported the top-level README.md?

src/imp/linux_raw/vdso_wrappers.rs
```diff
@@ -352,6 +352,8 @@ fn init() {
     if let Some(vdso) = vdso::Vdso::new() {
         #[cfg(target_arch = "x86_64")]
         let ptr = vdso.sym(cstr!("LINUX_2.6"), cstr!("__vdso_clock_gettime"));
+        #[cfg(target_arch = "arm")]
+        let ptr = vdso.sym(cstr!("LINUX_2.6.39"), cstr!("__vdso_clock_gettime"));
```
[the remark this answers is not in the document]
Urgau: No, you're right that a mistake on my side/

src/imp/linux_raw/arch/outline/arm.S
```diff
@@ -0,0 +1,92 @@
...
+    .size rsix_reordered_syscall, .-rsix_reordered_syscall
+
+    .section    .text.rsix_reordered_syscall1_noreturn,"ax",%progbits
+    .p2align    4
+rsix_reordered_syscall1_noreturn:
+    .cfi_startproc
+    push    {r7, lr}
+    ldr     r7, [sp, #28]
+    svc     #0
+    .inst   0xe7ffdefe
```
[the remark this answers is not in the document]
Urgau: As far as I understand there is "trap" mnemonic for ARM asm. I added a comment and changed the value to be a "trap" equivalent instead of a debugger breakpoint as I wrote it originally.

src/imp/linux_raw/arch/outline/arm.S
Urgau: I used the same structure as the other architecture, so didn't pay to much attention about `.cfi_startproc` and `.cfi_endproc` but it indeed seems that it will be better to use `.fnstart` and `.fnend` (I have changed this). GLibc also use [`.fnstart`](https://github.com/bminor/glibc/blob/abd383584b16dd0fb1bbf40e4ece65ebe7b839ec/sysdeps/unix/sysv/linux/arm/sysdep.h#L226) and [`.fnend`](https://github.com/bminor/glibc/blob/abd383584b16dd0fb1bbf40e4ece65ebe7b839ec/sysdeps/unix/sysv/linux/arm/sysdep.h#L309).

As for `.cantunwind` I also added it because as you said syscalls shouldn't (can't ?) unwind. GLibc doens't use it for `syscall` but does for some other (a bit weird).

Urgau:
I've updated the PR with your comments and I also switch from "armv7" to "armv5te" to better reflect/test the minimum supported  ARM version.

> Wow, this is really cool! I'm curious how you found the process of adding a new architecture. Were the reoredered_syscall parts confusing? Or other parts? sweat_smile

I proceeded in a logical way; I first enable the `linux_raw` backend in the Cargo.toml, then I temporarily disable  the check for "arm.S" in the `build.rs`, then I fixed every compiler errors until getting to the inline assembly where I've copied from another architecture and update it accordingly (I don't remember exactly how I found that I needed to use `scv 0` and `r7` for the syscall number); after that I updated the `main.yaml` with the new architecture so I wouldn't need to setup every archs on my pc; then got a headache because only `pread64` and `pwrite64` tests fails (it took my a while to finally figure out the correct definition for arm; the linux kernel source code is not very clear on that end); and then I finally implemented the outline assembly using the `musl` source code (didn't help me) and godbolt with the rust `rsix_reordered_syscall6` to get the proper assembly.

At first the `reoredered_syscall` functions where a little bit weird but looking at their definitions was enough.

> Could you also add `arm` as supported the top-level README.md?

Done.

sunfishcode:
Thanks!

src/imp/linux_raw/arch/outline/arm.rs
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// Then we define inline wrapper functions that do the reordering.
+mod reorder {
+    use super::*;
+
+    #[inline]
+    #[must_use]
+    pub(in crate::imp::linux_raw) unsafe fn syscall0_readonly(nr: SyscallNumber) -> RetReg<R0> {
+        let u = std::mem::MaybeUninit::uninit().assume_init();
```
bjorn3, line 154: This is instant UB. `std::mem::uninitialized()` was depracated in favor of `MaybeUninit` because it was equivalent to `MaybeUninit::uninit().assume_init()`, which is UB for almost all return types. `MaybeUninit` being one of the only exceptions.

Urgau: Yah, this is UB is a kinda what we want. All the `u1`, `u2`, ... arguments are just here so that `nr` (syscall number) can always be at the 8th place but we don't care about what values are in those values. We could make them take `0` but that would add another instruction to set it, but because we don't care I'm not sure that worth it.

bjorn3: No, you want an uninit integer, which we aren't sure yet is allowed at all. Even if it is allowed, calling `.assume_uninit()` is UB when you haven't initialized the `MaybeUninit` yet. With UB anything is allowed including compiling it correctly. A future compiler version may replace `MaybeUninit::uninit().assume_init()` with `intrinsics::unreachable()`, which makes the syscall trivially unreachable.

sunfishcode: On one hand, compilers are very unlikely to actually do this here; we don't do arithmetic on these values, we don't branch on them, print them, share them across threads, or observe their values in any way, and they have type `usize` where every bitpattern has an inert meaning. Compilers like LLVM sometimes generate code that does what we're doing within in their own IRs. In short, we don't do anything related to the reasons why `.assume_uninit()` has UB.

But on the other other hand, the alternative here is to just have the outline assembly move the `nr` parameter into the desired register, which is cheap and straightforward. So I filed https://github.com/bytecodealliance/rsix/issues/62 to do that.

Urgau: Oh! I now get your point. I've posted a fix https://github.com/bytecodealliance/rsix/pull/64 for `arm` and `riscv64`. Thanks for pointing this out.

bjorn3: > we don't do arithmetic on these values, we don't branch on them, print them, share them across threads, or observe their values in any way

This doesn't matter: https://www.ralfj.de/blog/2020/07/15/unused-data.html

> and they have type usize where every bitpattern has an inert meaning

The question is whether integers have undef as valid bitpattern. Undef is not a random bitpattern, it is a fundamentally different value that may or may not be lowered to a random bitpattern at machine level: https://www.ralfj.de/blog/2019/07/14/uninit.html If it is decided that undef is not a valid bitpattern for integers, this code will be instant UB.

sunfishcode: We don't even observe the values in dead code here; there's nothing to speculate, hoist, sink, strength-reduce, or partially evaluate. But I ultimately do agree; I'm just a little sad that the RISC-V outline assembly has to be more than `ecall; ret` :-).

PR #55, opened 2021-09-13
Implement `chdir`, `fchdir`, `getcwd` and `ttyname`

Urgau:
This pull-request implement the following functions: `chdir`, `fchdir`, `getcwd` and `ttyname` with their unit test (expect for `ttyname`, because I don't know how to test it reliably and simply).

`getcwd` and `ttyname` use the same routine that was already present for reusing an `OsString` if possible.

### DIFF
--- a/src/fs/cwd.rs
+++ b/src/fs/cwd.rs
@@ -1,5 +1,7 @@
+use std::ffi::OsString;
+
 use crate::imp;
-use crate::io::RawFd;
+use crate::io::{self, RawFd};
 use io_lifetimes::BorrowedFd;
 
 /// `AT_FDCWD`—Returns a handle representing the current working directory.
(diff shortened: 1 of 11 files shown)


src/imp/linux_raw/syscalls.rs
```diff
@@ -3012,6 +3017,56 @@ pub(crate) fn gettid() -> Pid {
...
+
+    // Get the proc path of the tty
+    let procname = procfdname(fd.as_raw_fd());
+
+    // And finaly gatter the ttyname by reading the proc path
+    readlink(
+        // SATEFY: procfdname always return a nul terminated string
+        unsafe { CStr::from_bytes_with_nul_unchecked(&procname) },
+        buf,
+    )?;
```
sunfishcode, line 3065: `readlink("/proc/self/fd/...")` isn't always reliable; file handles for deleted files or pipes end up being symlinks to special strings which aren't paths. We should check that the resulting path can be opened with `stat` and that the result's `st_dev`+`st_ino` matches that of `fd`.

src/imp/linux_raw/syscalls.rs
```diff
@@ -3012,6 +3017,56 @@ pub(crate) fn gettid() -> Pid {
...
+    buf
+}
+
+#[inline]
+pub(crate) fn ttyname(fd: BorrowedFd<'_>, buf: &mut [u8]) -> io::Result<()> {
+    // Check that the fd is really a tty
+    ioctl_tiocgwinsz(fd)?;
+
+    // Get the proc path of the tty
+    let procname = procfdname(fd.as_raw_fd());
```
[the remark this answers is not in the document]
Urgau: I should have check more carefully :grinning:. Done.

src/fs/cwd.rs
```diff
@@ -27,3 +29,37 @@ pub fn cwd() -> BorrowedFd<'static> {
         BorrowedFd::<'static>::borrow_raw_fd(at_fdcwd)
     }
 }
+
+/// `getcwd(reuse)`
```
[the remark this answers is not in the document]
Urgau: Done. I was indeed a bit confuse about that, it's the reason why I wrote `reuse` but didn't seem right to me. Not writing anything can have a meaning. I don't know what would be better; in the mean time a removes `reuse` part.

src/imp/linux_raw/syscalls.rs
Urgau: Yeah you're right, I was a little bit lazy after 25 github actions to get every `cfg` right. Done.

sunfishcode: I can sympathize. It's a lot :-}.

sunfishcode:
Looks good!

PR #57, opened 2021-09-14
Fix undefined behavior in `ttyname`.

sunfishcode:
Ensure that the buffer has enough storage for us to add a
NUL-terminator, and pass the subrange of the buffer which contains a
NUL-terminated string to `CStr::from_bytes_with_nul`.

I missed these in the initial code review.

For safety, this also switches from `from_bytes_with_nul_unchecked` to
`from_bytes_with_nul` for now.

### DIFF
--- a/src/imp/linux_raw/syscalls.rs
+++ b/src/imp/linux_raw/syscalls.rs
@@ -3029,8 +3029,9 @@ pub(crate) fn ttyname(fd: BorrowedFd<'_>, buf: &mut [u8]) -> io::Result<()> {
     let r = readlinkat(proc_self_fd, DecInt::from_fd(&fd).as_c_str(), buf)?;
 
     // If the number of bytes is equal to the buffer length, truncation may
-    // have occurred. And we need one extra byte for a NUL terminator.
-    if buf.len() - r <= 1 {
+    // have occurred. This check also ensures that we have enough space for
+    // adding a NUL terminator.
+    if r == buf.len() {
         return Err(crate::io::Error::RANGE);
     }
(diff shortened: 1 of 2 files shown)


Urgau:
Oh! Sorry about that. Thanks for catching it.

src/imp/linux_raw/syscalls.rs
```diff
@@ -3028,17 +3028,16 @@ pub(crate) fn ttyname(fd: BorrowedFd<'_>, buf: &mut [u8]) -> io::Result<()> {
     // Gatter the ttyname by reading the 'fd' file inside 'proc_self_fd'
     let r = readlinkat(proc_self_fd, DecInt::from_fd(&fd).as_c_str(), buf)?;
 
-    // Make if possible the string nul terminated
-    if r == buf.len() {
+    // If the number of bytes is equal to the buffer length, truncation may
+    // have occurred. And we need one extra byte for a NUL terminator.
+    if buf.len() - r <= 1 {
```
Urgau, line 3033: But I don't understand why you changed that.

Urgau: [musl](https://github.com/ifduyue/musl/blob/3f701faace7addc75d16dea8a6cd769fa5b3f260/src/unistd/ttyname_r.c#L18) had the same exact check.

sunfishcode: Ah, right. I was thinking we needed to reserve an extra byte for the NUL terminator, but the original check effectively does that. I've restored the original check now.

PR #64, opened 2021-09-15
Fix undefined behavior in arm and riscv64 outline assembly

Urgau:
This pull-request fix a undefined behavior in the arm and riscv64 outline assembly.

The undefined behavior occurs when doing `let u = std::mem::MaybeUninit::uninit().assume_init();`, which is obviously UB but was considered ok because `u` was never used in the underline calls (our asm and kernel) but as with many things we were wrong:
> No, you want an uninit integer, which we aren't sure yet is allowed at all. Even if it is allowed, calling .assume_uninit() is UB when you haven't initialized the MaybeUninit yet. With UB anything is allowed including compiling it correctly. A future compiler version may replace MaybeUninit::uninit().assume_init() with intrinsics::unreachable(), which makes the syscall trivially unreachable.

The UB is fix by using the same technique already used by `aarch64` and `x86_64` which is to put the `nr` after the last argument instead of always putting it in the 8th place (by adding extra dummy arguments).

cc @bjorn3


sunfishcode:
Looks good, thanks for fixing this!

PR #86, opened 2021-10-20
remove windows support from Arg

nivkner:
according to the readme, rsix not support Windows,
however the Arg trait in `src/path/arg.rs` had the method `as_os_str` conditionally compiled for windows,
which means it would never be used, so i removed that method.


sunfishcode:
Makes sense! At one point I anticipated moving `Arg` into its own crate to use in other projects which support Windows, but I ended up not needing it.

PR #91, opened 2021-10-21
Add fork

nivkner:
in `linux_raw`, `fork` is implemented using `clone` because it is supported on more architectures.


sunfishcode:
Thanks!

PR #99, opened 2021-10-29
WIP: Add ioctl_blkpbszget

Mek101:
Hi: first ever rust contribution to any mid-sized project, so it might be a bit rough on the edges(?).

The `ioctl` `BLKPBSZGET` system call is very poorly documented despite being [present since 2.6.32](https://github.com/karelzak/util-linux/blob/0a08200bd5664d1849e477f7f776ab4d13bb8422/lib/blkdev.c#L213) and I wrote the implementation mostly around what little I could gather. In syntheses: given a file descriptor of a block device, it writes to a uint pointer the physical block size of such device.

According to [this stackoverflow response](https://stackoverflow.com/questions/15944254/physical-disk-block-size-on-posix-using-c-c) it's only available on linux; macOS, *BDS and other unixes using different arguments for `ioctl`.

This is currently marked as WIP because linux-sys-raw is apparently missing the `BLKPBSZGET` definitions, tracking issue [here](https://github.com/sunfishcode/linux-raw-sys/issues/7)

### DIFF
--- a/src/imp/libc/io/syscalls.rs
+++ b/src/imp/libc/io/syscalls.rs
@@ -326,7 +326,7 @@ pub(crate) fn eventfd(initval: u32, flags: EventfdFlags) -> io::Result<OwnedFd> 
     unsafe { syscall_ret_owned_fd(c::syscall(c::SYS_eventfd2, initval, flags.bits())) }
 }
 
-#[cfg(any(target_os = "android", target_os = "linux"))]
+#[cfg(target_os = "linux")]
 #[inline]
 pub(crate) fn ioctl_blksszget(fd: BorrowedFd) -> io::Result<u32> {
     let mut result = MaybeUninit::<c::c_uint>::uninit();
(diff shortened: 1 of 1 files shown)


Mek101:
After the merge of https://github.com/sunfishcode/linux-raw-sys/pull/8, this PR compiles correctly on the `linux-raw` backend by pointing the `linux-raw-sys` dependency to the git repository. However it seems that the crate `libc` doesn't support either argument.

sunfishcode:
If you want, you could submit a PR to the libc crate to add `BLKPBSZGET` and `BLKSSZGET` there. However, since these are very specialized features, it'd also be ok to add them to only linux_raw here.

Mek101:
I've opened an upstream issue about it: https://github.com/rust-lang/libc/issues/2500, let's see if they respond.
As for submitting a PR upstream myself, I need to take a look on how the crate works before giving it a try.

Mek101:
Sent a PR to libc

Mek101:
Libc PR is merged: https://github.com/rust-lang/libc/pull/2548

sunfishcode:
Great, so this just needs to be updated to use the new libc, and to fix a few merge conflicts, it looks like it should be ready to go!

Mek101:
The libc with the PR in it hasn't been released yet. Right now it's only on master

Mek101:
Ok, I checked, and apparently BLK* are available on libc >=0.2.111

Mek101:
Temporarily removed android support since I forgotten to add it upstream in libc.

Mek101:
Should I squash everything in a single commit?

sunfishcode:
For PRs with iterative commits, I use squash-and-merge so that's not necessary :-). Thanks for contributing this feature!

PR #108, opened 2021-11-07
allocate for execve outside of `imp::syscalls`

nivkner:
this moves all allocations outside the call to `imp::syscalls::execve`,
this would allow a future implementation of `posix_spawn` to use `execve` in the child process without worrying about the allocator lock causing deadlock.

essentially the idea is that functions that would be used inside `posix_spawn` would be `imp::syscalls` functions, which are safe to call inside `fork`/`vfork`, without constraining the public API to be `async-signal-safe`.


sunfishcode:
Thanks!

PR #109, opened 2021-11-08
fix sched_getaffinity

nivkner:
this function currently panics because it expects the syscall to behave like the `glibc` wrapper, which returns either 0 or an error.
the linux syscall returns a value on success.

### DIFF
--- a/src/imp/linux_raw/syscalls.rs
+++ b/src/imp/linux_raw/syscalls.rs
@@ -491,12 +491,15 @@ pub(crate) fn getrandom(buf: &mut [u8], flags: GetRandomFlags) -> io::Result<usi
 #[inline]
 pub(crate) fn sched_getaffinity(pid: Pid, cpuset: &mut RawCpuSet) -> io::Result<()> {
     unsafe {
-        ret(syscall3(
+        // the raw linux syscall returns the size (in bytes) of the cpumask_t data type,
+        // that is used internally by the kernel to represent the CPU set bit mask
+        ret_usize(syscall3(
             nr(__NR_sched_getaffinity),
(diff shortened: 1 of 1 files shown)


src/imp/linux_raw/syscalls.rs
```diff
@@ -491,12 +491,15 @@ pub(crate) fn getrandom(buf: &mut [u8], flags: GetRandomFlags) -> io::Result<usi
...
-        ret(syscall3(
+        // the raw linux syscall returns the size (in bytes) of the cpumask_t data type,
+        // that is used internally by the kernel to represent the CPU set bit mask
+        ret_usize(syscall3(
             nr(__NR_sched_getaffinity),
             c_uint(pid.as_raw()),
             size_of::<RawCpuSet, _>(),
             by_mut(&mut cpuset.bits),
         ))
+        .map(|_| ())
```
sunfishcode, line 502: If the kernel's size is less than `RawCpuSet`'s size, does the kernel still zero out the rest of `RawCpuSet` or do we need to do that manually?

nivkner: according to [the manual](https://linux.die.net/man/2/sched_getaffinity) the kernel uses an `unsigned long`.

sunfishcode: If it leaves the rest as-is, that seems like a problem if our users are expecting a full `RawCpuSet`'s worth of data. It's fine if `RawCpuSet` is bigger than what the kernel uses; we just need to make sure we set any extra bits to zero.

sunfishcode:
Thanks! I'll publish a new release soon.

PR #149, opened 2022-01-03
Compare the pre-built versions only in a git checkout.

PiotrSikora:
Signed-off-by: Piotr Sikora <piotrsikora@google.com>


PiotrSikora:
Note that the original `assert` breaks `feature=cc`, since it depends on the host's compiler, and `cargo build --features=cc` always panics on my system due to a changed `librustix_outline_x86_64.a`.

Performing those checks only in a git checkout prevents breaking crates that depend on it, and should result in those checks done only during the development of `rustix`, and not when using it.

sunfishcode:
Makes sense, thanks!

PR #160, opened 2022-01-07
Add `msync` (#158)

### LINKED ISSUE #158: Is libc::msync function implemented?
Hi, is `libc::msync` or equivalent function implemented?

al8n:
Hi, this is a try to implement `msync` and fix #158, for such low-level programming, I am a total newbie and not familiar with which `target_os` should be configured. I am looking forward to the instructions, thanks!

### DIFF
--- a/src/imp/libc/io/syscalls.rs
+++ b/src/imp/libc/io/syscalls.rs
@@ -16,6 +16,8 @@ use super::super::offset::{libc_preadv2, libc_pwritev2};
 use super::Advice as IoAdvice;
 #[cfg(target_os = "linux")]
 use super::MremapFlags;
+#[cfg(not(target_os = "wasi"))]
+use super::MsyncFlags;
 #[cfg(not(any(target_os = "ios", target_os = "macos", target_os = "wasi")))]
 use super::PipeFlags;
 use super::PollFd;
@@ -300,6 +302,18 @@ pub(crate) fn madvise(addr: *mut c::c_void, len: usize, advice: IoAdvice) -> io:
         } else {
             unsafe { ret(c::madvise(addr, len, advice as c::c_int)) }
         }
+    }
+}
+
+#[cfg(not(target_os = "wasi"))]
+pub(crate) unsafe fn msync(addr: *mut c::c_void, len: usize, flags: MsyncFlags) -> io::Result<()> {
+    let err = c::msync(addr, len, flags.bits());
+
+    // `msync` returns its error status rather than using `errno`.
+    if err == 0 {
+        Ok(())
+    } else {
+        Err(io::Error(err))
     }
 }
 
--- a/src/imp/libc/io/types.rs
+++ b/src/imp/libc/io/types.rs
@@ -259,6 +259,23 @@ bitflags! {
     }
 }
 
+#[cfg(not(target_os = "wasi"))]
+bitflags! {
+    /// `MS_*` flags for use with [`msync`].
+    ///
+    /// [`msync`]: crate::io::msync
+    pub struct MsyncFlags: i32 {
+        /// `MS_SYNC` Requests an update and waits for it to complete.
+        const SYNC = c::MS_SYNC;
+        /// `MS_ASYNC` Specifies that an update be scheduled,
+        /// but the call returns immediately.
+        const ASYNC = c::MS_ASYNC;
+        /// `MS_INVALIDATE` Asks to invalidate other mappings of the same file (so
+        /// that they can be updated with the fresh values just written).
+        const INVALIDATE = c::MS_INVALIDATE;
+    }
+}
+
 #[cfg(any(target_os = "android", target_os = "linux"))]
 bitflags! {
     /// `MLOCK_*` flags for use with [`mlock_with`].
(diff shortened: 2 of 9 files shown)


src/io/mod.rs
```diff
@@ -21,6 +21,8 @@ mod is_read_write;
 mod madvise;
 #[cfg(not(any(windows, target_os = "wasi")))]
 mod mmap;
+#[cfg(not(any(windows, target_os = "redox", target_os = "wasi")))]
```
Urgau, line 24:
```suggestion
#[cfg(not(any(windows, target_os = "wasi")))]
```

src/io/mod.rs
```diff
@@ -71,6 +73,8 @@ pub use mmap::{
 pub use mmap::{mlock_with, MlockFlags};
 #[cfg(any(linux_raw, all(libc, target_os = "linux")))]
 pub use mmap::{mremap, mremap_fixed, MremapFlags};
+#[cfg(not(any(windows, target_os = "redox", target_os = "wasi")))]
```
Urgau, line 76:
```suggestion
#[cfg(not(any(windows, target_os = "wasi")))]
```

src/imp/libc/io/syscalls.rs
```diff
@@ -16,6 +16,7 @@ use super::super::offset::{libc_preadv2, libc_pwritev2};
 use super::Advice as IoAdvice;
 #[cfg(target_os = "linux")]
 use super::MremapFlags;
+use super::MsyncFlags;
```
Urgau, line 19:
```suggestion
#[cfg(not(target_os = "wasi"))]
use super::MsyncFlags;
```

tests/io/mmap.rs
```diff
@@ -142,3 +142,18 @@ fn test_madvise() {
         munmap(addr, 8192).unwrap();
     }
 }
+
+#[test]
+fn test_msync() {
```
Urgau, line 147: Putting the test here is a bit too restrictive as the file is:
```rust
#![cfg(not(any(target_os = "redox", target_os = "wasi")))]
```
but `msync` is supported on `redox`. Not sure if it's a big deal at all.

Urgau (approved):
Not the maintainer but this looks good.
I just leave some suggestions about `cfg`, they should fix the CI.

src/imp/libc/io/types.rs
```diff
@@ -259,6 +259,23 @@ bitflags! {
     }
 }
 
+#[cfg(not(target_os = "wasi"))]
+bitflags! {
+    /// `MS_*` flags for use with [`msync`].
+    ///
+    /// [`msync`]: crate::io::msync
+    pub struct MsyncFlags: i32 {
+        /// `MS_SYNC` Requests an update and waits for it to complete.
```
cgwalters, line 268: Looking at the other `bitflags!` in this crate, it seems like in some cases the doc comment just repeats the identifier - probably to avoid all the additional work of copying documentation in.

But rustdoc doesn't require us to repeat the identifier name.  I think it'd be cleaner to drop it.  Like this:

```suggestion
        /// Requests an update and waits for it to complete.
```

And similarly elsewhere.

cgwalters (commented):
Not a real contributor here, so a superficial LGTM apart from nits.

I'd also squash this commit down into one, and copy the text you have in the PR description into the commit message too.

Take a look at e.g. https://github.com/bytecodealliance/rustix/commit/bc8d627fe1de2503a83985cd75cf6206e8b06ce9 for a recent commit in this repo.  Also see https://chris.beams.io/posts/git-commit/

(Again, I am not an "official" reviewer here, but I plan to contribute some so I'm trying to "pay it forward" by also doing some review myself)

src/imp/libc/io/types.rs
sunfishcode: The idea behind including the C identifier is to make it easier for people porting libc or C code, people familiar with the C APIs, and people looking at the POSIX/Linux/etc. docs which rustix links to, to help them search for things and see how things correspond.

In this specific case, "Requests an update and waits for it to complete" is a fine summary comment, but if someone really needs to know the fine details, they'll need to consult other sources. It'd indeed be cleaner if rustix could have self-sufficient documentation, but, as you say, that's a lot of additional work. We may do that work some day, but for now, I think it makes sense to include the C identifiers.

al8n:
Thanks for all of the reviewers! Hi, @sunfishcode, I want to do more work for this crate, is there some other I can help with?

sunfishcode:
Very cool! I hope to have time to review this, probably in the next few days. For other things to work on, some syscalls which have come up recently that rustix doesn't have yet are Linux's `splice`, `vmsplice`, and `tee`.

sunfishcode:
I should also mention, if you work on `splice`, it's worth adding a mention [here](https://github.com/sunfishcode/mustang/issues/91) to help avoid others doing duplicate work.

src/imp/libc/io/types.rs
cgwalters: OK, makes sense.  Though what may be even better is adding it as [an alias](https://doc.rust-lang.org/rustdoc/advanced-features.html#add-aliases-for-an-item-in-documentation-search)?

sunfishcode:
The code here looks good; the only thing left to do is adjust the `cfg` lines for the tests: https://github.com/bytecodealliance/rustix/pull/160/files#r781276952.

tests/io/mmap.rs
al8n: Finished!

sunfishcode:
Cool, thanks!

PR #164, opened 2022-01-10
Avoid shared references to mutable state in `readv` et al.

sunfishcode:
Change the signatures for `readv`, `writev`, `preadv`, `pwritev`, and
similar to use `&mut [IoSliceMut]` rather than `&[IoSliceMut]`.

If we were to write userspace implementations of these functions instead
of making system calls, using `&[IoSliceMut]` would lead to compile-time
errors. For example, code like this:

```rust
fn readv(bufs: &[IoSliceMut]) {
    if let Some(buf) = bufs.get(0) {
        if let Some(first) = buf.get_mut(0) {
            *first = 0;
        }
    }
}
```

would get this error:

```
error[E0596]: cannot borrow `*buf` as mutable, as it is behind a `&` reference
 --> test.rs:5:30
  |
4 |     if let Some(buf) = bufs.get(0) {
  |                 --- help: consider changing this to be a mutable reference: `&mut IoSliceMut<'_>`
5 |         if let Some(first) = buf.get_mut(0) {
  |                              ^^^^^^^^^^^^^^ `buf` is a `&` reference, so the data it refers to cannot be borrowed as mutable

```

Rustix API signatures shouldn't be defined in ways that would require
unsafe code in the implementation. Rustix does use unsafe code to make
actual system calls, but that's an implementation detail that shouldn't
be visible in the API.

This also matches how std's [`read_vectored`] et al are defined.

[`read_vectored`]: https://doc.rust-lang.org/stable/std/io/trait.Read.html#method.read_vectored

### DIFF
--- a/src/imp/libc/io/syscalls.rs
+++ b/src/imp/libc/io/syscalls.rs
@@ -89,7 +89,7 @@ pub(crate) fn pwrite(fd: BorrowedFd<'_>, buf: &[u8], offset: u64) -> io::Result<
     Ok(nwritten as usize)
 }
 
-pub(crate) fn readv(fd: BorrowedFd<'_>, bufs: &[IoSliceMut]) -> io::Result<usize> {
+pub(crate) fn readv(fd: BorrowedFd<'_>, bufs: &mut [IoSliceMut]) -> io::Result<usize> {
     let nread = unsafe {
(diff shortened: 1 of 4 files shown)


cgwalters (approved):

sunfishcode:
Thanks for the review!

PR #165, opened 2022-01-12
Remove the `test_wait` test.

sunfishcode:
This test is dangerous as it can end up waiting for unrelated child
processes. In theory, we could fix this by spawning a separate process
that we completely control and running the test within that process,
but for now, just disable the test.

### DIFF
--- a/tests/process/wait.rs
+++ b/tests/process/wait.rs
@@ -23,21 +23,3 @@ fn test_waitpid() {
         .unwrap();
     assert!(status.stopped());
 }
-
-#[test]
-#[serial]
-fn test_wait() {
-    let child = Command::new("yes")
-        .stdout(Stdio::null())
-        .stderr(Stdio::null())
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
This test is also relying implicitly on the fact that nothing else in the address space is forking a process at the same time.  Granted, that's unlikely.  But still, I think the "wait for any child process" mode is really only something a Unix `init` process should be doing.  So longer term I think it'd make sense to explicitly fork these tests into their own process, not just serialize the two tests.

sunfishcode:
Yeah, that makes sense. I think it makes to remove that test for now; I've filed https://github.com/bytecodealliance/rustix/issues/168 to track this.

sunfishcode:
We now use `serial_test` to ensure that only one test is doing a `wait` at a time, which seems to be working reliably. And the test that was removed here is now in the tree.

PR #167, opened 2022-01-12
Test with the the "cc" feature enabled, in CI.

sunfishcode:
Enabling the "cc" feature causes the build.rs script to check that the
precompiled libraries match the libraries generated from source.

This also updates the libraries, in which the only change is the debug
info string mentioning the compilation directory chaging from "rsix"
to "rustix".


cgwalters (approved):
Seems sane, but it'd be nice to avoid committing binary files into git.  Can we check the sha256 digest or something instead?

sunfishcode:
Checking these into git isn't ideal, but it means that users don't need to install extra toolchain packages in order to build rustix. And hopefully this is only temporary until Rust stabilizes the features we need to get rid of the outline assembly files.

PR #169, opened 2022-01-13
Make the outline assembly files compatible with llvm-mc.

sunfishcode:
I did some investigation to see if building the outline assembly files
with llvm-mc might give us a way to avoid needing to have pre-assembled
libraries in the repo. It turns out Rust doesn't currently ship an llvm-mc,
so it doesn't solve that problem yet. But the changes needed to be
compatible with llvm-mc might open up other options, so I'm submitting
them as an independent change here.

This makes the assembly files compatible with llvm-mc:

 - Rename '.S' files to '.s' files, which corresponds to not preprocessing
   them with the C preprocessor, which we don't need.

 - Remove '.globl' declarations for symbols which are also '.weak', as these
   are redundant and get a warning in llvm-mc.


cgwalters (approved):
Superficial LGTM

sunfishcode:
Thanks!

PR #170, opened 2022-01-13
Make the `times` elements in `utimensat` and `futimens` a struct.

sunfishcode:
Instead of a 2-element array where users have to know which timestamp
goes at which index, use a `repr(C)` struct so that it has named
fields, while having the same layout.


cgwalters (approved):
Looks sane to me

PR #171, opened 2022-01-13
Make `CpuSet::new` a const fn.

sunfishcode:


sunfishcode:
Ah, this won't work in Rust 1.48.

PR #172, opened 2022-01-13
Refactor the `Mode` and `FileType` types.

sunfishcode:
This makes several changes to the `Mode` and `FileType` API, based on
experience porting code to rustix.

 - Remove the file-type `S_IF*` constants (eg. `S_IFREG`) from the `Mode`
   bitflags type. This way, `Mode` is just for permission modes, and the
   fact that Unix packs the permissions mode bits with the file type
   bits is just an implementation detail.

 - Also, remove the leading `I` from the `Mode` arm names, as `S_I` is
   the common prefix in the C API.

 - Make `mknodat` take an explicit `FileType` argument, now that the mode
   doesn't include the file type.

And add documentation advertising `Mode::from_raw_mode` and
`FileType::from_raw_mode` to users of `statat` and `fstat`.

### DIFF
--- a/src/fs/at.rs
+++ b/src/fs/at.rs
@@ -18,9 +18,9 @@ use crate::process::{Gid, Uid};
 use crate::{imp, path};
 use alloc::vec::Vec;
 use imp::fd::{AsFd, BorrowedFd};
-#[cfg(not(any(target_os = "ios", target_os = "macos", target_os = "wasi",)))]
-use imp::fs::Dev;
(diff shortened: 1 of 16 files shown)


cgwalters (approved):
Makes sense to me!  Having cleanly separate types for these is definitely nice.  I know in some places in my C code I'm passing `S_IFREG` into my `chmod()` calls etc.

sunfishcode:
Cool, thanks!

PR #174, opened 2022-01-17
Export `rustix::io::proc_self`

koute:


sunfishcode:
Thanks for the PR! It turns out, it's a goal of rustix's procfs code to avoid following bind mounts on top of procfs, and handing out  `proc_self` handles and having uses open files under it means users could end up following bind mounts. This is something I had already been working on (for /proc/self/maps in my case), so I've now finished my patch and generalized it to /proc/self/pagemap, and it's now posted as https://github.com/bytecodealliance/rustix/pull/176. With that, you can just call `rustix::io::proc_self_pagemap()` to read the pagemap file. Does that look like it'll work for you?

koute:
Yep, that would definitely work! Thanks! :+1:

PR #176, opened 2022-01-18
Implement `proc_self_maps`, `proc_self_pagemap`, and `proc_self_fdinfo`.

sunfishcode:
Add support for opening files within procfs, and add public APIs for
accessing `/proc/self/maps`, `/proc/self/pagemap`, and
`proc/self/<fd>/fdinfo`.

### DIFF
--- a/src/io/procfs.rs
+++ b/src/io/procfs.rs
@@ -463,6 +463,10 @@ fn open_and_check_file(dir: BorrowedFd, dir_stat: &Stat, name: &ZStr) -> io::Res
     // if we can find a regular file with an inode and name that matches the
     // file we just opened. If we can't find it, there could be a file bind
     // mount on top of the file we want.
+    //
+    // TODO: With Linux 5.8 we might be able to use `statx and
+    // `STATX_ATTR_MOUNT_ROOT` to detect mountpoints directly instead of doing
+    // this scanning.
     let dir = Dir::from(dot).map_err(|_err| io::Error::NOTSUP)?;
     for entry in dir {
         let entry = entry.map_err(|_err| io::Error::NOTSUP)?;
--- a/src/io/procfs.rs
+++ b/src/io/procfs.rs
@@ -5,6 +5,16 @@
 //! This module does a considerable amount of work to determine whether `/proc`
 //! is mounted, with actual `procfs`, and without any additional mount points
 //! on top of the paths we open.
+//!
+//! Why all the effort to detect bind mount points? People are doing all kinds
+//! of things with Linux containers these days, with many different privilege
+//! schemes, and we want to avoid making any unnecessary assumptions. Rustix
+//! and its users will sometimes use procfs *implicitly* (when Linux gives them
+//! no better options), in ways that aren't obvious from their public APIs.
+//! These filesystem accesses might not be visible to someone auditing the main
+//! code of an application for places which may be influenced by the filesystem
+//! namespace. So with the checking here, they may fail, but they won't be able
+//! to succeed with bogus results.
 
 use crate::fd::{AsFd, BorrowedFd};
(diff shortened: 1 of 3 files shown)


src/io/procfs.rs
```diff
@@ -306,3 +328,151 @@ type StaticFd = OnceCell<(OwnedFd, Stat)>;
...
+    let dot_stat = fstat(&dot).map_err(|_err| io::Error::NOTSUP)?;
+    if (dot_stat.st_dev, dot_stat.st_ino) != (dir_stat.st_dev, dir_stat.st_ino) {
+        return Err(io::Error::NOTSUP);
+    }
+
+    // `is_mountpoint` only works on directory mount points, not file mount
+    // points. To detect file mount points, scan the parent directory to see
+    // if we can find a regular file with an inode and name that matches the
+    // file we just opened. If we can't find it, there could be a file bind
+    // mount on top of the file we want.
```
cgwalters, line 455: Is there a real world attack that resulted from something creating a hostile bind mount over `/proc`?  I struggle to think of a case where malicious code could create a bind mount over e.g. `/proc/<pid>/fdinfo` or something but not also have the capability to create the same mount over e.g. `/lib/ld-linux.so.2` and have compromised the process from the start too.

(The answer here is probably better put in the doc comments at the top I guess)

cgwalters (approved):
(Mostly superficial again) LGTM

src/io/procfs.rs
sunfishcode: I don't know of any real-world attacks. It's mainly about how people are doing all kinds of different things with Linux containers these days, with all kinds of different blends of what we once called "root" and "user" privileges, so in general we want to avoid as many vectors for misdirection as possible. I've added some docs.

For a hypothetical example, processes which launch outside the sandbox and then use `chroot` or other things to enter it wouldn't be vulnerable to `/lib/ld-linux.so.2` inside the sandbox being redirected. Such programs may need to be careful about trusting filesystem paths once they've entered the sandbox, but rustix and especially cap-std will sometimes use procfs implicitly, in ways that aren't documented or obvious from the public API, which could make it tricky to avoid any such trust. With the checking here, rustix and its users may fail in some cases, but they shouldn't be able to succeed with bogus data.

cgwalters: OK.  I think though if you have `openat()` there is no reason to use `chroot()`.  `chroot()` is really a dangerous thing to use if you potentially have *anything* in your address space which isn't explicitly expecting it.

(Tangential but interesting side fact; One of the rationales for https://github.com/coreos/rpm-ostree/ avoiding using the [rpm shared library](https://github.com/rpm-software-management/rpm/) is actually because it's a systemd service offering a DBus API, and librpm since the very first days has used `chroot()` to run scripts.  rpm-ostree reimplements all that, creating a container for each script.  We also use `openat()` and will be using cap-std in more places)

IOW I think cap-std should really be thought of as a replacement for `chroot()`.

I'm not going to argue against the extra checks here, I just wanted to understand.  The new comment seems fine.

cgwalters: Maybe it's useful: systemd [uses statx](https://github.com/systemd/systemd/blob/main/src/basic/mountpoint-util.c#L193) for this.

sunfishcode: Thanks for asking questions! This has been helpful for me. I tend to agree about `chroot`, so I myself probably wouldn't write code that way, but I'm aiming to have rustix be ok with someone else doing it.

Also thanks for that systemd link; I didn't know about `STATX_ATTR_MOUNT_ROOT`. That looks new in Linux 5.8 so while we may use that to optimize, we'd still need the fallback code. I've added a comment about it.

PR #177, opened 2022-01-18
Add several features that are used in cargo

sunfishcode:
This adds `setsid`, `kill`, `statfs` system calls, and supporting types, constants, and utilities, which are things from libc that cargo currently uses explicitly. Cargo also uses libc transitively through std, but that's a [bigger project](https://blog.sunfishcode.online/port-std-to-rustix/).

This is based on #175.

### DIFF
--- a/src/imp/libc/process/types.rs
+++ b/src/imp/libc/process/types.rs
@@ -135,6 +135,121 @@ impl Resource {
     pub const Rss: Self = Self::As;
 }
 
+/// A signal number for use with [`kill`].
+///
+/// [`kill`]: crate::process::kill
+#[cfg(not(target_os = "wasi"))]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[repr(i32)]
+pub enum Signal {
+    /// `SIGHUP`
+    Hup = c::SIGHUP,
+    /// `SIGINT`
+    Int = c::SIGINT,
+    /// `SIGQUIT`
+    Quit = c::SIGQUIT,
+    /// `SIGILL`
+    Ill = c::SIGILL,
+    /// `SIGTRAP`
+    Trap = c::SIGTRAP,
+    /// `SIGABRT`, aka `SIGIOT`
+    #[doc(alias = "Iot")]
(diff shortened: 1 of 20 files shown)


src/imp/libc/process/types.rs
```diff
@@ -135,6 +135,121 @@ impl Resource {
...
 
+/// A signal number for use with [`kill`].
+///
+/// [`kill`]: crate::process::kill
+#[cfg(not(target_os = "wasi"))]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[repr(i32)]
+pub enum Signal {
+    /// `SIGHUP`
+    Hup = c::SIGHUP,
```
cgwalters, line 146: This is a taste thing but I think this would be better if the elements stayed uppercase.  The whole unix "let's drop the vowels to save space" makes having these rendered as capitalized words read strangely to me.  `Alrm` looks particularly odd for example.  Even if it's `Signal::Alrm`.  I think it's more obviously "Unix" if it's `Signal::ALRM`.  This is how the nix crate has it at least.

cgwalters (approved):
Mostly superficial LGTM

src/imp/libc/process/types.rs
sunfishcode: Good idea! I've now opted to go for `Signal::Alarm`, and similar for `Abrt`, `Chld`, and a few similar ones. This is consistent with rustix things like `CREATE` for `O_CREAT` and others.

Rustix overall aims to balance human-friendliness with C familiarity, and these "missing vowels" cases are easy places where we can add friendliness without losing much familiarity.

sunfishcode: Ah, I should also mention, the reason for avoiding all-caps is to be consistent with the [Rust convention](https://rust-lang.github.io/api-guidelines/naming.html) for enum variant names.

Overall, rustix is aiming for a balance between Rust-friendliness, human-friendliness, and C familiarity.

cgwalters: Ah, OK.  That makes sense.

PR #178, opened 2022-01-18
Write more comments.

sunfishcode:
Make a pass over the crate adding and cleanup up comments and doc aliases.


cgwalters (approved):
With this can we `#![deny(missing_docs)]`?

sunfishcode:
Thanks for bringing that up! We already had a `deny(missing_docs)`, though we previously had a few `allow(missing_docs)` local overrides, but this PR got pretty close to removing the last of the overrides, so I've now gone ahead and added the last few comments and removed the last overrides.

PR #180, opened 2022-01-19
Fix `readlinkat` syscalls to not be `readonly`.

sunfishcode:
The `readlinkat` syscall writes into the output buffer, so it is not
`readonly`, so use a non-`readonly` syscall wrapper for it. I believe
this bug isn't observable outside the crate, due to the restricted
ways that the actual syscall wrapper is used.

Also, mark several more syscalls that don't mutate memory as
`readonly`.


cgwalters:
> Also, mark several more syscalls that don't mutate memory as
`readonly`.

Probably would have been cleaner as a separate commit?

cgwalters (approved):

sunfishcode:
I've now split the fix part into a separate patch.

PR #184, opened 2022-01-23
impl `Debug` for `DecInt`

cgwalters:
This is a best practice to be friendly to users who may want
to add things to their own structs and want to simply `#[derive(Debug)]`
and I'm assuming its omission was accidental.


sunfishcode:
Yes, you assume correctly :-).

The CI failure here is just that the import of `fmt` is only used under `cfg(feature = "std")`.

sunfishcode:
Thanks!

PR #185, opened 2022-01-24
tests/statfs: Don't check `f_files != 0`

cgwalters:
At least as far as I can tell, the Linux btrfs code does not
set `f_files`.  I think likely because inode allocations are fully
dynamic.

Testing the block count should be sufficient.


cgwalters:
See e.g. https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/btrfs/super.c?h=v5.16#n2236 (no reference to `f_files`)
vs https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/xfs/xfs_super.c?h=v5.16#n820

PR #186, opened 2022-01-24
procfs: Don't try to validate the uid/gid of `/proc`

cgwalters:
I typically develop inside a https://github.com/containers/toolbox/
container.  In this scenario:

```
$ ls -ald /proc
dr-xr-xr-x. 526 nobody nobody 0 Jan 12 14:47 /proc
$
```

And that's expected and normal; the real root uid from outside
the user namespace is mapped to `nobody`; distinct from the uid 0
inside the userns.

Honestly, I am still somewhat skeptical of the value of all of
these checks.  We're already validating that `/proc`'s filesystem
magic is `PROC_SUPER_MAGIC` - that seems really more than sufficient.

### DIFF
--- a/src/io/procfs.rs
+++ b/src/io/procfs.rs
@@ -35,6 +35,7 @@ const PROC_ROOT_INO: u64 = 1;
 
 // Identify an entry within "/proc", to determine which anomalies to
 // check for.
+#[derive(Debug)]
 enum Kind {
     Proc,
     Pid,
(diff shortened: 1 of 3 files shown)


sunfishcode:
Yeah, some of that logic is likely superfluous. I was just being extra cautious and checking everything I could find, because it's difficult to be sure that all the ways of potentially subverting /proc are covered, with all the potential forms of confused-deputy problems that might be possible. I think the change here sounds reasonable though. As you say, with `PROC_SUPER_MAGIC`, we know it's really procfs regardless of who is listed as having mounted it.

PR #187, opened 2022-01-24
Add ioctl_blksszget and ioctl_blkpbszget to android.

Mek101:
In the previous pr (#99) I had to remove the syscalls form android because they weren't implemented in libc. Now with libc 0.2.113 they are: https://github.com/rust-lang/libc/pull/2626


sunfishcode:
Looks good overall; looks like there's just one more type fix needed.

Mek101:
The issue seems to be that certain ioctl implementations require a unsigned int while others want a signed int.
Apparently I hadn't properly set the constants types when I contributed to libc. https://github.com/rust-lang/libc/pull/2652

sunfishcode:
Thanks for seeing this through!

PR #189, opened 2022-01-24
procfs: Add a wrapper to open a directory in `/proc`

cgwalters:
This helper avoids the duplication of the open flags for example,
and is prep for further work.

### DIFF
--- a/src/io/procfs.rs
+++ b/src/io/procfs.rs
@@ -207,6 +207,17 @@ fn is_mountpoint(file: BorrowedFd<'_>) -> bool {
     }
 }
 
+/// Open a directory in `/proc`, mapping all errors to `io::Error::NOTSUP`.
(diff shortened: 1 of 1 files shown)


src/io/procfs.rs
```diff
@@ -207,6 +207,17 @@ fn is_mountpoint(file: BorrowedFd<'_>) -> bool {
     }
 }
 
+// Open a directory in `/proc`, mapping all errors to `-ENOTSUP`
```
sunfishcode, line 210: While the value of the ultimate error is indeed `-ENOTSUP` with the negation, that's a lower level of abstraction from the perspective of this code. This code should think of it as `io::Error::NOTSUP` or similar.

Also, this comment can be a triple-slash comment, since it is still a documentation comment, even though it's a private function.

sunfishcode (commented):
Nice cleanup! Just one minor comment below:

PR #190, opened 2022-01-24
Add a CI check for the prebuild debug-mode libraries.

sunfishcode:
The existing CI code checked it for --release builds; this ensures it's
checked for debug builds as well.

This also updates the debug builds where became out of date after
bc8ea07a added comments to the .s files, changing the line numbers in
the .debug_line sections.


sunfishcode:
The current working directory path of the build is embedded in the debug info and is different between my build and the github build.

sunfishcode:
It looks like building the libraries in a directory named "/home/runner/work/rustix/rustix" was sufficient to reproduce the same binaries that are produced in github CI here!

cgwalters (approved):
I haven't fully followed this but - are these prebuilt libraries only needed for CI?  Can we skip publishing them in the crates.io source and skip the tests if they're not present?

Right now our projects vendor sources and I'd like to avoid vendoring `.a` files.

sunfishcode:
These are used to avoid requiring downstream users needing to have (cross-) compilers installed.

You can build rustix without using them, by enabling the "cc" feature in cargo, and having the appropriate compilers installed. That will build everything from source.

Rustix also has inline asm that it can use instead. This currently only works on nightly Rust, but it's [in the process of being stabilized](https://github.com/rust-lang/rust/pull/91728), so in the future it'll work on stable Rust as well.

PR #191, opened 2022-01-25
Implement `setrlimit` and `prlimit`.

sunfishcode:
rustix already had the types defined for `getrlimit`, so this just adds
the new `setrlimit` and Linux-specific `prlimit`.

### DIFF
--- a/src/imp/libc/process/syscalls.rs
+++ b/src/imp/libc/process/syscalls.rs
@@ -295,16 +295,14 @@ pub(crate) fn getrlimit(limit: Resource) -> Rlimit {
 #[cfg(not(any(target_os = "fuchsia", target_os = "redox", target_os = "wasi")))]
 #[inline]
 pub(crate) fn setrlimit(limit: Resource, new: Rlimit) -> io::Result<()> {
-    let rlim_cur: libc::rlim_t = match new.current {
+    let rlim_cur = match new.current {
         Some(r) => r.try_into().map_err(|_| io::Error::INVAL)?,
-        None => None,
-    }
-    .unwrap_or(LIBC_RLIM_INFINITY as _);
-    let rlim_max: libc::rlim_t = match new.maximum {
+        None => LIBC_RLIM_INFINITY as _,
+    };
+    let rlim_max = match new.maximum {
         Some(r) => r.try_into().map_err(|_| io::Error::INVAL)?,
-        None => None,
-    }
-    .unwrap_or(LIBC_RLIM_INFINITY as _);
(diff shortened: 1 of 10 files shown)


src/imp/libc/process/syscalls.rs
```diff
@@ -288,6 +292,51 @@ pub(crate) fn getrlimit(limit: Resource) -> Rlimit {
...
+            limit as _,
+            as_ptr(&lim),
+            result.as_mut_ptr(),
+        ));
+        let result = result.assume_init();
+        let current = if result.rlim_cur == LIBC_RLIM_INFINITY {
+            None
+        } else {
+            result.rlim_cur.try_into().ok()
+        };
```
cgwalters, line 330: I am not totally following the `try_into()` bits here around the edge/maximum cases.  May be worth factoring this into a helper function that has a bit of commentary?

(Here and elsewhere)

src/imp/linux_raw/process/syscalls.rs
```diff
@@ -387,6 +390,97 @@ pub(crate) fn getrlimit(limit: Resource) -> Rlimit {
...
+        rlim_cur: new
+            .current
+            .unwrap_or(linux_raw_sys::general::RLIM_INFINITY as _),
+        rlim_max: new
+            .maximum
+            .unwrap_or(linux_raw_sys::general::RLIM_INFINITY as _),
+    };
+    #[cfg(target_pointer_width = "32")]
+    unsafe {
+        match ret(syscall4(
```
cgwalters, line 405: I only skimmed this code, seems OK.  I just have to express my reservations on an ongoing basis around the direct syscall path.  There are so many corner cases and bits of complexity to consider around 64 vs 32 bit, any architecture oddities etc.

(But again, the code in a quick skim seems fine)

cgwalters (approved):

src/imp/libc/process/syscalls.rs
sunfishcode: Good idea, this was getting pretty cluttered. I've now factored out helper functions to do the conversions, and fixed those `.try_into().ok()` things, which we shouldn't ever do.

src/imp/linux_raw/process/syscalls.rs
sunfishcode: Acknowledged. I think Rust's general strictness about integer conversions, along with `try_into()` and `Result` are great tools that help a lot here, compared to my experience writing this kind of code in C. But even so, bugs in this area in practice are something I'm keeping an eye out for, because above all, rustix needs to be a practical library.

PR #195, opened 2022-01-25
Factor out several more type conversions in the syscalls code.

sunfishcode:
This adds conversion functions for converting `Pid`, `Signal`, and
`Advice` values into syscall argument registers.


cgwalters (approved):
Definitely much nicer to read than having the conversions inline!

PR #196, opened 2022-01-26
Implement `fcntl_add_seals` etc.

sunfishcode:
Implement `fcntl_add_seals`, and add a `SealFlags` to describe seal flag
values in `fcntl_add_seals` and `fcntl_get_seals`.

Also, add `MFD_HUGETLB` and accompanying flag values.

### DIFF
--- a/src/imp/libc/fs/types.rs
+++ b/src/imp/libc/fs/types.rs
@@ -506,6 +506,61 @@ bitflags! {
 
         /// `MFD_ALLOW_SEALING`
         const ALLOW_SEALING = c::MFD_ALLOW_SEALING;
+
+        /// `MFD_HUGETLB` (since Linux 4.14)
+        const HUGETLB = c::MFD_HUGETLB;
+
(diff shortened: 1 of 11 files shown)


src/imp/libc/fs/types.rs
```diff
@@ -506,6 +506,61 @@ bitflags! {
...
+        /// `MFD_HUGE_16GB`
+        const HUGE_16GB = c::MFD_HUGE_16GB;
+    }
+}
+
+#[cfg(any(
+    target_os = "android",
+    target_os = "linux",
+    target_os = "fuchsia",
+    target_os = "freebsd"
```
cgwalters, line 544: Wait, that's interesting - I had thought this was a Linux-specific thing.  Interesting to see it re-implemented in Fuschia.  And I had no idea FreeBSD did them too.  Ohh, looks like it may have been motivated by linux source/binary compatibility.

cgwalters (approved):

PR #197, opened 2022-01-26
Add Linux userfaultfd ioctl support.

sunfishcode:
This adds functions and types for the Linux userfaultfd syscall and several
associated ioctls.


cgwalters (approved):
Only skimmed, LGTM.  But...I am a bit concerned at the number of new APIs added here that aren't exercised via tests.

I am curious, is there a specific Rust user of `userfaultfd` you have in mind?  Perhaps we should be considering a "reverse dependency testing" setup.  I know for example you were looking at porting cargo to rustix, so it could make sense to e.g. run cargo's test suite in CI here, etc.

sunfishcode:
> Only skimmed, LGTM. But...I am a bit concerned at the number of new APIs added here that aren't exercised via tests.
> 
> I am curious, is there a specific Rust user of `userfaultfd` you have in mind? Perhaps we should be considering a "reverse dependency testing" setup. I know for example you were looking at porting cargo to rustix, so it could make sense to e.g. run cargo's test suite in CI here, etc.

The immediate user would be [this crate](https://github.com/fastly/userfaultfd-rs). However, I'm still just experimenting with it, and am not yet sure how far this will go.

I agree that it'd be good to have more tests in general. I do a fair amount of reverse-dependency testing manually right now, however I don't know of a good way to automate this.

A pattern I'm hoping which helps here is the one used in #206, which is to write tests in an example program, so that they run in a separate process, to avoid interfering with the test harness process. Not everything will need that, but I'm hoping that pattern will help for things that do.

cgwalters:
> however I don't know of a good way to automate this.

I have a similar situation in containers-image-proxy-rs where we do:

https://github.com/containers/containers-image-proxy-rs/blob/3d83477f527fc8791da699ddce5f2605aeb20beb/.github/workflows/ci.yaml#L53
Which then invokes:
https://github.com/containers/containers-image-proxy-rs/blob/3d83477f527fc8791da699ddce5f2605aeb20beb/ci/test-ostree-rs-ext.sh

That said, the scope of this for rustix is much, much larger and so it might need to be opt-in.  (And this gets into CI budgets etc.)

PR #200, opened 2022-01-27
Update to linux-raw 0.0.39.

sunfishcode:
This version drops support for multiple Linux API versions, so this drops
all the `v5_4::` and `v5_11::` qualifiers.

### DIFF
--- a/src/imp/linux_raw/net/syscalls.rs
+++ b/src/imp/linux_raw/net/syscalls.rs
@@ -821,6 +821,10 @@ pub(crate) mod sockopt {
     use crate::net::{Ipv4Addr, Ipv6Addr, SocketType};
     use core::convert::TryInto;
     use core::time::Duration;
+    use linux_raw_sys::general::{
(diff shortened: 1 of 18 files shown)


src/imp/linux_raw/net/syscalls.rs
```diff
@@ -998,13 +1002,50 @@ pub(crate) mod sockopt {
...
                 if timeout == DURATION_ZERO {
                     return Err(io::Error::INVAL);
                 }
+                let mut timeout = __kernel_timespec {
+                    tv_sec: timeout.as_secs().try_into().unwrap_or(i64::MAX),
+                    tv_nsec: timeout.subsec_nanos() as _,
+                };
+                if timeout.tv_sec == 0 && timeout.tv_nsec == 0 {
+                    timeout.tv_nsec = 1;
+                }
```
cgwalters, line 1016: This is another nontrivial conversion logic that seems like it'd be best as a helper?

cgwalters (approved):

src/imp/linux_raw/net/syscalls.rs
sunfishcode: Good idea; done..

PR #202, opened 2022-01-28
Refactor the code that does vDSO checks into a helper function.

sunfishcode:
Move all the safety checks out of `init_from_sysinfo_ehdr` and into a
helper function. This makes `init_from_sysinfo_ehdr` less huge and
clarifies the split between the safety checks and the actual vDSO
parsing code.


cgwalters (approved):
Superficial LGTM

PR #203, opened 2022-01-28
Outline assembly code refactoring

sunfishcode:
This PR contains three separate patches to rustix's Rust code wrapping its outline assembly code: two minor code simplifications, and expanded comments.


cgwalters (approved):
Very superficial LGTM

PR #206, opened 2022-01-31
Support Linux 2.6.32

sunfishcode:
Rust supports Linux on x86-64 back to Linux 2.6.32, so in theory rustix should as well. In practice, I now have a Linux 2.6.32 box to test on. This PR contains several patches needed to get it to build and pass the tests.

I'm hoping that the `dup2_to_replace_stdio` test can be a pattern we can follow for writing other tests as well. Put the test logic into an example, and then run the example from the test. The main advantage of this is that we can manipulate the host process in ways that would otherwise break the test harness. And maybe the example will be useful as an actual example too.

This doesn't yet set up anything in CI, because I don't currently know how to test things like this in CI. But hopefully it's a step in the right direction.


examples/dup2_to_replace_stdio.rs
```diff
@@ -0,0 +1,40 @@
...
+    use io_lifetimes::AsFilelike;
+    use rustix::io::{dup2, pipe};
+    use std::io::{BufRead, BufReader, Write};
+    use std::mem::forget;
+
+    let (reader, writer) = pipe().unwrap();
+    let (stdin, stdout) = unsafe { (rustix::io::take_stdin(), rustix::io::take_stdout()) };
+    dup2(&reader, &stdin).unwrap();
+    dup2(&writer, &stdout).unwrap();
+    forget(stdin);
```
cgwalters, line 15: Hmm, why does `dup2` take `&OwnedFd` instead of consuming an `OwnedFd`?

cgwalters (approved):
One unrelated comment but otherwise seems sane to me.

examples/dup2_to_replace_stdio.rs
sunfishcode: It's a good question :-). It could take an `OwnedFd` and return an `OwnedFd` with the same value, and it'd be almost exactly the same. But there are a few situations where taking a `&OwnedFd` works where consuming a `OwnedFd` doesn't. For example:

```rust
struct MyType {
    fd: OwnedFd
}
impl MyType {
    fn frob(&mut self) -> io::Result<()> {
        self.fd = dup2(&something, self.fd)
    }
}
```
This doesn't work, because `self` is borrowed, so the type checker won't allow `self.fd` to be consumed. But taking an `&OwnedFd` means you can do `dup2(&something, &self.fd)`, which does work.

cgwalters: Couldn't this be fixed by adding something like
`fn dup2_to(src: &Fd, dest: &mut OwnedFd) -> io::Result<()>` ?

sunfishcode: Do you mean in addition to having a `dup2` that takes an `OwnedFd` by value? We could do that, but it's unclear what advantage that would have over the current `dup2` API.

cgwalters: Right, in addition.  Dunno, not a big deal right now.  I know this is just a test case.  But `mem::forget()` to me is usually a big red flag and at a minimum deserves a comment.

sunfishcode: I've opened https://github.com/bytecodealliance/rustix/pull/207 to add comments.

The `mem::forget` would still be needed, even with a `dup2` function that consumes the `OwnedFd`, because such a function would return a new `OwnedFd`, and then we'd have to `mem::forget` that here.

I think the real tricky part of this example isn't the `dup2` per se, but the `stdin` and `stdout`. We don't want the behavior of tracking their ownership and closing them when they're dropped. We just want to install them as new ambient file descriptors for `println` et al to use.

PR #212, opened 2022-02-03
Add a CI script to test rustix's reverse dependencies.

sunfishcode:
As suggested [here], add a CI script to test some of rustix's reverse
dependencies. For now, this runs on `workflow_dispatch` rather than all
PRs, as it takes a lot of time, and there are still some parts of the
test script that need fixing.

[here]: https://github.com/bytecodealliance/rustix/pull/197#issuecomment-1026926506

### DIFF
--- a/.github/workflows/test-users.yml
+++ b/.github/workflows/test-users.yml
@@ -22,9 +22,11 @@ jobs:
     - uses: ./.github/actions/install-rust
       with:
         toolchain: ${{ matrix.rust }}
-
+    - uses: actions/checkout@v2
+      with:
+        repository: sunfishcode/async-io
+        path: async-io
     - run: sed -i 's/^version = "\([[:digit:].]*\)-alpha.0"$/version = "\1"/' Cargo.toml
-    - run: git clone https://github.com/sunfishcode/async-io
     - run: cd async-io && echo '[patch.crates-io]' >> Cargo.toml
     - run: cd async-io && echo 'rustix = { path = ".." }' >> Cargo.toml
     - run: cd async-io && cargo test
@@ -46,9 +48,11 @@ jobs:
     - uses: ./.github/actions/install-rust
       with:
         toolchain: ${{ matrix.rust }}
-
+    - uses: actions/checkout@v2
+      with:
+        repository: bytecodealliance/cap-std
+        path: cap-std
     - run: sed -i 's/^version = "\([[:digit:].]*\)-alpha.0"$/version = "\1"/' Cargo.toml
-    - run: git clone https://github.com/bytecodealliance/cap-std
     - run: cd cap-std && echo '[patch.crates-io]' >> Cargo.toml
     - run: cd cap-std && echo 'rustix = { path = ".." }' >> Cargo.toml
     - run: cd cap-std && cargo test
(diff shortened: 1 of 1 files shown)


.github/workflows/test-users.yml
```diff
@@ -0,0 +1,332 @@
...
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - uses: ./.github/actions/install-rust
+      with:
+        toolchain: ${{ matrix.rust }}
+
+    - run: sed -i 's/^version = "\([[:digit:].]*\)-alpha.0"$/version = "\1"/' Cargo.toml
+    - run: git clone https://github.com/sunfishcode/async-io
```
cgwalters, line 27: I tend to use `--depth=1` for these types of things; doesn't matter with this repo right now, but does start to matter with much longer lived ones.  But even higher level I think it makes sense to use the `checkout` action as is [done here](https://github.com/containers/containers-image-proxy-rs/blob/3d83477f527fc8791da699ddce5f2605aeb20beb/.github/workflows/ci.yaml#L54) which in theory I think can e.g. do caching behind the scenes (and might do so).

.github/workflows/test-users.yml
```diff
@@ -0,0 +1,332 @@
...
+            rust: stable
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - uses: ./.github/actions/install-rust
+      with:
+        toolchain: ${{ matrix.rust }}
+
+    - run: sed -i 's/^version = "\([[:digit:].]*\)-alpha.0"$/version = "\1"/' Cargo.toml
```
cgwalters, line 50: Related to this, personally I find the prior `cargo release` default of bumping to an alpha version causes much more pain than problems it solves.  I did this exact thing when I was trying to test a rustix patch locally too.

This is https://github.com/crate-ci/cargo-release/issues/82

.github/workflows/test-users.yml
```diff
@@ -0,0 +1,332 @@
...
+    - run: git clone https://github.com/cgwalters/ostree-ext
```
cgwalters, line 251: Please use https://github.com/ostreedev/ostree-rs-ext/

.github/workflows/test-users.yml
```diff
@@ -0,0 +1,332 @@
...
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - uses: ./.github/actions/install-rust
+      with:
+        toolchain: ${{ matrix.rust }}
+    - run: git clone https://github.com/sunfishcode/rust
+    - run: cd rust && git checkout rustix
+    - run: cd rust && sed -i 's/\<git = "https:\/\/github\.com\/bytecodealliance\/rustix", branch = "rustc-dep-of-std"/path = "..\/..\/.."/' library/std/Cargo.toml
```
cgwalters, line 328: Neat.

cgwalters (commented):
Skimmed this, LGTM aside from some nits.

A lot of duplication here, I wonder if it could be made a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows#example-reusable-workflow)?  I've been meaning to dig into that.

(My daily life involves at least 5 different CI systems, GH actions is only a portion of the workload and so I am certainly not an expert at it)

.github/workflows/test-users.yml
sunfishcode: Makes sense; thanks for that link. It's counter-intuitive to me, but it seems like the way the tooling wants us to go here, so I'll stop using --dev-version.

.github/workflows/test-users.yml
sunfishcode: Done.

.github/workflows/test-users.yml
sunfishcode: Good idea, using the `checkout` action seems straightforward, so I've now changed the script to use that.

PR #213, opened 2022-02-03
Fix several issues in the test-users script.

sunfishcode:
- In the Rust build, use a downloaded LLVM to avoid needing the dependencies
   needed to build LLVM.
 - Fix the name of the ostree-rs-ext repository.
 - Install libostree-dev for the ostree-rs-ext build.
 - Launch a dbus-daemon for the dbus-rs tests to use.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -193,7 +193,8 @@ jobs:
             target: aarch64-unknown-linux-gnu
             gcc_package: gcc-aarch64-linux-gnu
             gcc: aarch64-linux-gnu-gcc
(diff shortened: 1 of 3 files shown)


sunfishcode:
One slippery part about these `[patch.crates-io]` lines is that it looks like the crate's rustix dependency version has to exactly match the rustix version. Several users are depending on "0.33.0", but the rustix crate is now at "0.33.2", which causes it to silently ignore the patch. I may try some more sed scripts, but it's awkward.

sunfishcode:
There's more to do to get test_users.yml fully working, but this gets it closer, and it adds caching for the qemu builds, so I'm going to merge this and iterate in followup PRs.

PR #216, opened 2022-02-04
Powerpc64le linux_raw support.

### LINKED ISSUE #188: powerpc64le linux_raw support
Unless someone else is already working on it, I'm thinking about adding powerpc64le support in the linux_raw backend soon.

sunfishcode:
This mostly follows other architectures, however Rust's inline asm is currently
considered experimental on PowerPC, so this uses outline asm by default even on
Rust nightly.

Fixes #188.


cgwalters (approved):
Only skimmed this; I don't speak ppc64 asm.  But it seems sane to me.

PR #223, opened 2022-02-12
mips64le and mipsel linux_raw support

sunfishcode:
This patch series add mips64el and mipsel support in linux_raw, and fixes two test bugs that it uncovered.

mips is similar to powerpc64 in most ways that matter here, with the most special thing being that 32-bit mips requires a 7-argument syscall.


Cargo.toml
```diff
@@ -103,3 +103,7 @@ targets = [
     "x86_64-apple-darwin",
     "x86_64-pc-windows-msvc",
 ]
+
+# Patch the rustix dependency in is-terminal.
+[patch.crates-io]
+rustix = { path = "." }
```
cgwalters, line 109: Can `Cargo.toml` uploaded to crates.io use `patch` though by default?

sunfishcode: Oh, good point.

Actually, I think using is-terminal here is the mistake. We're effectively testing against our own implementation. I'll pull this part of the patch, and submit a different PR to fix this.

cgwalters: Yeah, I think we probably need to do a reverse-dependency style test for is-terminal or so.

PR #224, opened 2022-02-12
Misc powerpc64le cleanups

sunfishcode:
This contains several minor cleanups that I noticed while working on mips. The change from `out` to `lateout` isn't necessary, but is here to keep the code consistent with other platforms.


cgwalters (approved):
Superficial LGTM

PR #225, opened 2022-02-12
Add `#[inline]` attributes to small functions.

sunfishcode:
These functions inline into a small number of instructions, so mark them
`#[inline]`.


sunfishcode:
That's a good point. rustix doesn't use `inline(always)`, but I know that I use `inline` in places that intuitively want top-down inlining rather than bottom-up. I've removed such `inline`s from this PR, and I'll keep an eye out for more of these.

Issue #215, opened 2022-02-04
Support opting-in to libc even on x86_64 etc.


cgwalters:
As you know I like a lot of the rustix design, the I/O safety, etc.  But the "no-libc" path from my use cases I feel brings more risk than reward.  One blunt way to look at this is my employer also employs people to work on glibc.  We need to support 4 different architectures.  The current status quo of using direct syscalls on e.g. x86_64 but libc on ppc64le means that suddenly my software could have ppc64le specific bugs.  (And we already have problems with CI capacity on that arch)

Basically, WDYT about exposing `use-libc` which would turn on `rustix_use_libc`?  (Or is this possible today?)

sunfishcode:
It's possible today, by setting the `RUSTFLAGS` environment variable to `--cfg=rustix_use_libc` when building.

Also, #216 is a PR adding support for direct syscalls on ppc64le.

cgwalters:
> It's possible today, by setting the RUSTFLAGS environment variable to --cfg=rustix_use_libc when building.

Yeah, but I want this to happen by default when one does a `cargo build` on my projects.

sunfishcode:
I didn't do this initially because [cargo features are meant to be additive](https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification) and at the time I didn't know of a way to model use_libc in this way. However, I now I think I can see how this could work: we could have a `use_libc` feature, and also a `use_linux_raw` feature, and require users enable at least one of those two, and if the user enables both at the same time, we'd pick one. I think that would work.

cgwalters:
OK, thanks - I may take a swing at implementing that in the next few days.

There's obviously a much larger picture thing here about the future of rustix and the overhead of two backends, and *especially* what the default is if/when rustix is a backend of `std`.  

I've used the argument several times about Rust being more friendly to us in the "traditional Linux" space i.e. Debian/Fedora and derivatives because (unlike notably Go) it links to libc by default.  There's a lot of issues and bugs that have happened due to Go's decision, https://marcan.st/2017/12/debugging-an-evil-go-runtime-bug/ is a fun one.  But there's all the stuff around NSS modules, etc.

Anyways I won't belabor this point.  If you're good with a feature for this then I think that's great for now.

PR #227, opened 2022-02-14
Make `backend-libc` and `backend-linux-raw` features

cgwalters:
Closes: https://github.com/bytecodealliance/rustix/issues/215

Today, rustix supports `libc` and `linux-raw-sys` as backends.
The default is to use `linux-raw-sys` if available (i.e. we're
on Linux on a supported architecture).

There's some support for requiring libc via an environment variable,
but this operates outside the `cargo` native build system and hence
works less reliably for projects that want libc always.

Change things so that `backend-libc` and `backend-linux-raw` are explicit
cargo features; one of them must be set.  In this proposal, the default
remains `backend-linux-raw`.

More rationale:

I am a big fan of the rustix approach of I/O safety, and there's
a lot of very nice code in this crate.

However, I am less excited by replacing libc.  For the forseeable
future for at least most of my employer's group's usage, all
production usage of rustix will have libc linked in.

I don't want to discover that our application is broken because
some obscure system call on a less-common architecture in a corner
case is broken.

I also want to avoid the pre-compiled assembly code when vendoring.

This support for `backend-libc` helps address those concerns.

A side benefit: the `Cargo.toml` no longer needs
to replicate the architecture checking conditionals
from the build script.

---

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -47,7 +47,8 @@ fn main() {
     let rustix_linux_raw = var("CARGO_FEATURE_BACKEND_LINUX_RAW").is_ok();
 
     let backend = if !(rustix_libc || rustix_linux_raw) {
-        eprintln!("rustix: At least one of use-libc or use-linux-raw must be set");
+        // https://internals.rust-lang.org/t/clearer-display-of-build-script-errors-in-cargo/14222/5
+        eprintln!("error: At least one of use-libc or use-linux-raw must be set");
         std::process::exit(1)
     } else if rustix_libc {
         // If backend-libc is set (as it must be explicitly), then it takes precedence
--- a/build.rs
+++ b/build.rs
@@ -68,6 +68,8 @@ fn main() {
             use_feature("libc");
         }
         Backend::LinuxRaw => {
+            // In the future we could probably support `android` here too, but it looks like that
+            // needs at least a few fixes.
             let can_use_linux_raw =
                 os_name == "linux" && std::fs::metadata(&asm_name).is_ok() && can_do_raw_arch;
             if !can_use_linux_raw {
--- a/build.rs
+++ b/build.rs
@@ -6,6 +6,12 @@ use std::io::Write;
 /// The directory for out-of-line ("outline") libraries.
 const OUTLINE_PATH: &str = "src/imp/linux_raw/arch/outline";
 
+#[derive(PartialEq, Eq)]
+enum Backend {
+    Libc,
+    LinuxRaw,
+}
+
 fn main() {
     // Don't rerun this on changes other than build.rs, as we only depend on
     // the rustc version.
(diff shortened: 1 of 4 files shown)


cgwalters:
OK right, I should have tested this on a linux-but-not-raw-capable arch.  And I think I now understand that today we need to replicate the build script conditionals in `Cargo.toml` and `build.rs` because the build script can't go back and affect anything computed in `Cargo.toml` unfortunately (is that right?)

Urgau:
> OK right, I should have tested this on a linux-but-not-raw-capable arch. And I think I now understand that today we need to replicate the build script conditionals in `Cargo.toml` and `build.rs` because the build script can't go back and affect anything computed in `Cargo.toml` unfortunately (is that right?)

Yes, I think that right.

> However, now crate users who want to explicitly opt in to libc and avoiding the raw backend can enable the `use-libc` feature.

This is problematic, as cargo features are **additive** so it must be possible to have `use-libc` and `use-linux-raw-sys` at the same time or that's not currently possible and I don't think that something that will ever be.
That's the reason why the `Cargo.toml` have `cfg` deps, you can't have non-additive features.

cgwalters:
> This is problematic, as cargo features are additive so it must be possible to have use-libc and use-linux-raw-sys at the same time or that's not currently possible and I don't think that something that will ever be.

This was mentioned in https://github.com/bytecodealliance/rustix/issues/215#issuecomment-1030451315

I think what I missed though is the "require users to pick" bit of the comment.  Done now.

cgwalters:
OK, I see that we now need to bifurcate the `std` and `rustc-dep-of-std ` features into e.g. `std-libc` and `std-linux-raw-sys` because having the `std` feature defined as
```
std = ["io-lifetimes", "linux-raw-sys/std"]
```

means that `linux-raw-sys` is implicitly enabled right now too.

cgwalters:
BTW only tangentially related to this, it's confusing for this having `is-terminal` in `dev-dependencies` because it pulls in rustix too - I was misled by the dependency chain there.

build.rs
```diff
@@ -23,40 +29,63 @@ fn main() {
...
+    match backend {
+        Backend::Libc => {
+            use_feature("libc");
+        }
+        Backend::LinuxRaw => {
+            let can_use_linux_raw =
+                os_name == "linux" && std::fs::metadata(&asm_name).is_ok() && can_do_raw_arch;
+            if !can_use_linux_raw {
+                eprintln!(
+                    "linux-raw-sys feature is set, but not supported on OS {} and architecture {}",
```
sunfishcode, line 69: I'd prefer to name this feature "linux-raw", as the "-sys" is kind of an implementation detail.

sunfishcode (commented):
Yes, rustix has always supported using the libc backend, and it even has CI tests for this on platforms where it's not on by default. As we discussed, it currently uses a rustc flag instead of cargo feature flags because I didn't know how to do this with cargo feature flags at the time. I appreciate the patch to implement this with cargo feature flags!

(And as an aside, use of the prebuilt libraries are already controlled by a cargo feature flag; build with --features=cc or similar to build everything from source. And even this will go away once inline asm is stablilized.)

At the same time, I am building the linux_raw backend to be usable, even if not everyone will chose to use it. I do a lot of testing today, and expect to add more over time.

Most of this is testing effectively focuses on correctly converting arguments, return values, and pointed-to structs. And that's where the greatest potential for bugs is. But it doesn't test is all the environments: different distros, different container setups, different security settings, different toolchain options, and so on. These don't change the arguments, return values, and pointed-to structs, but they do change things like the linker, the debugger, the backtrace mechanism, compilation flags, the set of supported syscalls, and so on.

Also, some of the testing currently uses linux-user qemu, and while it's very good, it does differ from running regular Linux on hardware in some ways.

So, my preference is to leave the default as linux-raw, to help confirm it works on more than just my own computers and CI. If there is anything about rustix or linux_raw which makes it unsuitable in some environment, I'd like to learn about it.

cgwalters:
OK, the features are now `backend-libc` and `backend-linux-raw`, and I reworked to retain the default of `backend-linux-raw`.  Also tweaked things so that if both are set, then we choose `backend-libc`, but we do print a warning.

sunfishcode:
> Hmm. Was it intentional today that android != linux here?

Yeah; in theory android could be supported in the same way as regular linux, but it requires additional toolchain packages to be installed on the host, and I haven't sorted that out yet.

cgwalters:
OK, I don't understand why non-linux (e.g. freebsd/fuchsia) builds aren't pulling in libc/errno.

sunfishcode:
I think what's needed here is to have two dependency lines on errno+libc, one optional for platforms where it's optional, and one not, something like this:

```toml
[target.'cfg(target_os = "linux")'.dependencies]
libc = { version = "...", optional = true }
errno = { version = "...", optional = true }

[target.'cfg(not(target_os = "linux"))'.dependencies]
libc = { "version = "..." }
errno = { "version = "..." }
```

cgwalters:
Hmm.  But shouldn't the `backend-libc = ["errno", "libc"]` feature mean that when we do `cargo build --features=backend-libc` those crate dependencies are enabled?

cgwalters:
@sunfishcode The io-lifetimes error on wasi looks like a regression from https://github.com/sunfishcode/io-lifetimes/commit/730d9edd160336e2a18ae9b3720718ec3983369e I think.  If I do
```
$ cargo update -p io-lifetimes --precise 0.5.1
    Updating crates.io index
    Updating io-lifetimes v0.5.2 -> v0.5.1
$
```
then it builds here.

I'm guessing you know this offhand, but if you're busy I can try to dig in.

sunfishcode:
Yes, I can take a look here, later today after some meetings.

cgwalters:
Superceded by https://github.com/bytecodealliance/rustix/pull/229

PR #229, opened 2022-02-16
Switch to a single feature, `use-libc`, for the libc backend.

sunfishcode:
This is based on #227. As I was looking into it, the approach of using multiple backend-* features was turning out to get much more complex than I had initially expected.

To hopefully simplify things, for rustix and users, I'm now exploring just having a single "use-libc" feature, which, when enabled, selects the libc backend. This way, users don't need to use `default-features = false` and remember to enable `std`. I'm hoping this is still ok with the idea of cargo features being additive, because the backend in use is mostly not visible in the API. There are a few places where some types differ, but users of rustix need to be prepared for those types to vary between platforms already.

This also makes it easier to re-introduce the --cfg=rustix_use_libc, allowing users to use either option: the cargo feature flag, or the cfg flag. The cfg flag is useful for end users compiling crates that don't directly depend on rustix.

### DIFF
--- a/tests/backends.rs
+++ b/tests/backends.rs
@@ -55,6 +55,9 @@ fn test_backends() {
     );
 }
 
+/// Test whether the crate at directory `dir` has a dependency on `dependency`,
+/// setting the environment variables `envs` and unsetting the environment
+/// variables `remove_envs` when running `cargo`.
 fn has_dependency(
     dir: &str,
     envs: &[(&str, &str)],
@@ -78,5 +81,10 @@ fn has_dependency(
 
     let child = command.output().unwrap();
 
+    // `cargo tree --invert=foo` can fail in two different ways: it exits with
+    // a non-zero status if the dependency is not present in the Cargo.toml
+    // cofiguration, and it exists with a zero status and prints nothing if the
+    // dependency is present but optional and not enabled. So we check for
+    // both here.
     child.status.success() && !child.stdout.is_empty()
 }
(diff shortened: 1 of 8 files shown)


tests/backends.rs
```diff
@@ -0,0 +1,82 @@
...
+    for (key, value) in envs {
+        command.env(key, value);
+    }
+    for key in remove_envs {
+        command.env_remove(key);
+    }
+
+    let child = command.output().unwrap();
+
+    child.status.success() && !child.stdout.is_empty()
```
cgwalters, line 81: My default is to propagate errors; swallowing them should IMO always deserve a comment.

In this case, any reason not to:
```
if !child.status.success() {
  panic!("Failed to invoke cargo")
}
```
at least?

cgwalters (approved):
Thanks so much for picking this up!

In general I'm OK if a maintainer pushes additional commits to my PRs too.

One minor downside of this approach is that `linux-raw-sys` will be compiled and in the dependency tree, but not used (right?).

Anyways, LGTM overall and thanks so much for picking this up!

tests/backends.rs
sunfishcode: It turns out `cargo tree` fails with a non-zero status sometimes, and with just printing nothing other times. So we have to check for both. I've now added a comment about this.

sunfishcode:
> One minor downside of this approach is that `linux-raw-sys` will be compiled and in the dependency tree, but not used (right?).

Yes, that's right.

PR #232, opened 2022-02-16
Use the new `impl AsFd for &T` support in io-lifetimes.

sunfishcode:
With rust-lang/rust#93888 now in nightly, and sunfishcode/io-lifetimes#18 now
in io-lifetimes 0.5.2, we can now simplify the `AsFd` pattern from
`fn foo<Fd: AsFd>(fd: &Fd)` to `fn foo<Fd: AsFd>(fd: Fd)`, without the `&`.

This also follows the emerging convention in std, as seen in the new
[`fchown`] function.

This means that users can now pass `BorrowedFd` values directly to
functions that expect `AsFd` arguments, without having to write an
extra `&` when passing them.

[`fchown`]: https://doc.rust-lang.org/nightly/std/os/unix/fs/fn.fchown.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ cc = { version = "1.0.68", optional = true }
 [dependencies]
 bitflags = "1.2.1"
 itoa = { version = "1.0.1", default-features = false, optional = true }
-io-lifetimes = { version = "0.5.1", default-features = false, optional = true }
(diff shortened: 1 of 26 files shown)


cgwalters (approved):
Oh nice, this looks a lot better!  I was confused by the need to `&` a `BorrowedFd` indeed.

PR #233, opened 2022-02-16
Avoid redundant instructions in x86 inline asm

sunfishcode:
This is inspired by an idea from @m-ou-se in #project-inline-asm in the Rust zulip. We can't use esi as an explicit inline asm operand, but we can request a register and test whether we got esi using `.ifnes`. When we do get esi, avoid swapping it into place.

And, eliminate a redundant push/pop in `indirect_syscall5`.

While here, add some comments and tidy up the arch/inline/mod.rs file.


Urgau:
There are some ongoing debate in T-lang about what subsets of assembly instructions are guaranteed to be available and some that they want to [disallow](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/asm!.20and.20backends/near/270250624) (because a different codegen or implementation might not have it), this includes the `.if`/`.ifnes`/... operands.

sunfishcode:
> There are some ongoing debate in T-lang about what subsets of assembly instructions are guaranteed to be available and some that they want to [disallow](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/asm!.20and.20backends/near/270250624) (because a different codegen or implementation might not have it), this includes the `.if`/`.ifnes`/... operands.

Interesting, thanks for mentioning that! Restricting the set of directives is a good idea, and I encourage T-lang to do so; [there are a lot of directives and they can interact with compilers in undocumented and undesigned ways](https://github.com/bytecodealliance/wasmtime/issues/1041#issuecomment-531574394).

If `.if`/`.endif`/etc. are disallowed, we can simply remove those directives here once nightly starts enforcing the new rules. They're a *very* minor optimization in this context.

sunfishcode:
Ah, looking at the list [here](https://github.com/rust-lang/reference/pull/1168), it appears the plan is not to enforce the restrictions, but to say that things not on the list have assembler-specific behavior. Accordingly, I've now removed the optimization that uses `.ifnes` from this PR, to avoid depending on what will be considered assembler-specific behavior.

PR #235, opened 2022-02-16
Use `weak!`/`syscall! to support symbols missing from older glibc.

sunfishcode:
`gettid`, `getrandom`, `getauxval`, and `prlimit64` are not supported in
the oldest versions of glibc supported by Rust. Use the `weak!` and
`syscall!` macros to call these, to avoid linker errors.

### DIFF
--- a/src/imp/libc/rand/syscalls.rs
+++ b/src/imp/libc/rand/syscalls.rs
@@ -8,7 +8,7 @@ use {super::super::c, super::super::conv::ret_ssize_t, crate::io, crate::rand::G
 #[cfg(target_os = "linux")]
 pub(crate) fn getrandom(buf: &mut [u8], flags: GetRandomFlags) -> io::Result<usize> {
     // `getrandom` wasn't supported in glibc until 2.25.
-    syscall! {
+    weak_or_syscall! {
         fn getrandom(buf: *mut c::c_void, buflen: c::size_t, flags: c::c_uint) via SYS_getrandom -> c::ssize_t
     }
 
--- a/src/imp/libc/rand/syscalls.rs
+++ b/src/imp/libc/rand/syscalls.rs
@@ -7,12 +7,12 @@ use {super::super::c, super::super::conv::ret_ssize_t, crate::io, crate::rand::G
 
 #[cfg(target_os = "linux")]
 pub(crate) fn getrandom(buf: &mut [u8], flags: GetRandomFlags) -> io::Result<usize> {
-    let nread = unsafe {
-        ret_ssize_t(c::getrandom(
-            buf.as_mut_ptr().cast(),
-            buf.len(),
-            flags.bits(),
-        ))?
-    };
+    // `getrandom` wasn't supported in glibc until 2.25.
+    syscall! {
+        fn getrandom(buf: *mut c::c_void, buflen: c::size_t, flags: c::c_uint) via SYS_getrandom -> c::ssize_t
+    }
+
+    let nread =
+        unsafe { ret_ssize_t(getrandom(buf.as_mut_ptr().cast(), buf.len(), flags.bits()))? };
     Ok(nread as usize)
 }
--- a/src/imp/libc/fs/syscalls.rs
+++ b/src/imp/libc/fs/syscalls.rs
@@ -363,7 +363,7 @@ pub(crate) fn fclonefileat(
             dst_dirfd: BorrowedFd<'_>,
             dst: *const c::c_char,
             flags: c::c_int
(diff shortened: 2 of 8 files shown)


src/imp/libc/rand/syscalls.rs
```diff
@@ -7,12 +7,12 @@ use {super::super::c, super::super::conv::ret_ssize_t, crate::io, crate::rand::G
...
-            flags.bits(),
-        ))?
-    };
+    // `getrandom` wasn't supported in glibc until 2.25.
+    syscall! {
+        fn getrandom(buf: *mut c::c_void, buflen: c::size_t, flags: c::c_uint) via SYS_getrandom -> c::ssize_t
+    }
+
+    let nread =
+        unsafe { ret_ssize_t(getrandom(buf.as_mut_ptr().cast(), buf.len(), flags.bits()))? };
```
cgwalters, line 16: Am I understanding this code correctly in that we now always invoke the system call?  Maybe I'm misreading the macro implementations, but I am not seeing the behavior of `syscall!` trying the weak pointer and falling back to the system call.

sunfishcode: Yes, that's right; this works by always using `libc::syscall`. Do you think it would be better to try the libc wrapper first and fall back?

cgwalters: Of course in that in any system that is including non-Rust C/C++ code (like my case) unless one goes to extreme lengths on that side too, the resulting binary is  pulling in a hard dependency on specific glibc versions anyways.  So for me the dlsym dance is unnecessary.   I don't think this really matters, it almost certainly isn't worth trying to expand the matrix here to be (hard requirement, weak requirement, direct syscall) unless performance needs drive it.

sunfishcode: Yeah, and the `dlsym` code is just a workaround for `#[weak_extern]` not being stable yet in Rust. When they are, we'll switch `weak!` and `weakcall!` to use that instead of `dlsym`.

Thinking about this more though, I can see the argument that we should try to use the libc wrappers; that way people expecting to be able to replace them with `LD_PRELOAD`, or break on them in gdb, can do so. I've now added a `weak_or_syscall!` macro which implements this; does this look reasonable?

cgwalters:
> Thinking about this more though, I can see the argument that we should try to use the libc wrappers; that way people expecting to be able to replace them with LD_PRELOAD, or break on them in gdb, can do so.

Well it's not just that, it's also reducing the matrix for possible bugs.  One corner case I know of is e.g. [clone()'s per-architecture quirks](https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/src/basic/raw-clone.h#L37).  That's papered over by glibc wrapper (actually, I am not sure why systemd isn't using it today).

By using the libc wrapper, the bug blast radius is reduced to only people who have an older glibc - and if they have an older glibc, it's also likely (containerization aside) that they have an older kernel that doesn't support the system call anyways.

Now, obviously all this is needed in the linux-raw backend, and...hmm, actually, is the implementation of `fork()` buggy there?

sunfishcode:
> > Thinking about this more though, I can see the argument that we should try to use the libc wrappers; that way people expecting to be able to replace them with LD_PRELOAD, or break on them in gdb, can do so.
> 
> Well it's not just that, it's also reducing the matrix for possible bugs. One corner case I know of is e.g. [clone()'s per-architecture quirks](https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/src/basic/raw-clone.h#L37). That's papered over by glibc wrapper (actually, I am not sure why systemd isn't using it today).
> 
> By using the libc wrapper, the bug blast radius is reduced to only people who have an older glibc - and if they have an older glibc, it's also likely (containerization aside) that they have an older kernel that doesn't support the system call anyways.

Makes sense.

> Now, obviously all this is needed in the linux-raw backend, and...hmm, actually, is the implementation of `fork()` buggy there?

Rustix doesn't have linux-raw support on sparc, s390x, or cris, so it doesn't seem to have the issue in the link above, at least.

It also happens that rustix's `fork` is `doc(hidden)` and in its current form will probably never be intended for even rustix's low-level users to use. It's `unsafe` of course, but even attempting to document what users must do to use it safely is enormously complex. It already has a 77-line comment, and odds are it's not covering everything yet :-}. `fork`/`clone` is a very different kind of thing than regular I/O syscalls, so rustix's min strategy of focusing on safe wrappers that don't otherwise change the abstraction level doesn't work.

PR #265, opened 2022-03-31
Preserve pointer provenance in system call arguments and return values.

sunfishcode:
Make `ArgReg` and `RetReg` hold Linux system call argument and return values
in a `*mut ()` rather than a `usize`. This allows us to avoid casting
pointers into `usize` before passing them into system calls.

This allows the linux_raw backend to preserve pointer provenance all the
way up to the inline asm making the system calls (or, when compiled
without support for inline asm, all the way up to the FFI to the outline
assembly code).


cgwalters (approved):
High level LGTM, one optional.

sunfishcode:
> I guess that all of this is required because of the `int | ptr -> ptr` rule ?

Yes. When it's carrying a logical integer value, either an int or ptr type will work, but when it's carrying a pointer, it needs to use a pointer type to preserve the provenance.

And to be sure, none of this is *required* by Rust. But strict provenance does have a lot of resonance with what rustix is doing, so it seems like an interesting experiment for rustix to participate in.

PR #268, opened 2022-04-01
Fall back to `libc::syscall` for `openat` if needed

cgwalters:
This works around https://sourceware.org/bugzilla/show_bug.cgi?id=17523
Where the `openat` wrapper is buggy if `O_TMPFILE` is provided
on (now quite old) glibc versions.

Use the presence of the `getrandom` symbol to detect if we have
a fixed glibc or not.

Will be used by https://github.com/bytecodealliance/cap-std/pull/239
which is wiring up use of `O_TMPFILE` in cap-std.


cgwalters:
(Draft since I haven't fully tested this, and am getting an unused variable warning I don't understand - but I want to know if this is on the right track)

src/imp/libc/fs/syscalls.rs
```diff
@@ -123,6 +123,28 @@ pub(crate) fn openat(
     oflags: OFlags,
     mode: Mode,
 ) -> io::Result<OwnedFd> {
+    // Work around https://sourceware.org/bugzilla/show_bug.cgi?id=17523
+    // If we have a getrandom() symbol, then we know we have a glibc that
+    // has fixed the bug.  This can be removed when Rust bumps its minimum
+    // glibc version.
+    if !crate::imp::is_atleast_glibc_2_25() {
```
cgwalters, line 130: We could also skip this check if `O_TMPFILE` isn't in `oflags` perhaps?

sunfishcode: Yes, that would be good to do.

src/imp/libc/rand/syscalls.rs
```diff
@@ -5,6 +5,9 @@
 #[cfg(target_os = "linux")]
 use {super::super::c, super::super::conv::ret_ssize_t, crate::io, crate::rand::GetRandomFlags};
 
+// A duplicate weak symbol because we use this to detect the glibc version.
+weak! { pub(crate) fn libc_getrandom_weak(*mut c::c_void, c::size_t, c::c_uint) -> c::ssize_t }
```
sunfishcode, line 9: Can this macro invocation be placed inside the body of `is_atleast_glibc_2_25`, since we don't use it for actually calling `getrandom`?

sunfishcode:
Yes, overall this looks reasonable.

cgwalters:
OK, I couldn't figure out the unused `dirfd` warning in using `syscall!` so I ended up inlining that.

But that aside...this is now tested!  CI here and my main dev environment (recent linux+glibc) covers the "happy path".  But to test the fallback path, I set up a rhel6 container (`registry.access.redhat.com/rhel6/rhel:6.10`) which definitely has an old glibc `glibc-2.12-1.192.el6.x86_64` (that I believe has the bug, though I didn't verify the bugfix wasn't backported) and using my host's kernel (`5.16.14-200.fc35.x86_64`) which does support `O_TMPFILE`.

With a strategic use of `dbg!` I saw that the test suite was entering the fallback path.

src/imp/libc/fs/syscalls.rs
```diff
@@ -116,13 +116,41 @@ use {
...
 #[cfg(not(target_os = "redox"))]
 pub(crate) fn openat(
     dirfd: BorrowedFd<'_>,
     path: &ZStr,
     oflags: OFlags,
     mode: Mode,
 ) -> io::Result<OwnedFd> {
+    // Work around https://sourceware.org/bugzilla/show_bug.cgi?id=17523
+    // Basically old glibc versions don't handle O_TMPFILE correctly.
+    #[cfg(any(target_os = "android", target_os = "linux"))]
```
cgwalters, line 150: (I guess since we know android doesn't use glibc, we could just make this `target_os = "linux"`?)

cgwalters:
Fixed the missing `#[cfg]` in the `openat` mod for tests.

(What is up with windows-2016?  Some sort of GH actions scheduling/availability problem for that OS?)

src/imp/libc/fs/syscalls.rs
sunfishcode: Would `target_env = "gnu"` work here? That should properly exclude this code for android, and musl too, which presumably doesn't have this bug.

sunfishcode:
For anyone following along, windows-2016 is [now removed](https://github.com/bytecodealliance/rustix/pull/269).

src/imp/libc/fs/syscalls.rs
cgwalters: Makes sense.  Pushed that change...and then realized after I did so that it'd likely apply to `x86_64-pc-windows-gnu` and indeed, that build now fails)

So, pushed another fix to also require `unix`.

sunfishcode:
Looks good, thanks!

PR #271, opened 2022-04-05
Fix a y2038 bug in `statat` on 32-bit platforms.

sunfishcode:
Linux's `stat64` on 32-bit platforms has 32-bit `time_t` fields for the
timestamps. To fix this, use a custom stat structure for linux_raw on
32-bit platforms, and populate it using `statx` when available.


cgwalters (approved):
(Superficial) LGTM

Issue #250, opened 2022-03-08
Put net, time, and other parts of the API behind feature flags


sunfishcode:
To reduce compile times for users, and to be able to disable the auxv initialization for users that don't need it, rustix should have feature flags such as "net" for enabling `rustix::net`, and similar flags for "time", "process" and others.

PR #272, opened 2022-04-05
Refactoring in preparation for introducing feature flags

sunfishcode:
This is a series of refactoring patches that begins preparing for #250, putting the public modules behind feature flags.

The first patch here moves implementation details out of mod.rs files and into underlying source files. This way it's easier to enable or disable features by enabling or disabling the `mod` declarations, as there aren't as many implementation details scattered around.

The next two patches make rustix use its `c` module more consistently, which reduces differences between the libc and linux_raw backends and in the future should make it easier to factor out more of the redundancy between the two. And it eliminates a lot of clutter too.


cgwalters (approved):
> And, it requires teaching the `syscall` macro how to pass `BorrowedFd`
values to `syscall` calls.

Ohhh, that was the bug I was hitting with `dirfd` in my previous PR.  Will look at switching back to `syscall!` after this.

I only skimmed this mostly, but LGTM.

PR #280, opened 2022-04-12
Use `c_uint` rather than `bool` in `getsockopt` calls.

sunfishcode:
`getsockopt` and `setsockopt` represent boolean values as integers, not as
`bool` types. The code already handled this for `setsockopt`, but not
for `getsockopt`. Handle it for `getsockopt`, and add assertions to catch
accidental uses of `bool.

### DIFF
--- a/src/imp/libc/net/syscalls.rs
+++ b/src/imp/libc/net/syscalls.rs
@@ -400,9 +400,15 @@ pub(crate) mod sockopt {
     #[inline]
     fn getsockopt<T>(fd: BorrowedFd<'_>, level: i32, optname: i32) -> io::Result<T> {
         use super::*;
+
+        let mut optlen = core::mem::size_of::<T>().try_into().unwrap();
+        debug_assert!(
+            optlen >= 4,
+            "Socket APIs don't ever use `bool` directly"
+        );
+
         unsafe {
             let mut value = MaybeUninit::<T>::uninit();
-            let mut optlen = core::mem::size_of::<T>().try_into().unwrap();
             ret(c::getsockopt(
                 borrowed_fd(fd),
                 level,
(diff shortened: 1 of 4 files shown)


src/imp/libc/net/syscalls.rs
```diff
@@ -400,9 +400,15 @@ pub(crate) mod sockopt {
     #[inline]
     fn getsockopt<T>(fd: BorrowedFd<'_>, level: i32, optname: i32) -> io::Result<T> {
```
cgwalters, line 401: I think I'd say we shouldn't support any Rust type for `T` here but rather have `getsockopt_int` and `getsockopt_ptr` or so?

cgwalters (approved):
Looks correct to me as is.

src/imp/libc/net/syscalls.rs
sunfishcode: The `T` could also be types like `struct linger` or `struct timespec`, or typedefs which may be 32-bit or 64-bit depending on the platform, so it's fairly handy to make it generic.

[Rust's standard library does this too](https://github.com/rust-lang/rust/blob/master/library/std/src/sys_common/net.rs#L79). In looking that up, I noticed that Rust uses `T: Copy`, which should help catch some kinds of misuse, so I've added that here too.

cgwalters: It'd be nice if there was some way to express though `T: #[repr(C)]` or so.  Hmm, seems like e.g. [Layout](https://doc.rust-lang.org/std/alloc/struct.Layout.html) could have something like a `const fn is_c()` or something?

Anyways, still LGTM.

cgwalters:
Also, if I understand correctly, this isn't just an endianness bug, we're potentially doing writes to the stack outside of the allocated space for the bool right?

sunfishcode:
It passes `size_of::<T>()` as the initial size of the buffer, and `getsockopt` truncates the value to fit that size. In POSIX's words, "If the size of the option value  is greater than option_len, the value stored in the object pointed to by the option_value argument shall be silently truncated.". When `T` is bool, the buffer size is 1, so `getsockopt` only writes 1 byte. The value of that byte is determined by endianness. So I think this is just an endianness issue.

sunfishcode:
Also, immediately after the `getsockopt` call is an `assert_eq` (and not a `debug_assert_eq`, so it's on even in release builds) that the returned `optlen` is equal to `sizeof::<T>()`. So if this ever did somehow write beyond the end of the buffer, we'd likely find out about it quickly.

PR #304, opened 2022-05-02
Replace many conversion functions with `From` impls.

sunfishcode:
Replace many of the `conv` functions for converting types into `ArgReg`
values with `impl From<T> for ArgReg` impls. This requires a little more
boilerplate in the inline asm code to call `.into()` on all the
arguments, but it makes the main syscall code much less verbose.

### DIFF
--- a/src/imp/linux_raw/arch/inline/aarch64.rs
+++ b/src/imp/linux_raw/arch/inline/aarch64.rs
@@ -21,12 +21,15 @@ pub(in crate::imp) unsafe fn syscall0_readonly(nr: SyscallNumber<'_>) -> RetReg<
 
 #[inline]
 #[must_use]
-pub(in crate::imp) unsafe fn syscall1(nr: SyscallNumber<'_>, a0: ArgReg<'_, A0>) -> RetReg<R0> {
+pub(in crate::imp) unsafe fn syscall1<'a>(
+    nr: SyscallNumber<'_>,
+    a0: impl Into<ArgReg<'a, A0>>,
+) -> RetReg<R0> {
     let r0;
     asm!(
         "svc 0",
         in("x8") nr.to_asm(),
-        inlateout("x0") a0.to_asm() => r0,
+        inlateout("x0") a0.into().to_asm() => r0,
         options(nostack, preserves_flags)
     );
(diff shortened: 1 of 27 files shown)


src/imp/linux_raw/arch/inline/aarch64.rs
```diff
@@ -21,259 +21,265 @@ pub(in crate::imp) unsafe fn syscall0_readonly(nr: SyscallNumber<'_>) -> RetReg<
...
 #[must_use]
-pub(in crate::imp) unsafe fn syscall3_readonly(
+pub(in crate::imp) unsafe fn syscall3_readonly<'a>(
     nr: SyscallNumber<'_>,
-    a0: ArgReg<'_, A0>,
-    a1: ArgReg<'_, A1>,
-    a2: ArgReg<'_, A2>,
+    a0: impl Into<ArgReg<'a, A0>>,
+    a1: impl Into<ArgReg<'a, A1>>,
+    a2: impl Into<ArgReg<'a, A2>>,
```
cgwalters, line 130: Any concerns about the implicit generics here potentially introducing multiple copies of this and other functions?  Might we need to do the "manual forward to non-generic impl" trick?  (Seems like that'd be painful here though...)

Or maybe the inline asm is small enough this doesn't matter.

cgwalters (approved):
I only gave this a very superficial skim...I think it seems sane.

src/imp/linux_raw/arch/inline/aarch64.rs
sunfishcode: Ah, that's a good point. In release builds these I expect these to be inlined. But I should look more closely at what happens in practice, and also what happens in debug builds.

cgwalters: Probably simplest to `#[inline(always)]`?

sunfishcode: It's a good idea, but thinking about this more, I think this is a good use case for a few macros. They can help factor out even more of the redundancy, including a bunch of `#cfg`'s in the syscalls.rs files, which is quite nice and appears enough for me to set aside my general dislike of macros in this case :-).

PR #312, opened 2022-05-10
Change `dup2`'s second operand from `&OwnedFd` to `AsFd`.

sunfishcode:
And similar for `dup3`.

The idea behind using `&OwnedFd` was that `dup2`'s second operand isn't like a
normal borrow. It effectively closes the old file descriptor, and
creates a new one with the same index. This could break assumptions of
classes that have an `AsFd` to allow users to do special I/O operations,
but which don't expect users can close and reopen their file descriptor
as some completely unrelated resource.

However, the existence of things like [`FilelikeView`], as well as the
`ManuallyDrop` pattern, mean that `&OwnedFd` doesn't actually prevent
users from using `dup2` on a `BorrowedFd`. Even using `&mut OwnedFd`
wouldn't be sufficient, because one can obtain mutable `FilelikeView`
objects, in order to use them with functions like [`std::io::Read::read`].

So instead of trying to prohibit it, change `rustix`'s stance to allow
it. Unwise use of `dup2` may cause libraries to behave strangely, but it
should not cause them to have Undefined Behavior.

[`FilelikeView`]: https://docs.rs/io-lifetimes/latest/io_lifetimes/views/struct.FilelikeView.html
[`std::io::Read::read`]: https://doc.rust-lang.org/stable/std/io/trait.Read.html#tymethod.read

### DIFF
--- a/examples/dup2_to_replace_stdio.rs
+++ b/examples/dup2_to_replace_stdio.rs
@@ -19,10 +19,6 @@ fn main() {
     let (stdin, stdout) = unsafe { (rustix::io::take_stdin(), rustix::io::take_stdout()) };
 
     // Use `dup2` to copy our new file descriptors over the stdio file descriptors.
-    //
-    // These take their second argument as an `&OwnedFd` rather than the usual
-    // `impl AsFd` because they conceptually do a `close` on the original file
-    // descriptor, which one shouldn't be able to do with just a `BorrowedFd`.
     dup2(&reader, &stdin).unwrap();
     dup2(&writer, &stdout).unwrap();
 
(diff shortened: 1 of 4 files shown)


sunfishcode:
This is a subtle issue, and I'm not yet convinced this PR is the best approach. Constructing a temporary `OwnedFd` from a `BorrowedFd` does require `unsafe`, and it was recently discovered that the view pattern [needs an `unsafe` marker trait](https://github.com/sunfishcode/io-lifetimes/pull/32). The landscape here is still evolving.

sunfishcode:
Another option here would be to use `&mut OwnedFd`.

sunfishcode:
https://github.com/bytecodealliance/rustix/pull/332 is a PR implementing the option to use `&mut OwnedFd`, which is an alternative to this PR.

PR #326, opened 2022-05-13
Match casting type to the destination iov_len's type. fix #324

### LINKED ISSUE #324: `as u64` when left hand side of assignment can be 32 bit
Cant `iov_len` be `u32`?
https://github.com/bytecodealliance/rustix/blob/23841ae92b5d1ef772bb9a91dff067f67e5ec070/src/imp/linux_raw/io/io_slice.rs#L26

AsafFisher:
fix #326

### DIFF
--- a/src/imp/linux_raw/io/io_slice.rs
+++ b/src/imp/linux_raw/io/io_slice.rs
@@ -3,10 +3,10 @@
 //! dca3f1b786efd27be3b325ed1e01e247aa589c3b.
 
 #![allow(unsafe_code)]
-
 use super::super::c;
 use core::marker::PhantomData;
 use core::slice;
+use linux_raw_sys::general::__kernel_size_t;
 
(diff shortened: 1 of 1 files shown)


sunfishcode:
The patch looks good; it looks like the CI failure is something related to the latest nightly Rust. Locally, rustup only pulls in 2022-05-11, and the patch compiles, but the CI is somehow getting 2022-05-12. We can either temporarily pin to an older nightly, or just wait until newer nightlies are available to see if they fix this.

sunfishcode:
The patch looks good; it looks like the CI failure is something related to the latest nightly Rust. Locally, rustup only pulls in 2022-05-11, and the patch compiles, but the CI is somehow getting 2022-05-12. We can either temporarily pin to an older nightly, or just wait until newer nightlies are available to see if they fix this.

AsafFisher:
Merge?

sunfishcode:
Yep, the issue seems to have been in the nightly build, which seems to be fixed now.

PR #332, opened 2022-05-17
Change `dup2`'s second operand from `&OwnedFd` to `&mut OwnedFd`.

sunfishcode:
And similar for `dup3`.

The idea behind using `&OwnedFd` is that `dup2`'s second operand isn't like a
normal borrow. It effectively closes the old file descriptor, and creates a
new one with the same index. This could break assumptions of classes that have
an `AsFd` to allow users to do special I/O operations, but which don't expect
users can close and reopen their file descriptor as some completely unrelated
resource.

However, the existence of things like [`FilelikeView`], as well as the
`ManuallyDrop` pattern, mean that `&OwnedFd` doesn't actually prevent
users from using `dup2` on a `BorrowedFd`.

With sunfishcode/io-lifetimes#32 though, `&mut OwnedFd` would be
sufficient, because it removes the `DerefMut` implementation.

So change `rustix` stance to be that `dup2` requires `&mut OwnedFd`.

This means that it's no longer possible to pass the same file descriptor
to both operands of `dup2` or `dup3` with safe Rust, which means it's not
possible to observe the difference in behavior in that case, so remove
the `dup3.rs` test.

[`FilelikeView`]: https://docs.rs/io-lifetimes/latest/io_lifetimes/views/struct.FilelikeView.html


sunfishcode:
See https://github.com/sunfishcode/io-lifetimes/pull/32#issuecomment-1130874596 for some discussion of this.

sunfishcode:
I also like how `&mut OwnedFd` is also what you need if you want to reassign an `OwnedFd` to an unrelated `OwnedFd` value, which is conceptually very similar to what `dup2` does.

sunfishcode:
This also aligns with what's written about `dup2` in [the standard library documentation](https://doc.rust-lang.org/nightly/std/os/unix/io/index.html).

And, this is sufficient to support the redirect-stdio-with-dup2 use case, which is the most common use case for `dup2`.

sunfishcode:
Merging, as there don't seem to be any major issues here. And we can always revisit this if the thinking around views or anything else changes.

PR #337, opened 2022-05-21
Remove `ZStr` and `ZString`.

### LINKED ISSUE #336: Replace `ZStr` with `CStr`
Rust has moved `CStr` into libcore and `CString` into liballoc: https://github.com/rust-lang/rust/pull/94079

As an aside, I'm still fond of the idea of using `u8` instead of `c_char` for this, because there isn't any reason to care about the signedness of C's `char` in nul-terminated string contexts&mdash;if you want to interpret the contents, you need to know the encoding, and the signedness of `char` doesn't help with that. Exposing `c_char` at this level just means that more places are exposed to gratuitous platform differences. But it's also true that in practice this issue is just a minor nuisance. Rust already had `CStr` with `c_char`, and replacing it with something that uses `u8` would create a lot of churn, so keeping it makes sense.

So now that `CStr` has moved to core in nightly, that eliminates the need for `ZStr` in rustix, so rustix should switch to using `CStr` now.

sunfishcode:
Use `CStr` and `CString` instead, now that nightly has then in core and
alloc, respectively.

Fixes #336.


sunfishcode:
This also removes the need for several scary `features`s in src/lib.rs, about 58 unsafe blocks, and almost 2000 lines of code.

cgwalters (approved):
Makes sense to me; one less string type to keep in mind is good overall.

Issue #220, opened 2022-02-10
`fchownat` and `fchown` should take `Option<Uid>` and `Option<Gid>`


sunfishcode:
In the POSIX `chown` functions the user and group IDs may be -1, indicating no change. rustix should expose this via `Option` types.

palash25:
hi @sunfishcode I am new to rust so I wanted to try solving this issue

correct me if I am wrong here, currently rustix would only allow non-negative uid and gid to be passed to the functions because the underlying types `RawUid` and `RawGid` are `u32` so you want to provide the user with the Option to pass `None` when they want to pass -1 but if I try to do this

```rust
pub(crate) fn fchown(fd: BorrowedFd<'_>, owner: Option<Uid>, group: Option<Gid>) -> io::Result<()> {
    unsafe {
        let ow = match owner {
            Some(o) => o.as_raw(),
            None => -1,
        };

        let gr = match owner {
            Some(g) => g.as_raw(),
            None => -1,
        };

        ret(syscall_readonly!(
            __NR_fchown,
            fd,
            c_uint(ow),
            c_uint(gr)
        ))
    }
}
```

it won't compile since `RawUid` is `u32`  and if I change to `pub type RawUid = i32;` and use `c_int` instead of `c_uint` other parts of the code still break because of the `i32` `u32` mismatch

PR #348, opened 2022-06-02
Use options for fchown and fchownat arguments

palash25:
Wraps Uid and Gid arguments as Options so that users can provide
None as a value in place of -1

Closes: #220


sunfishcode:
Currently it's necessary to pass --features=all-apis when running the tests, because otherwise most of the API is disabled. That's not ideal, and I'm looking for ways to avoid needing to do that, but at the moment, this is how it works.

The other thing that's going on here is that the CI tests lots of different configurations. You can probably cover most of it by testing with `--features=all-apis` and `--features=all-apis,use-libc`.

sunfishcode:
I've now opened https://github.com/bytecodealliance/rustix/pull/353 to add a CONTRIBUTING.md file to document how to run the tests, and will hopefully collect more information over time, as people ask questions :-).

sunfishcode:
This now has a few minor merge conflicts; in the upstream code `linux_raw_sys::general::PRIO_USER` was changed to `PRIO_USER`. Please let me know if I can help!

palash25:
i won't be needing help with rebasing the PR but I am stuck at a different problem @sunfishcode 

when I tried to make the same change in fchown method dedined in imp/libc/fs/syscalls.rs 

```rust
pub(crate) fn fchown(fd: BorrowedFd<'_>, owner: Option<Uid>, group: Option<Gid>) -> io::Result<()> {
    unsafe {
        let ow = match owner {
            Some(o) => o.as_raw(),
            None => -1,
        };

        let gr = match group {
            Some(g) => g.as_raw(),
            None => -1,
        };

        syscall_ret(c::syscall(
            c::SYS_fchown,
            borrowed_fd(fd),
            ow,
            gr,
        ))
    }
}

```

`cargo check --features=all-apis,use-libc` fails with the error

```
error[E0277]: the trait bound `u32: Neg` is not satisfied
   --> src/imp/libc/fs/syscalls.rs:890:21
    |
890 |             None => -1,
    |                     ^^ the trait `Neg` is not implemented for `u32`
    |
    = help: the following implementations were found:
              <&f32 as Neg>
              <&f64 as Neg>
              <&i128 as Neg>
              <&i16 as Neg>
            and 12 others
```

but i already changed the `RawUid` and `RawGid` types to `i32` when i initially made these changes for linux_raw fchown and they both use the same Uid and Gid types, but for some reason in case of libc it is not taking into account the latest i change made to the underlying type

do you know whats happening here?

sunfishcode:
I don't yet know what's going on, but one thing that looks unusual is the use of `c::syscall` here. Rustix does use `c::syscall` in some situation where the regular libc wrappers aren't usable, but we generally use the proper C wrappers, such as `c::fchown`, as much as possible.

palash25:
>  one thing that looks unusual is the use of c::syscall here.

it seems that you added `c::syscall`  here https://github.com/bytecodealliance/rustix/commit/8b8e11be2cc6585ae8d341198d8e2d2c70a55dc2#diff-67136b9423f61b76aa840df9cf5c4d32882b00c673f038a99052b5deccad3e04R542  although i can't tell the difference between that and `c::fchown` function i can't find their definitions in the codebase.

sunfishcode:
Ah, right. I should have added a comment to `fchown` similar to the [one for `fchmod`](https://github.com/bytecodealliance/rustix/blob/main/src/imp/libc/fs/syscalls.rs#L872).

Also, I didn't notice this earlier, but it appears libc's `uid_t` is actually unsigned. That's surprising, given that `fchown` uses -1 as a special value, but apparently that's what it does. So then I think it makes sense to go back to having `RawUid`/`RawGid` be unsigned, and to use the `MAX` value in place of `-1`.

palash25:
by max do you mean `u32::MAX`? if yes then how would that solve the issue? won't fchown consider the value of MAX to be a user/group ID instead of no change?

sunfishcode:
Yes, I mean `u32::MAX`. It has the same bit-pattern as `-1`. In C, a `-1` would be implicitly converted into a maximum-unsigned value, which is how `fchown` works in C. Rust doesn't implicitly convert here, so we just need to use the explicit maximum value.

palash25:
thats a neat trick, thanks for the explaination i have made the requested changes and the tests were passing locally 

forgot to run the format tool will push those changes in a minute

src/imp/libc/fs/syscalls.rs
```diff
@@ -883,20 +886,28 @@ pub(crate) fn fchmod(fd: BorrowedFd<'_>, mode: Mode) -> io::Result<()> {
 }
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
-pub(crate) fn fchown(fd: BorrowedFd<'_>, owner: Uid, group: Gid) -> io::Result<()> {
+pub(crate) fn fchown(fd: BorrowedFd<'_>, owner: Option<Uid>, group: Option<Gid>) -> io::Result<()> {
+    // Use `c::syscall` rather than `c::fchown` because some libc
+    // implementations, such as musl, add extra logic to `fchown` to emulate
+    // support for `O_PATH`, which uses `/proc` outside our control and
+    // interferes with our own use of `O_PATH`.
```
palash25, line 893: i copied the comment from chmod since you said you forgot it the last time, but I haven't looked into the musl implementation for fchown so let me know if the reason here is the same as the one for chmod or something different i would make that change too (if any)

src/utils.rs
```diff
@@ -11,3 +13,17 @@ pub(crate) const fn as_ptr<T>(t: &T) -> *const T {
...
+        None => u32::MAX,
+    };
+
+    let gr = match group {
+        Some(g) => g.as_raw(),
+        None => u32::MAX,
+    };
+
+    (ow, gr)
+}
```
palash25, line 29: forgot to add the newline, will do in the next push

src/utils.rs
```diff
@@ -11,3 +13,17 @@ pub(crate) const fn as_ptr<T>(t: &T) -> *const T {
 pub(crate) fn as_mut_ptr<T>(t: &mut T) -> *mut T {
     t
 }
+
+pub fn unwrap_fchown_args(owner: Option<Uid>, group: Option<Gid>) -> (u32, u32) {
```
palash25, line 17: feel free to suggest a better function name, i don't really like this one

palash25:
most of the nightly build failure seem to be unrelated, correct me if I am wrong

sunfishcode:
The build error is that src/utils.rs is compiled on Windows, but Windows doesn't have `Uid` and `Gid` types. Would it work to move `unwrap_fchown_args` to be a `pub(crate)` function in src/process/id.rs?

palash25:
i just tried that, moved the function to process/id.rs and made it pub(crate) but at the callsite i get an import error  

```
error[E0425]: cannot find function `unwrap_fchown_args` in module `crate::process`
   --> src/imp/linux_raw/fs/syscalls.rs:163:40
    |
163 |         let (ow, gr) = crate::process::unwrap_fchown_args(owner, group);
    |                                        ^^^^^^^^^^^^^^^^^^ not found in `crate::process`
    |
note: function `crate::process::id::unwrap_fchown_args` exists but is inaccessible
   --> src/process/id.rs:283:1
    |
283 | pub(crate) fn unwrap_fchown_args(owner: Option<Uid>, group: Option<Gid>) -> (u32, u32) {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not accessible

error[E0425]: cannot find function `unwrap_fchown_args` in module `crate::process`

```

sunfishcode:
The problem there is that `id` isn't `pub(crate)`. You can add a `pub(crate) use id::unwrap_fchown_args` to src/process/mod.rs and then use it as `crate::process::unwrap_fchown_args`.

Also, while here, "unwrap" tends to mean [something different](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.unwrap) in Rust; would it make sense to name this `translate_fchown_args` or so?

palash25:
thanks for the help @sunfishcode i have pushed the changes please take a look.

palash25:
thanks for pointing it out, added

sunfishcode:
Thanks!

PR #356, opened 2022-06-06
Update to io-lifetimes 0.7.0.

sunfishcode:
And serial_test 0.7.0.


sunfishcode:
Merging this dependency version bump so that we can test the io-lifetimes 0.7.0 release.

PR #359, opened 2022-06-08
Add `INSECURE` to `GetRandomFlags`.

sunfishcode:


sunfishcode:
This is needed by mustang, as std is now using `GRND_INSECRE`, so c-scape needs to be able to recognize it.

PR #380, opened 2022-07-11
Add getpgid

carbotaniuman:


src/process/id.rs
```diff
@@ -267,6 +267,20 @@ pub fn getppid() -> Option<Pid> {
     backend::process::syscalls::getppid()
 }
 
+/// `getpgid()`—Returns the process group ID of the given process.
```
sunfishcode, line 270: There should be a `pid` inside the parens there. The current convention is "pseudo-code for what this looks like in C".

sunfishcode (commented):
Out of curiousity, is this for the `bat` command? If so, I'm curious where the call actually comes from, and if they're using to get the current process group (the `None` case) or another process group.

carbotaniuman:
> Out of curiousity, is this for the bat command? 

Yup! The call appears to bottom out in libgit2's `rand.c` calculating a seed. I think everything basically just wants the `getpgrp` case where pid is None.

PR #385, opened 2022-07-20
Switch from `.init_array` constructors to /proc/self/auxv.

### LINKED ISSUE #382: Make it possible to selectively use `libc` for `param` mod
In the `linux_raw` backend, the `param` module must be loaded before any user-defined constructor functions. In our use case we are building a Ruby native gem, and cannot assure that our dylib is loaded early enough, so [we end up failing this assertion](https://github.com/bytecodealliance/rustix/blob/1ceb659b9f6ec35628b6dddee3a2ef881d460af4/src/backend/linux_raw/param/auxv.rs#L183).

This can be fixed by using the `use-libc`, but that also means we lose the benefits of Rust implementation (inlining, etc.) for other parts of this crate. Would it be possible to opt-in to just using libc for _the just `param` module_ (something like `use-libc-param`)?

sunfishcode:
In the linux_raw backend, switch from using a `.init_array` constructor
for obtaining the aux values to reading them from /proc/self/auxv. This avoids
problems in situation where other Rust code can run before the constructor,
potentially distrupting the `__environ` value.

Also, for the linux_raw backend, introduce a new "use-libc-auxv" feature,
which enables use of libc to read the aux values, instead of reading
them from /proc/self/auxv.

The "use-libc-auxv" option is enabled by default, because it's more
efficient and doesn't depend on /proc, so it's likely better for most
users.

Mustang, for its part, continues to be able to use the incoming auxv on
the stack because it controls program startup. Since it doesn't have to
worry about the hazards of /proc or QEMU, it can trust the incoming
values, and do less checking.

Fixes #382.


src/utils.rs
```diff
@@ -11,3 +11,20 @@ pub(crate) const fn as_ptr<T>(t: &T) -> *const T {
 pub(crate) fn as_mut_ptr<T>(t: &mut T) -> *mut T {
     t
 }
+
+/// Convert a `*const c_void` to a `*const T`, checking that it is not null,
+/// misaligned, or pointing to a region of memory that wraps around the address
+/// space.
+#[allow(dead_code)]
+pub(crate) fn check_raw_pointer<T>(value: *const core::ffi::c_void) -> Option<*const T> {
```
ianks, line 19: For my own knowledge, I'm curious if there is any reason not to use `NonNull<T>` here so we could pack the `Option`?

cgwalters:
Shouldn't this PR call out a lot more loudly that the default has now switched to `libc`, and anyone who wants the linux-raw backend needs to opt-in via `default-features = false; features = ["linux-raw"]`?

It seems worth splitting out into a separate preparatory patch even I'd say.

cgwalters:
Alternatively, perhaps the implication here due to the indirection implied by a new `use-libc-auxv` feature is that the default might switch back later.  But still, assuming that this is planned to be part of a semver-compatible update it seems at least highlighting somewhat loudly that the backend has switched and isn't likely to switch back *soon* (right?)

src/utils.rs
sunfishcode: Only that it hadn't occurred to me :-). But looking into it, `NonNull` is `*mut` rather than `*const`, which is a little inconvenient in src/backend/linux_raw/vdso.rs where it's used a bunch. I'll need to think about this more.

sunfishcode:
> Shouldn't this PR call out a lot more loudly that the default has now switched to `libc`, and anyone who wants the linux-raw backend needs to opt-in via `default-features = false; features = ["linux-raw"]`?

The default is still linux_raw; it's just using libc for the aux values, which isn't user-visible. All `std`-using Rust programs are already linking in libc, on non-Mustang Linux platforms, so adding a dependency on libc by default shouldn't be a breaking change.

It's possible this is a breaking change for `no_std` users. Rustix has very few `no_std` users other than Mustang (which isn't affected by this change) and the rustix port of `std` (which I myself am the only user of and can deal with it). And, rustix's `no_std` support depends on nightly Rust and gets randomly broken by nightly Rust changes occasionally already. So I'm thinking this doesn't motivate a semver bump.

src/utils.rs
sunfishcode: I've now added a patch to do this. It is a little more verbose, but I think it's ok.

PR #396, opened 2022-09-05
Fix compilation errors on solaris.

gco:
Successfully compiles with --features=all-apis.   All but two tests pass.

Also depends on rust-lang/libc#2900 for other tests to pass.


sunfishcode:
Thanks!

Issue #387, opened 2022-08-13
`setns` or `prctl` seem to be missing from `rustix`.


vi:
Are Linux-specific process management syscalls in scope for `rustix`?

sunfishcode:
Yes; they're in scope, and not yet implemented.

sunfishcode:
Indeed, thanks!

PR #398, opened 2022-09-10
Add support for `setns()`

koutheir:
Contributes to #387.


src/thread/setns.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use bitflags::bitflags;
+
+use crate::backend::c::c_int;
+use crate::backend::thread::syscalls;
+use crate::fd::BorrowedFd;
+use crate::io;
+
+/// New time namespace.
+const CLONE_NEWTIME: u32 = 0x00_00_00_80;
```
sunfishcode, line 13: Can these constants use `linux_raw_sys::general::CLONE_NEWTIME` and similar instead of hard-coding the constants here?

src/thread/setns.rs
```diff
@@ -0,0 +1,122 @@
...
+            CLONE_NEWCGROUP => Ok(Self::ControlGroup),
+            CLONE_NEWUTS => Ok(Self::HostNameAndNISDomainName),
+            CLONE_NEWIPC => Ok(Self::InterProcessCommunication),
+            CLONE_NEWUSER => Ok(Self::User),
+            CLONE_NEWPID => Ok(Self::ProcessID),
+            CLONE_NEWNET => Ok(Self::Network),
+            _ => Err(io::Errno::RANGE),
+        }
+    }
+}
```
sunfishcode, line 89: rustix doesn't currently provide these kinds of `TryFrom` impls. Typically the way this works is that regular Rust users should always be using the enum type, so it don't need these, and mustang implements this conversion manually. Is there a need to do something different here?

src/backend/linux_raw/thread/syscalls.rs
```diff
@@ -278,3 +282,9 @@ unsafe fn futex_old(
         c_uint(val3)
     ))
 }
+
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[inline]
+pub(crate) fn setns(fd: BorrowedFd, nstype: c::c_int) -> io::Result<c::c_int> {
+    unsafe { ret_c_int(syscall!(__NR_setns, fd, c_int(nstype))) }
```
sunfishcode, line 289: Can this be `syscall_readonly!`, since setns doesn't mutate any memory?

koutheir: > setns doesn't mutate any memory?

`setns` does mutate memory, at least on the kernel level, and maybe even at the level of the process address space, even if that memory is not directly reachable through pointers, but rather through function calls.

sunfishcode: Yes; `readonly` in this context just refers to the program's own pointers. For example, rustix uses `readonly` with the `write` system call.

koutheir: I didn't find any documentation regarding the `syscall_readonly!()` macro. Can you explain?
1. its purpose,
2. the difference in semantics compared to `syscall!()`, and
3. the consequences of using/misusing it.

sunfishcode: It's the same as `syscall!` but with the [readonly option](https://doc.rust-lang.org/reference/inline-assembly.html#options) enabled. As documented there, the meaning of readonly in asm is: "does not write to any memory. This allows the compiler to cache the values of unmodified global variables in registers across the asm! block since it knows that they are not written to by the asm!.". Rust isn't aware of memory outside its own address space, so mutating memory in the kernel doesn't impact the readonly option. The consequence for misusing it is Undefined Behavior.

I'll submit a separate PR to add this information in a comment.

koutheir:
Should I resolve the conflicts with the `main` branch?

sunfishcode:
Yes; please do. Those are conflicting with the `prctl` PR which just landed :smile:

koutheir:
I don't know why the check "CI / Test (macos-10.15) (pull_request)" was cancelled.

sunfishcode:
It appears github is deprecating the macos-10.15 environment:

> This is a scheduled macOS-10.15 brownout. The macOS-10.15 environment is deprecated and will be removed on December 1st, 2022. For more details, see https://github.com/actions/virtual-environments/issues/5583

I'll have to remove that from github actions, but that'll be a separate PR.

koutheir:
Are there still issues with this PR?

sunfishcode:
Looks good!

PR #401, opened 2022-09-20
Add ipv6 multicast hops support

ETKNeil:
This PR focus on adding ipv6 multicast hops support:
See: https://man7.org/linux/man-pages/man7/ipv6.7.html
See: https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options

This is the same as multicast_ttl but for ipv6, it's also missing in Rust std

### DIFF
--- a/src/backend/libc/net/syscalls.rs
+++ b/src/backend/libc/net/syscalls.rs
@@ -700,6 +700,24 @@ pub(crate) mod sockopt {
     }
 
     #[inline]
+    pub(crate) fn set_ipv6_multicast_hops(
+        fd: BorrowedFd<'_>,
+        multicast_hops: u32,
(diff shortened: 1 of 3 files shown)


ETKNeil:
the current checks failing are because https://docs.rs/libc/0.2.133/libc/fn.makedev.html removed the unsafe (see https://docs.rs/libc/0.2.132/libc/fn.makedev.html )

sunfishcode:
The makedev fix has now landed on main; can you rebase this?

ETKNeil:
> The makedev fix has now landed on main; can you rebase this?

I rebased that branch onto your main

sunfishcode:
Thanks!

Issue #415, opened 2022-09-27
Failed to build on armv7 musl target on Alpine Linux


messense:
https://gitlab.alpinelinux.org/messense/aports/-/jobs/853878

rustc version information:

```bash
$ rustc -vV
rustc 1.64.0
binary: rustc
commit-hash: unknown
commit-date: unknown
host: armv7-alpine-linux-musleabihf
release: 1.64.0
LLVM version: 14.0.6
```

error message:

```
error: cannot use register `r7`: the frame pointer (r7) cannot be used as an operand for inline asm
  --> /home/buildozer/.cargo/registry/src/github.com-1285ae84e5963aae/rustix-0.35.10/src/backend/linux_raw/arch/inline/arm.rs:13:9
   |
13 |         in("r7") nr.to_asm(),
   |         ^^^^^^^^^^^^^^^^^^^^
error: cannot use register `r7`: the frame pointer (r7) cannot be used as an operand for inline asm
  --> /home/buildozer/.cargo/registry/src/github.com-1285ae84e5963aae/rustix-0.35.10/src/backend/linux_raw/arch/inline/arm.rs:25:9
   |
25 |         in("r7") nr.to_asm(),
   |         ^^^^^^^^^^^^^^^^^^^^
error: cannot use register `r7`: the frame pointer (r7) cannot be used as an operand for inline asm
  --> /home/buildozer/.cargo/registry/src/github.com-1285ae84e5963aae/rustix-0.35.10/src/backend/linux_raw/arch/inline/arm.rs:40:9
   |
40 |         in("r7") nr.to_asm(),
   |         ^^^^^^^^^^^^^^^^^^^^
error: cannot use register `r7`: the frame pointer (r7) cannot be used as an operand for inline asm
  --> /home/buildozer/.cargo/registry/src/github.com-1285ae84e5963aae/rustix-0.35.10/src/backend/linux_raw/arch/inline/arm.rs:51:9
   |
51 |         in("r7") nr.to_asm(),
   |         ^^^^^^^^^^^^^^^^^^^^
error: cannot use register `r7`: the frame pointer (r7) cannot be used as an operand for inline asm
  --> /home/buildozer/.cargo/registry/src/github.com-1285ae84e5963aae/rustix-0.35.10/src/backend/linux_raw/arch/inline/arm.rs:66:9
   |
66 |         in("r7") nr.to_asm(),
   |         ^^^^^^^^^^^^^^^^^^^^
[...]

PR #416, opened 2022-09-27
Fix linux-raw inline asm for thumb-mode.

sunfishcode:
In thumb-mode, r7 is reserved as the frame pointer and isn't permitted as an inline asm operand. It's still the syscall-number operand in syscalls though, so pass the syscall-number operand in in a separate register, copy it into r7 inside the inline asm, and then restore the value of r7 when we're done.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -159,6 +159,7 @@ jobs:
         rustup target add
         x86_64-unknown-redox
         wasm32-wasi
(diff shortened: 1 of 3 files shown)


sunfishcode:
In theory this fixes #415, but I haven't tested it yet.

sunfishcode:
I've now tested this with some custom target configuration.

PR #419, opened 2022-09-28
Pass `--target=$TARGET` when running `rustc` in `build.rs`

bsilver8192:
I'm using this in an environment where only the `$TARGET` rust standard library is available, which means `use std::arch::asm;` doesn't work for the host platform when cross compiling, even when it works for the target platform. Also in general, this is trying to detect things about the target, not the host.

For context, I'm using [Bazel](https://bazel.build/) with [rules_rust's cargo_build_script](http://bazelbuild.github.io/rules_rust/flatten.html#cargo_build_script), via [cargo-raze](https://github.com/google/cargo-raze). Native x86_64-unknown-linux-gnu compilation works fine without this patch, but cross compiling to aarch64-unknown-linux-gnu requires this patch.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -174,10 +174,13 @@ fn can_compile(code: &str) -> bool {
     use std::process::Stdio;
     let out_dir = var("OUT_DIR").unwrap();
     let rustc = var("RUSTC").unwrap();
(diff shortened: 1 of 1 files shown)


sunfishcode:
Thanks!

PR #420, opened 2022-09-30
Add logic to build.rs to autodetect whether we need to use thumb-mode.

sunfishcode:
`#[cfg(target_feature = "thumb-mode")]` isn't available on stable, so use a build.rs script to detect whether we can use `r7` in inline asm on ARM targets.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -101,6 +101,11 @@ fn main() {
         }
     }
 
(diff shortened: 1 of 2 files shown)


nibon7:
LGTM. Thanks.

messense (approved):

PR #421, opened 2022-10-02
add splice syscall

nivkner:
depends on https://github.com/sunfishcode/linux-raw-sys/pull/35 to merge, points to fork at the moment for testing

### DIFF
--- a/src/backend/libc/io/types.rs
+++ b/src/backend/libc/io/types.rs
@@ -1,6 +1,8 @@
 use super::super::c;
 #[cfg(not(target_os = "wasi"))]
 use bitflags::bitflags;
+#[cfg(any(target_os = "android", target_os = "linux"))]
+use core::marker::PhantomData;
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
 bitflags! {
@@ -119,23 +121,32 @@ pub(crate) const STDERR_FILENO: c::c_int = c::STDERR_FILENO;
 /// and therefore can be shared or mutated as needed.
 #[cfg(any(target_os = "android", target_os = "linux"))]
 #[repr(transparent)]
-pub struct IoSliceRaw(c::iovec);
+pub struct IoSliceRaw<'a> {
+    _buf: c::iovec,
+    _lifetime: PhantomData<&'a ()>,
+}
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
-impl IoSliceRaw {
+impl<'a> IoSliceRaw<'a> {
     /// Creates a new IoSlice wrapping a byte slice.
-    pub fn from_slice(buf: &[u8]) -> Self {
-        IoSliceRaw(c::iovec {
-            iov_base: buf.as_ptr() as *mut u8 as *mut c::c_void,
-            iov_len: buf.len() as _,
-        })
+    pub fn from_slice(buf: &'a [u8]) -> Self {
+        IoSliceRaw {
+            _buf: c::iovec {
+                iov_base: buf.as_ptr() as *mut u8 as *mut c::c_void,
+                iov_len: buf.len() as _,
+            },
+            _lifetime: PhantomData,
+        }
     }
 
     /// Creates a new IoSlice wrapping a mutable byte slice.
-    pub fn from_slice_mut(buf: &mut [u8]) -> Self {
-        IoSliceRaw(c::iovec {
-            iov_base: buf.as_mut_ptr() as *mut c::c_void,
-            iov_len: buf.len() as _,
-        })
+    pub fn from_slice_mut(buf: &'a mut [u8]) -> Self {
+        IoSliceRaw {
+            _buf: c::iovec {
+                iov_base: buf.as_mut_ptr() as *mut c::c_void,
+                iov_len: buf.len() as _,
+            },
+            _lifetime: PhantomData,
+        }
     }
 }
--- a/src/backend/libc/io/types.rs
+++ b/src/backend/libc/io/types.rs
@@ -37,8 +37,6 @@ bitflags! {
         const NONBLOCK = c::SPLICE_F_NONBLOCK;
         /// `SPLICE_F_MORE`
         const MORE = c::SPLICE_F_MORE;
-        /// `SPLICE_F_GIFT`
-        const GIFT = c::SPLICE_F_GIFT;
     }
 }
 
--- a/src/backend/libc/io/types.rs
+++ b/src/backend/libc/io/types.rs
@@ -27,6 +27,21 @@ bitflags! {
     }
 }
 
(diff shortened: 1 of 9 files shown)


nivkner:
updated changes to use `linux-raw-sys` crate from crates.io now that the aforementioned PR has been merged

nivkner:
problem: by adding the flag `SPLICE_F_GIFT` to `vmsplice`, the user must uphold additional requirements that make it effectively `unsafe` (cant modify memory, buffers must be page aligned).

nivkner:
another problem for `vmsplice` (and the reason it was removed from PR):
according to the [manual](https://man7.org/linux/man-pages/man2/vmsplice.2.html), the behavior of the syscall changes based on the end of the pipe used (the read side or the write side).

so you can't pass `IoSlice` to vmsplice since the function can write to the buffers, but forcing the use of `IoSliceMut` even when the buffers are used for writing to the pipe seems overly restrictive.

ideally one could have separate `vmsplice` functions for reading and writing with the appropriate slice types, but the file descriptors are untyped (maybe they should?).

sunfishcode:
> problem: by adding the flag `SPLICE_F_GIFT` to `vmsplice`, the user must uphold additional requirements that make it effectively `unsafe` (cant modify memory, buffers must be page aligned).
> 
> for now removing it from `SpliceFlags`, it is unused by `splice`, and would allow introducing a separate (unsafe) gifting `vmsplice` without a breaking change

This sounds reasonable to me. We could add a separate `vmsplice_unsafe` function if `SPLICE_F_GIFT` functionality is desired.

> another problem for `vmsplice` (and the reason it was removed from PR): according to the [manual](https://man7.org/linux/man-pages/man2/vmsplice.2.html), the behavior of the syscall changes based on the end of the pipe used (the read side or the write side).
> 
> so you can't pass `IoSlice` to vmsplice since the function can write to the buffers, but forcing the use of `IoSliceMut` even when the buffers are used for writing to the pipe seems overly restrictive.

It does seem restrictive, but in a sense, that feels like an accurate reflection of the behavior of the system call, at this low level of abstraction.

> ideally one could have separate `vmsplice` functions for reading and writing with the appropriate slice types, but the pipe file descriptors are untyped (maybe they should be typed?).

Rustix does have `Dir` and `Epoll` which operate at a somewhat higher level of abstraction. It's debatable whether those are good APIs, especially `Epoll`, but it is something we can look at if there's a need for it.

> alternatively the function could use `fcntl_getfl` to check the correct pipe was used for a given function (check if its `O_RDONLY`), returning an error otherwise, but maybe that's too high level for `rustix`.

I agree, this is too high-level for rustix. `vmsplice` is all about low-level optimization, and adding an extra `fcntl` syscall to each call would conflict with that.

nivkner:
a problem with mutable slices at the lower level is that I'm not sure you can layer immutability on top of that: `IoSliceMut` isn't just mutable, its also unique since it shares the semantics of `&mut [u8]`.

is it valid rust to have code like this?
```rust
let input: &T = &data; // could be shared, no UnsafeCell
unsafe {
    // safe because func doesn't actually mutate the input
    let input = &mut *(data as *const T as *mut T);
    func(input)
}
```

sunfishcode:
Making `vmsplice` take raw pointers sounds reasnoable to me. Some system calls don't lend themselves to safe APIs at the abstraction level of individual function calls, and it's sounding like `vmsplice` might be such a function. Possibly there could be a higher-level pipe API which tracks the reader and writers ends of the pipe (and doesn't implement `From<OwnedFd>` or similar) which could know statically whether `vmsplice` will read or write; that could be built on top of a lower-level `vmsplice` that takes a raw pointer.

nivkner:
added a `IoSliceRaw` type, since no `iovec` wrapper in `rustix` had the semantics of a raw pointer.

marked`vmsplice` as `unsafe` since even with the type ensuring only valid buffers are accepted, the alias XOR mutability invariant still depends on the caller using the right fd.

since vmsplice is already unsafe, i reintroduced `SPLICE_F_GIFT` to better match the underlying syscall.

sunfishcode:
I'll hopefully have time for a more thorough review soon, but at a first readthrough, this looks good. The `IoSliceRaw` approach makes sense.

src/io/pipe.rs
```diff
@@ -53,3 +60,64 @@ pub fn pipe() -> io::Result<(OwnedFd, OwnedFd)> {
...
+///  - [Linux]
+///
+/// [Linux]: https://man7.org/linux/man-pages/man2/splice.2.html
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[inline]
+pub fn splice<FdIn: AsFd, FdOut: AsFd>(
+    fd_in: FdIn,
+    off_in: Option<u64>,
+    fd_out: FdOut,
+    off_out: Option<u64>,
```
sunfishcode, line 86: In the linux syscall, `off_in` and `off_out` are in-out parameters, that return how many bytes were read and written, respectively.

src/backend/libc/io/types.rs
```diff
@@ -97,3 +112,30 @@ pub(crate) const STDIN_FILENO: c::c_int = c::STDIN_FILENO;
...
 #[cfg(not(windows))]
 pub(crate) const STDERR_FILENO: c::c_int = c::STDERR_FILENO;
+
+/// A buffer type used with `vmsplice`.
+/// It is guaranteed to be ABI compatible with the iovec type on Unix platforms and WSABUF on Windows.
+/// Unlike `IoSlice` and `IoSliceMut` it is semantically like a raw pointer,
+/// and therefore can be shared or mutated as needed.
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[repr(transparent)]
+pub struct IoSliceRaw(c::iovec);
```
sunfishcode, line 122: `IoSlice` has a lifetime parameter, to prevent it from holding a dangling pointer. Would it make sense to add that to `IoSliceRaw` too, to prevent it from outliving the `buf` it's constructed from?

src/io/pipe.rs
nivkner: where is that written? according to the [manual](https://man7.org/linux/man-pages/man2/splice.2.html) the number of bytes read/written is the return value on success, with no mention of `off_*` being used for output.

sunfishcode: Empirically,`*off_in` is indeed written to by the system call. I don't actually know if the change in the value can ever differ from the returned size though, so perhaps it's redundant? I guess if it is indeed redundant, then we don't need to expose it here, though in that case, it's worth documenting that that aspect of the Linux syscall isn't reflected in the rustix API.

nivkner: i would argue that if this property itself is not documented, then its not worth documenting the changes to it.
it would just be confusing.

sunfishcode (approved):
Thanks!

PR #425, opened 2022-10-10
Fix dir iterator in case of an error.

trtt:
Do not resize internal buffer on error, otherwise iterator gets stuck in infinite loop with zeroed `dirent`s after the first error. Correct behavior is to return error indefinitely if it persists.
Error can be ENOENT if the directory is unlinked while iterating, for example.


sunfishcode (approved):
Looks good; thanks for the fix!

PR #429, opened 2022-10-15
Add part of the FreeBSD `procctl` API

niluxv:
Modelled mostly after `prctl.rs`. Includes only parent death signal (`PROC_PDEATHSIG_*`) and tracing (`PROC_TRACE_*`). The former has precisely the same signature as the linux `prctl` versions.

I named the trace control function `set_dumpable_behavior` as it is called on linux, since FreeBSD's traceable flag and Linux's dumpable flag do the same thing: both control the ability to trace as well as the ability to core dump.

### DIFF
--- a/src/backend/libc/process/types.rs
+++ b/src/backend/libc/process/types.rs
@@ -381,6 +381,8 @@ pub type RawUid = c::uid_t;
 /// A CPU identifier as a raw integer.
 #[cfg(any(target_os = "android", target_os = "linux"))]
 pub type RawCpuid = u32;
+#[cfg(target_os = "freebsd")]
+pub type RawId = c::id_t;
 
 #[cfg(not(target_os = "wasi"))]
 pub(crate) type RawUname = c::utsname;
--- a/src/process/procctl.rs
+++ b/src/process/procctl.rs
@@ -1,0 +1,180 @@
+//! Bindings for the FreeBSD `procctl` system call.
+//!
+//! There are similarities (but also differences) with Linux's `prctl` system call, whose interface
+//! is located in the `prctl.rs` file.
+
+#![allow(unsafe_code)]
+
+use core::mem::MaybeUninit;
+
+use crate::backend::c::{c_int, c_uint, c_void};
+use crate::backend::process::syscalls;
+use crate::backend::process::types::{RawId, Signal};
+use crate::io;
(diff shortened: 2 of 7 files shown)


src/backend/libc/process/types.rs
```diff
@@ -381,6 +381,8 @@ pub type RawUid = c::uid_t;
 /// A CPU identifier as a raw integer.
 #[cfg(any(target_os = "android", target_os = "linux"))]
 pub type RawCpuid = u32;
+#[cfg(not(target_os = "wasi"))]
+pub type RawId = c::id_t;
```
niluxv, line 385: This is annoying, but `procctl` requires an `id_t` argument and `id_t` is not `pid_t` on FreeBSD.

(I don't know why it uses `id_t`, since the value must be either a pid or a pgid, and `pid_t` and `pgid_t` are both `i32`, while `id_t` is `i64`.)

src/process/procctl.rs
```diff
@@ -0,0 +1,173 @@
+//! Bindings for the FreeBSD `procctl` system call.
```
sunfishcode, line 1: Could you add comments here and at the top of prctl.rs mentioning that these two files have some signatures and documentation in common? Hopefully that will help us keep them in sync if they evolve.

sunfishcode (commented):
Looks good overall.

src/process/procctl.rs
```diff
@@ -0,0 +1,173 @@
...
+/// the process upon delivery of a signal whose default behavior is to produce a core dump.
+///
+/// # References
+/// - [`procctl(PROC_TRACE_CTL,...)`]
+///
+/// [`procctl(PROC_TRACE_CTL,...)`]: https://www.freebsd.org/cgi/man.cgi?query=procctl&sektion=2
+#[inline]
+pub fn set_dumpable_behavior(process: ProcSelector, config: DumpableBehavior) -> io::Result<()> {
+    unsafe { procctl(PROC_TRACE_CTL, process, config as usize as *mut _) }
+}
```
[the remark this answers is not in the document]
sunfishcode: Oh, right. In that case, can you add something to the doc comment calling out this difference? I think it's ok to use the same name for now.

src/process/procctl.rs
niluxv: Done.

Issue #76, opened 2021-10-05
What might porting `std` to use `rustix` look like?


sunfishcode:
Porting Rust's std to use `rustix` instead of using libc directly for I/O on Unix-family platforms would factor out `unsafe` code, both for memory safety and [I/O safety](https://github.com/rust-lang/rfcs/blob/master/text/3128-io-safety.md), and raw error handling, from std's main implementation code, making it easier to maintain. Secondarily, it'd also bring a number of optimizations, such as avoiding the allocation that std does on every system call that takes a string, and making system calls directly and avoiding `errno`.

[cap-std](https://github.com/bytecodealliance/cap-std) and [mustang](https://github.com/sunfishcode/mustang) demonstrate that `rustix` can do most of the I/O that std does, and the pieces that are missing should be straightforward to add. `rustix` and libc can coexist, so the work could be done incrementally.

`rustix` does depend on some things from std, though there's a [working `no_std` branch](https://github.com/bytecodealliance/rsix/tree/no_std_demo), so we know the dependencies. The most interesting here are:

   - `std::os::raw::*`, `CStr`, `Ip*Addr`, `SocketAddr*`, `IoSlice*` - Rust has avoided moving these into `core` because they invove OS-specific types, so perhaps we could introduce a new `core_ffi` library and move these into it? Or, perhaps rustix should define its own versions of these which convert to and from the `std` versions. rustix doesn't need them to be full-featured, so it should be doable.
 
  - `BorrowedFd`, `OwnedFd`, etc. - Could these also make sense in a `core_ffi` library? Alternatively, rustix could similarly define its own.

   - `CString` - Similarly, this could go into a new `alloc_ffi` library, or rustix could similarly define its own.

Once we have the dependencies sorted out, I imagine the next steps would be to make a temporary fork of std, and port some calls to use rustix, and then make a proposal to the std maintainers and use the fork to show how it simplifies the code and factors out `unsafe`, as the primary motivators.

For any of this to happen, I'll need help! Both because it's a lot of work, and because if there isn't community interest in helping with this, it's a sign that this doesn't actually make sense to do.

 - Does the above sound reasonable? Is it missing anything major? Is there anything about `rustix`'s design that should be different? What questions should I be asking here that I'm not?

 - Are there people interested and available to implement pieces of this? Many of the pieces should be straightforward, and I'd be happy to mentor people and explain what needs to be done.

sunfishcode:
[...]

PR #430, opened 2022-10-15
Fix compilation of libc backend without std

niluxv:
Compilation failed without std for the libc backend (not the linux_raw one). This fixes the issues. Hopefully I didn't break anything else.

* Allow missing docs in `backend/libc/io/io_slice.rs` (`std` filler code)
* Use crate local `OwnedFd` instead of the one in `io_lifetimes` in `backend/libc/io/epoll.rs`
* Add missing BSD specific argument to `SocketAddrV6` constructor in `src/addr/net.rs` (`std` filler code)

I have to say I wonder why system specific `SocketAddrV4/6` representations are used in the `no_std` case while with `std` the platform independent `std::net::SocketAddrV4/6` are used. The platform specific representation feels like a portability hazard (as the final issue shows).

### DIFF
--- a/src/net/addr.rs
+++ b/src/net/addr.rs
@@ -447,15 +447,40 @@ impl SocketAddrV6 {
     #[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))]
     #[must_use]
     pub fn new(ip: Ipv6Addr, port: u16, flowinfo: u32, scope_id: u32) -> SocketAddrV6 {
-        SocketAddrV6 {
-            inner: sockaddr_in6_new(
-                c::AF_INET6 as c::sa_family_t,
-                port.to_be(),
-                flowinfo,
-                ip.inner,
-                scope_id,
-            ),
-        }
+        #[cfg(any(
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "haiku",
+            target_os = "ios",
+            target_os = "macos",
+            target_os = "netbsd",
+            target_os = "openbsd",
+        ))]
+        let inner = sockaddr_in6_new(
+            core::mem::size_of::<c::sockaddr_in6>() as u8,
+            c::AF_INET6 as c::sa_family_t,
+            port.to_be(),
+            flowinfo,
+            ip.inner,
+            scope_id,
+        );
+        #[cfg(not(any(
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "haiku",
+            target_os = "ios",
+            target_os = "macos",
+            target_os = "netbsd",
+            target_os = "openbsd",
+        )))]
+        let inner = sockaddr_in6_new(
+            c::AF_INET6 as c::sa_family_t,
+            port.to_be(),
+            flowinfo,
+            ip.inner,
+            scope_id,
+        );
+        SocketAddrV6 { inner }
     }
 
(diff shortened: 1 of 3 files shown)


src/net/addr.rs
```diff
@@ -447,15 +447,40 @@ impl SocketAddrV6 {
...
+            target_os = "netbsd",
+            target_os = "openbsd",
+        )))]
+        let inner = sockaddr_in6_new(
+            c::AF_INET6 as c::sa_family_t,
+            port.to_be(),
+            flowinfo,
+            ip.inner,
+            scope_id,
+        );
```
niluxv, line 482: It is ugly, but unfortunately attributes on expressions require an unstable feature, so you can't write
```rust
let inner = sockaddr_in6_new(
    #[cfg(any(target_os = "freebsd", ...))]
    core::mem::size_of::<c::sockaddr_in6>() as u8,
    c::AF_INET6 as c::sa_family_t,
    ...
);
```

niluxv:
Maybe it would also be a good idea to add CI so this won't break again in the future.

sunfishcode:
To be honest, I didn't anticipate anyone would try this with the libc backend 😂. no_std is all about not depending on an OS, and rustix is all about it interfacing with an OS. But I'll accept reasonable patches to improve it.

The `SocketAddr` situation changed in Rust 1.64 recently, and rustix just hasn't been updated to follow that yet.

I'll review this soon.

niluxv:
> To be honest, I didn't anticipate anyone would try this with the libc backend joy. no_std is all about not depending on an OS, and rustix is all about it interfacing with an OS.

Ok, fair. But given how few changes were required to make it compile, I'd say rustix has been designed to handle `no_std` also for the libc backend. It can't be by accident that it *almost* works... (There are quite some `#[cfg(not(feature = "std"))]` attributes in the libc backend code.) And FWIW being `no_std` is required for [implementing std in terms of rustix](https://github.com/bytecodealliance/rustix/issues/76).

> The `SocketAddr` situation changed in Rust 1.64 recently, and rustix just hasn't been updated to follow that yet.

Ah, I see. Would you be interested in a PR following std in this?

sunfishcode:
> > To be honest, I didn't anticipate anyone would try this with the libc backend joy. no_std is all about not depending on an OS, and rustix is all about it interfacing with an OS.
> 
> Ok, fair. But given how few changes were required to make it compile, I'd say rustix has been designed to handle `no_std` also for the libc backend. It can't be by accident that it _almost_ works... (There are quite some `#[cfg(not(feature = "std"))]` attributes in the libc backend code.) And FWIW being `no_std` is required for [implementing std in terms of rustix](https://github.com/bytecodealliance/rustix/issues/76).

Yes, I am familiar with that issue :grinning:. It was an experiment, and one that I or someone else may resume at some point, so I've not thrown any of it away; just didn't anticipate anyone else using it for anything else.

> > The `SocketAddr` situation changed in Rust 1.64 recently, and rustix just hasn't been updated to follow that yet.
> 
> Ah, I see. Would you be interested in a PR following std in this?

If you'd like to, that'd be great. That said, there is also talk of Rust [moving `SocketAddrV4` et al into core](https://github.com/rust-lang/rfcs/pull/2832), now that those types no longer depend on libc, and if that happens, rustix would be able to remove its copy of these types entirely, so another option is to just wait for now.

PR #433, opened 2022-10-24
Use platform independent `IpAddr` and `SocketAddr`

niluxv:
Update `src/net/ip.rs` and `src/net/addr.rs` to use the new std code, making these structures platform independent.

Doesn't include the new Debug/Display formatting code for these structures, although that could be added now that it doesn't depend on std anymore.


niluxv:
CI failures are caused by [this change in `libc`](https://github.com/rust-lang/libc/pull/2963).

niluxv:
The libc issue has been reverted, so CI should work again. Any way to retrigger CI without making changes to the PR?

niluxv:
Rebased, and the CI seems to pass now. (I don't know why macos test was cancelled.)

sunfishcode:
It looks like it was a transient CI failure.

PR #442, opened 2022-11-11
Fix compilation on macos without std

niluxv:


niluxv:
Looks like the CI failure is caused by the new io-lifetimes release (1.0.1).

sunfishcode:
Yeah, that's due to the io-lifetimes update. I've now submitted https://github.com/bytecodealliance/rustix/pull/443 to fix it.

PR #444, opened 2022-11-15
Change the procfs code to be tolerate of files being owned by root:root.

### LINKED ISSUE #441: `proc_self_status` errors when `/proc/self/status` is root owned
This happens when the process is not dumpable (which can have a variety of reasons, see [the proc manpage](https://man7.org/linux/man-pages/man5/proc.5.html) under "Files and directories" and [the prctl manpage](https://man7.org/linux/man-pages/man2/prctl.2.html) under "PR_SET_DUMPABLE").

The error that is returned:
```
Os { code: 95, kind: Uncategorized, message: "Operation not supported" }
```

(Note that `/proc/self/status` ~~is not root owned~~; the user has sufficient permissions to open that file. `std::fs::File::open` works without problems.)

sunfishcode:
When a process is marked as non-dumpable, Linux changes the permissions of the files under /proc/<pid> as owned by root:root. Relax the procfs code, including removing use of `O_NOATIME`, so that it works on files owned by `root:root`.

Fixes #441.

### DIFF
--- a/src/io/procfs.rs
+++ b/src/io/procfs.rs
@@ -75,7 +75,16 @@ fn check_proc_entry_with_stat(
     // Check the ownership of the directory. We can't do that for the toplevel
     // "/proc" though, because in e.g. a user namespace scenario, root outside
     // the container may be mapped to another uid like `nobody`.
-    if !matches!(kind, Kind::Proc) && (entry_stat.st_uid, entry_stat.st_gid) != (uid, gid) {
+    //
+    // In addition to allowing the expected uid:gid, also allow root:root, as
+    // some files in procfs can change [to root:root] when a process is marked
+    // as non-dumpable.
+    //
+    // [to root:root]: https://man7.org/linux/man-pages/man5/proc.5.html
+    if !matches!(kind, Kind::Proc)
+        && (entry_stat.st_uid, entry_stat.st_gid) != (uid, gid)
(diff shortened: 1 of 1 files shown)


src/io/procfs.rs
```diff
@@ -75,7 +75,16 @@ fn check_proc_entry_with_stat(
     // Check the ownership of the directory. We can't do that for the toplevel
     // "/proc" though, because in e.g. a user namespace scenario, root outside
     // the container may be mapped to another uid like `nobody`.
-    if !matches!(kind, Kind::Proc) && (entry_stat.st_uid, entry_stat.st_gid) != (uid, gid) {
+    //
+    // In addition to allowing the expected uid:gid, also allow root:root, as
```
cgwalters, line 79: I'm still a bit skeptical of the value of these checks versus just verifying the filesystem type is `procfs`, FWIW.

sunfishcode: It's admittedly murky. Linux doesn't really make any guarantees about what procfs *won't* do in future versions. The idea here was to check everything we practically can to hopefully catch anything unusual.

Thinking about this more, I wouldn't be surprised if some combination of user namespaces and `PR_SET_DUMPABLE` could cause even the new check here to spuriously fail. So maybe we should remove the uid check altogether.

sunfishcode: I've now submitted https://github.com/bytecodealliance/rustix/pull/450 removing the uid/gid checks entirely.

PR #448, opened 2022-11-15
Use stdlib with_c_str implementation

SUPERCILEX:
This removes unnecessary zeroing.


sunfishcode:
Cool! Could you add a `Safety` comment on the unsafe block? It doesn't need to say a lot; perhaps it can just mention that it's doing what the standard library code it's derived from does.

SUPERCILEX:
Sounds good, done!

sunfishcode:
Thanks!

Issue #445, opened 2022-11-15
Update feature gates?


SUPERCILEX:
I've been poking around the code and saw this: https://github.com/bytecodealliance/rustix/blob/74312285f64f0ae2a547c1911221fbabfa5c75c0/src/backend/linux_raw/mod.rs#L34

In general, is the stdlib version of I/O safety used or is a copy included in rustix? It would be nice if only the stdlib version was used (assuming the std feature is enabled).

sunfishcode:
It does use the std version when the std feature is enabled, and the Rust version is >= 1.63 (when I/O safety was stabilized).

The copy of `OwnedFd` etc. in the tree is only used in the no_std configuration.

SUPERCILEX:
That's good to know, thanks. Can the feature gate be re-enabled then?

sunfishcode:
Possibly. It'll probably take some other fiddling, but I'll take a quick look.

sunfishcode:
I've now posted https://github.com/bytecodealliance/rustix/pull/449.

SUPERCILEX:
Awesome, thanks!

sunfishcode:
Fixed in #449.

PR #449, opened 2022-11-15
Disable the internal `fs` module when the "fs" feature isn't enabled.

sunfishcode:
Rustix no longer has anything in the tree that needs the `fs` module when the public "fs" feature isn't enabled, so make the `fd` module conditional on `cfg(feature = "fs")`.


sunfishcode:
Fixes #445.

PR #455, opened 2022-11-21
Fix clippy lint

SUPERCILEX:
Not sure if the debug assert is worth it since it's just a rewording of the if check above.

### DIFF
--- a/src/path/arg.rs
+++ b/src/path/arg.rs
@@ -933,7 +933,7 @@ impl Arg for DecInt {
 }
 
 /// Runs a closure with `bytes` passed in as a `&CStr`.
-#[allow(unsafe_code)]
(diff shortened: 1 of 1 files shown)


sunfishcode:
I'm inclined to keep the `debug_assert`, because the `+ 1` for the NUL terminator is a little subtle.

I'm not comfortable with the clutter this patch adds. This is in subtle unsafe code, and when I'm reading this code, I want to think about the safety conditions, and not about a clippy lint that doesn't turn out to be helpful here. Could you move the clippy directive to apply to the whole function, or perhaps the whole file?

SUPERCILEX:
> I'm inclined to keep the debug_assert, because the + 1 for the NUL terminator is a little subtle.
>
> I'm not comfortable with the clutter this patch adds.

Agreed, moved it to the function.

Issue #451, opened 2022-11-16
Openness to a low-level getdents64 directory iterator?


SUPERCILEX:
The stdlib and nix use libc's readdir to do directory iteration while rustix uses getdents, all of which are slow. readdir is C, so it includes locking to support multithreading (which I'm pretty sure isn't necessary in rust due to the ownership model). nix is faster than the stdlib because it doesn't allocate anything while iterating whereas the stdlib allocates path names. rustix suffers from the same allocation problem, but is quite a bit faster than nix or the stdlib because it uses getdents.

A zero-compromise getdents implementation is still ~15% faster than rustix:

```
Benchmark 1: ./dents /tmp/ftzz-test
  Time (mean ± σ):     166.8 ms ±   1.4 ms    [User: 5.2 ms, System: 161.5 ms]
  Range (min … max):   165.4 ms … 169.8 ms    17 runs
 
Benchmark 2: ./nix /tmp/ftzz-test
  Time (mean ± σ):     201.7 ms ±   1.6 ms    [User: 41.4 ms, System: 160.3 ms]
  Range (min … max):   199.7 ms … 205.6 ms    14 runs
 
Benchmark 3: ./stdlib /tmp/ftzz-test
  Time (mean ± σ):     209.4 ms ±   1.2 ms    [User: 45.6 ms, System: 163.7 ms]
  Range (min … max):   207.8 ms … 212.6 ms    14 runs
 
Benchmark 4: ./rustix /tmp/ftzz-test
  Time (mean ± σ):     191.3 ms ±   2.1 ms    [User: 25.0 ms, System: 166.3 ms]
  Range (min … max):   186.6 ms … 194.9 ms    15 runs
 
Summary
  './dents /tmp/ftzz-test' ran
    1.15 ± 0.02 times faster than './rustix /tmp/ftzz-test'
    1.21 ± 0.01 times faster than './nix /tmp/ftzz-test'
    1.25 ± 0.01 times faster than './stdlib /tmp/ftzz-test'
```

I have an implementation open for nix: https://github.com/nix-rust/nix/pull/1856. The API requires a buffer and makes you handle resizing if you need that. The API is currently safe, but I need to see what happens if you seek a directory and if that's worth an unsafe (rustix's existing Dir should have the same problem of potentially garbage offsets since you can dup fds). I can explain more around the reasoning of the API if there's interest.

Would you guys also be interested? If so, I'll look into opening up a PR in the next few days.

[...]

PR #457, opened 2022-11-21
Add zero-compromise directory iteration

SUPERCILEX:
Closes https://github.com/bytecodealliance/rustix/issues/451

Background reading:
- https://man7.org/linux/man-pages/man2/getdents.2.html
- https://sourceware.org/bugzilla/show_bug.cgi?id=14
- https://sourcegraph.com/github.com/torvalds/linux/-/blob/fs/readdir.c?L310-352&subtree=true

### Notes

Based on the above background reading, the following assumptions appear to be sound:
- The kernel will align returned dirents and they will be contiguous.
- Seeking into garbage offsets is safe as the kernel will not return partial dirents. In fact, I've discovered that the `d_off` is actually a cookie that you can use to seek to the next dirent. It has no relation to byte offsets.

### Benchmarks

```
Benchmark 1: ./nix-raw /tmp/ftzz-test
  Time (mean ± σ):     197.4 ms ±   5.2 ms    [User: 5.8 ms, System: 190.1 ms]
  Range (min … max):   191.2 ms … 208.3 ms    15 runs
 
Benchmark 2: ./rustix-raw /tmp/ftzz-test
  Time (mean ± σ):     190.9 ms ±   4.1 ms    [User: 2.4 ms, System: 188.0 ms]
  Range (min … max):   186.7 ms … 202.1 ms    15 runs
 
Benchmark 3: ./nix /tmp/ftzz-test
  Time (mean ± σ):     228.7 ms ±   7.2 ms    [User: 39.9 ms, System: 188.7 ms]
  Range (min … max):   223.0 ms … 250.2 ms    13 runs
 
Benchmark 4: ./rustix /tmp/ftzz-test
  Time (mean ± σ):     246.3 ms ±   6.9 ms    [User: 40.9 ms, System: 201.8 ms]
  Range (min … max):   235.5 ms … 256.5 ms    11 runs
 
Benchmark 5: ./stdlib /tmp/ftzz-test
  Time (mean ± σ):     237.6 ms ±   5.2 ms    [User: 56.0 ms, System: 180.0 ms]
  Range (min … max):   232.0 ms … 246.8 ms    12 runs
 
Summary
  './rustix-raw /tmp/ftzz-test' ran
    1.03 ± 0.04 times faster than './nix-raw /tmp/ftzz-test'
    1.20 ± 0.05 times faster than './nix /tmp/ftzz-test'
    1.24 ± 0.04 times faster than './stdlib /tmp/ftzz-test'
    1.29 ± 0.05 times faster than './rustix /tmp/ftzz-test'
```

### Optimal buf size analysis

I set up benchmarks that used power-of-two sized buffers. 2^13 seemed optimal and it's also what the stdlib uses in its BufWriter and BufReader implementations.

```
Benchmark 1: ./rustix5 /tmp/ftzz-1
  Time (mean ± σ):     345.7 ms ±   2.8 ms    [User: 35.5 ms, System: 309.9 ms]
  Range (min … max):   343.4 ms … 353.1 ms    10 runs
 
Benchmark 2: ./rustix6 /tmp/ftzz-1
  Time (mean ± σ):     258.5 ms ±   1.9 ms    [User: 24.9 ms, System: 233.4 ms]
  Range (min … max):   255.3 ms … 260.8 ms    11 runs
 
Benchmark 3: ./rustix7 /tmp/ftzz-1
  Time (mean ± σ):     217.8 ms ±   3.4 ms    [User: 17.4 ms, System: 200.2 ms]
  Range (min … max):   214.6 ms … 228.4 ms    13 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark 4: ./rustix8 /tmp/ftzz-1
  Time (mean ± σ):     194.1 ms ±   0.9 ms    [User: 15.9 ms, System: 178.1 ms]
  Range (min … max):   192.9 ms … 196.1 ms    15 runs
 
Benchmark 5: ./rustix9 /tmp/ftzz-1
  Time (mean ± σ):     181.3 ms ±   0.8 ms    [User: 13.9 ms, System: 167.2 ms]
  Range (min … max):   179.0 ms … 182.7 ms    16 runs
 
Benchmark 6: ./rustix10 /tmp/ftzz-1
  Time (mean ± σ):     175.1 ms ±   1.6 ms    [User: 12.4 ms, System: 162.6 ms]
  Range (min … max):   171.0 ms … 177.6 ms    17 runs
 
Benchmark 7: ./rustix11 /tmp/ftzz-1
  Time (mean ± σ):     170.3 ms ±   2.2 ms    [User: 12.8 ms, System: 157.3 ms]
  Range (min … max):   167.6 ms … 173.1 ms    17 runs
 
Benchmark 8: ./rustix12 /tmp/ftzz-1
  Time (mean ± σ):     170.4 ms ±   0.6 ms    [User: 9.1 ms, System: 161.3 ms]
  Range (min … max):   170.0 ms … 172.1 ms    17 runs
 
Benchmark 9: ./rustix13 /tmp/ftzz-1
  Time (mean ± σ):     169.2 ms ±   1.8 ms    [User: 8.8 ms, System: 160.2 ms]
  Range (min … max):   165.4 ms … 173.1 ms    17 runs
 
Benchmark 10: ./rustix15 /tmp/ftzz-1
  Time (mean ± σ):     167.9 ms ±   2.2 ms    [User: 11.2 ms, System: 156.7 ms]
  Range (min … max):   164.7 ms … 172.5 ms    17 runs
 
Benchmark 11: ./rustix20 /tmp/ftzz-1
  Time (mean ± σ):     168.4 ms ±   1.0 ms    [User: 9.5 ms, System: 158.8 ms]
  Range (min … max):   167.7 ms … 171.8 ms    17 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Summary
  './rustix15 /tmp/ftzz-1' ran
    1.00 ± 0.01 times faster than './rustix20 /tmp/ftzz-1'
    1.01 ± 0.02 times faster than './rustix13 /tmp/ftzz-1'
    1.01 ± 0.02 times faster than './rustix11 /tmp/ftzz-1'
    1.01 ± 0.01 times faster than './rustix12 /tmp/ftzz-1'
    1.04 ± 0.02 times faster than './rustix10 /tmp/ftzz-1'
    1.08 ± 0.01 times faster than './rustix9 /tmp/ftzz-1'
    1.16 ± 0.02 times faster than './rustix8 /tmp/ftzz-1'
    1.30 ± 0.03 times faster than './rustix7 /tmp/ftzz-1'
    1.54 ± 0.02 times faster than './rustix6 /tmp/ftzz-1'
    2.06 ± 0.03 times faster than './rustix5 /tmp/ftzz-1'
```


src/backend/linux_raw/fs/raw_dir.rs
```diff
@@ -0,0 +1,204 @@
...
+
+/// Suggested buffer size for use in [`RawDir::new`]. There is no requirement to
+/// use this buffer size, but it was found to be a good tradeoff between memory
+/// use and syscall count.
+pub const DIR_BUF_LEN: usize = 8192;
+
+/// A directory iterator implemented with getdents.
+///
+/// This implementation:
+/// - Excludes deleted inodes (with ID 0).
```
sunfishcode, line 21: Could you say more about the motivation for excluding entries with inode 0? I don't see any mention of this in the `getdents` documentation. 

From some searching, it looks like glibc historically skipped entries with d_ino == 0, but has recently been changed to not skipping them:
 - https://sourceware.org/bugzilla/show_bug.cgi?id=12165
 - https://sourceware.org/pipermail/libc-alpha/2022-September/142059.html

src/backend/linux_raw/fs/raw_dir.rs
```diff
@@ -0,0 +1,204 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use std::mem::MaybeUninit;
+    /// # use rustix::fs::{cwd, Mode, OFlags, openat, RawDir};
+    ///
+    /// let fd = openat(cwd(), ".", OFlags::RDONLY | OFlags::DIRECTORY, Mode::empty()).unwrap();
+    ///
+    /// let mut buf = [MaybeUninit::uninit(); 2048];
```
sunfishcode, line 45: Can this example use `DIR_BUF_LEN`?

src/backend/linux_raw/fs/raw_dir.rs
```diff
@@ -0,0 +1,204 @@
...
+    ///         }
+    ///         break 'read;
+    ///     }
+    ///
+    ///     let new_capacity = max(buf.capacity() * 2, 1);
+    ///     buf.reserve(new_capacity);
+    /// }
+    /// ```
+    ///
+    /// Note that this is horribly inefficient as we'll most likely end up doing ~1 syscall per file.
```
sunfishcode, line 83: This doc comment is a little confusing; it gives the impression that users have a choice between doing something horribly inefficient or doing something that doesn't support arbitrarily-large file paths.

Can we guarantee that `DIR_BUF_LEN` is long enough to support any path that the host OS supports (considering `NAME_MAX`)?

src/backend/linux_raw/fs/raw_dir.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// Returns the type of this directory entry.
+    #[inline]
+    pub fn file_type(&self) -> FileType {
+        FileType::from_dirent_d_type(self.file_type)
+    }
+
+    /// Returns the inode number of this directory entry.
+    #[inline]
+    #[doc(alias = "ino")]
+    pub fn inode_number(&self) -> u64 {
```
sunfishcode, line 121: I like verbose names in general, however Rust's standard library calls a similar function [`ino`](https://doc.rust-lang.org/stable/std/os/unix/fs/trait.DirEntryExt.html#tymethod.ino), and the Unix field is named `d_ino`, so would it be ok to name this `ino()` for familiarity?

sunfishcode (commented):
Thanks for working on this! I still need to read through the main implementation code another time, but here are a few initial review comments.

src/backend/linux_raw/fs/raw_dir.rs
SUPERCILEX: Oh wow, that discussion is super recent. Thanks for looking into this! My implementation was derived from [glibc's readdir64_r](https://github.com/bminor/glibc/blob/b92a49359f33a461db080a33940d73f47c756126/sysdeps/unix/sysv/linux/readdir64_r.c) which hasn't merged in that patch I guess. Either way, happy to remove the inode check.

src/backend/linux_raw/fs/raw_dir.rs
SUPERCILEX: Yup, sounds good.

src/backend/linux_raw/fs/raw_dir.rs
SUPERCILEX: Before making docs changes, I want to make sure we're on the same page since I think we're coming at this from different angles. I was trying to demo the edge cases while assuming that people would know to use `Vec::spare_capacity_mut` for everyday life, but that's probably a bad assumption.

Here's how I'd like people to use the API:
- When using recursion, people should either use a vec or the stack. If using the stack, I'd lean towards a smaller value to minimize wasted space and skip [stack probes](https://docs.rs/compiler_builtins/latest/compiler_builtins/probestack/index.html), hence the 2048. If using the heap, then we can afford to waste more space, hence using 8192. So maybe this suggests having a constant is actually a bad idea? Instead we could tell people to "Use a buffer size of at least NAME_MAX+24 bytes. We suggest 2048 for stack allocated buffers and 8192 for heap allocated buffers." In practice, almost all file systems use 255 or smaller as their limit: https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits. Reiser appears to be the only deviant.
- When not using recursion, people would ideally have a cached vec somewhere that they re-use.

> Can we guarantee that DIR_BUF_LEN is long enough to support any path that the host OS supports (considering NAME_MAX)?

I don't think so unfortunately, which is why I'm leaning towards removing the constant. There are basically no guarantees on NAME_MAX: https://www.gnu.org/software/libc/manual/html_node/Limits-for-Files.html. Wikipedia has the current file system limits, but nothing prevents a future file system from removing the limit entirely for example.

Going back to the docs, what about having a simple heap example, a simple stack example, and then a production-ready buffer resizing example?

sunfishcode: > Going back to the docs, what about having a simple heap example, a simple stack example, and then a production-ready buffer resizing example?

That sounds good. I imagine you could even skip the simple heap example if you wanted to. I imagine the vast majority of Rust code will continue using `std::fs::ReadDir` to read directories, or [walkdir](https://crates.io/crates/walkdir) or so, so I imagine the main audience for using this API directly will be people doing low-level optimization work.

For the stack example, I'm not comfortable encouraging people to use fixed-sized buffers if OS's don't have a limit. If we're not confident enough to assume a `NAME_MAX` exists, it feels awkward to make suggestions to users that they bake in numbers like 2048. Could we instead give some guidance like, "only use this approach for reading directories with known layouts where all entries have names less than X" or so?

src/backend/linux_raw/fs/raw_dir.rs
```diff
@@ -0,0 +1,201 @@
...
+                            name_len += 1;
+                        }
+                        // Add 1 for the NUL byte
+                        name_len += 1;
+
+                        // SAFETY: We found the NUL byte and length above.
+                        unsafe {
+                            let file_name = CStr::from_bytes_with_nul_unchecked(
+                                slice::from_raw_parts(name_start, name_len),
+                            );
```
sunfishcode, line 184: Instead of searching for the NUL above and then doing `from_bytes_with_nul_unchecked` here, could this use `CStr::from_ptr`?

src/backend/linux_raw/fs/raw_dir.rs
```diff
@@ -0,0 +1,201 @@
...
+    ///             dbg!(&entry);
+    ///         }
+    ///         break 'read;
+    ///     }
+    ///
+    ///     let new_capacity = buf.capacity() * 2;
+    ///     buf.reserve(new_capacity);
+    /// }
+    /// ```
+    pub fn new(fd: Fd, buf: &'buf mut [MaybeUninit<u8>]) -> Self {
```
sunfishcode, line 85: Do you know if `getdents` requires the buffer to be aligned at all? If so, we should document that, and either change the type here or potentially make this `unsafe`.

sunfishcode (commented):
Cool, thanks for working on this!

I notice this is implemented only for the `linux_raw` backend. I think that's fine for now, though in case you're interested, a next step here is implementing it for the libc backend, using `libc::syscall` to call `getdents`.

src/backend/linux_raw/fs/raw_dir.rs
```diff
@@ -0,0 +1,201 @@
...
+            initialized: 0,
+            offset: 0,
+        }
+    }
+}
+
+/// A raw directory entry, similar to `std::fs::DirEntry`.
+///
+/// Note that unlike the std version, this may represent the `.` or `..` entries.
+#[derive(Debug)]
```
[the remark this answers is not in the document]
SUPERCILEX: Good catch, done.

SUPERCILEX:
> I notice this is implemented only for the linux_raw backend. I think that's fine for now, though in case you're interested, a next step here is implementing it for the libc backend, using libc::syscall to call getdents.

Done.

src/backend/linux_raw/fs/raw_dir.rs
SUPERCILEX: > Could we instead give some guidance like, "only use this approach for reading directories with known layouts where all entries have names less than X" or so?

I think that's reasonable, I added clarifications for the simple stack + heap examples. Happy to remove either if you'd rather not have the example at all.

src/backend/linux_raw/fs/raw_dir.rs
SUPERCILEX: Based on some experimentation, no:

```rust
        slice::from_raw_parts_mut(
            buf.as_mut_ptr().add(3) as *mut MaybeUninit<u8>,
            buf.capacity() - 3,
        )
```

That still works.

And this popped up in a search: https://github.com/microsoft/WSL/issues/1769

> Anyway it **isn't** an ABI requirement the pointer be 8 byte aligned

So I'm going to say I'm fairly confident the buffer doesn't have to be aligned. I would also expect the syscall to fail if it needs to be aligned which means it isn't a safety issue.

src/backend/linux_raw/fs/raw_dir.rs
```diff
@@ -0,0 +1,201 @@
...
+    #[inline]
+    #[doc(alias = "inode_number")]
+    pub fn ino(&self) -> u64 {
+        self.inode_number
+    }
+
+    /// Returns the seek cookie to the next directory entry.
+    #[inline]
+    #[doc(alias = "off")]
+    pub fn next_entry_cookie(&self) -> i64 {
```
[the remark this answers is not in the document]
SUPERCILEX: Happy to do that, but I think we'd need a transmute to make sure no funny business occurs on losing the sign (unless there's a safe way to re-interpret the bits I'm forgetting). Not sure that's worth it.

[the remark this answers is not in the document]
sunfishcode: It wouldn't need a transmute, just an `as u64` before returning it here, which is always a lossless cast.

src/backend/linux_raw/fs/raw_dir.rs
SUPERCILEX: ~We could, but that would make me sad (I actually have a debug_assert that uses from_ptr). The current implementation creates the CStr in constant time whereas using from_ptr will scale with the length of the file name.~

Son of a biscuit, premature optimization strikes again. from_raw is annoyingly faster than my constant-time thing because it uses avx2 instructions. Done. :)

I tested it with file names that were ~30 bytes long which I'd actually consider to be longer than average.

src/backend/linux_raw/fs/raw_dir.rs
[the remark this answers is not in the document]
SUPERCILEX: TIL, I guess I was thinking of overflow. Done.

SUPERCILEX:
Looks like I'm going to need help figuring out the right cfgs. Are there more `target_os`s than linux that support getdents64? Not sure where to look this up

sunfishcode:
The main other `target_os` that supports getdents64 is "android". So `any(target_os = "android", target_os = "linux")` should be good for this.

SUPERCILEX:
Thanks!

sunfishcode:
Oh, hrm. The `linux_raw_sys` bindings have the wrong type for x32. That might be pretty tricky to sort out.

Maybe what we could do here is just special-case x32 here. When `#[cfg(all(target_arch == "x86_64", target_pointer_width == "32"))]` is true, use a locally-defined `linux_dirent64`, otherwise use the `linux_dirent64` defined in `linux_raw_sys`. That's not super pretty, but maybe it'd be good enough for now?

SUPERCILEX:
That's a little sad, but done.

sunfishcode:
Thinking about this more, I believe I now have a better solution for x32: https://github.com/sunfishcode/linux-raw-sys/pull/36

It's just a matter of putting special cases in the right place :sweat_smile:.

sunfishcode:
Ok, that patch is now in linux-raw-sys 0.1.3. Could you try updating that and trying this patch without the special case for x86?

SUPERCILEX:
Sweet, thanks much nicer. Thanks!

sunfishcode:
Looks good, thanks!

SUPERCILEX:
Woohoo, thanks!

PR #464, opened 2022-11-28
Fix UB due to misaligned RawDir pointer

SUPERCILEX:
I done goofed. I believe my comments about the OS not breaking stuff when poor alignment is involved are still correct, but those points are actually irrelevant. From the Rust Abstract Machine standpoint, unaligned pointer reads are UB, so it doesn't matter if the OS returns structs that are aligned relative to the start pointer since that still means every single struct will be (consistently) misaligned.

Options considered:
- Use `read_unaligned`. On x86_64 and armv8 this generates perfectly acceptable code (though performance will presumably suffer), but on older arm impls and risc-v, the generated code is vomit worthy: https://rust.godbolt.org/z/5qq14ea4v
- Require the passed in buffer to be aligned and throw in an assertion to guarantee this or make the function unsafe and don't check it. This is kinda lame IMO b/c everyone is going to get it wrong or spend hours to get it right. Not worth it.
- [CHOSEN] Trim the buffer such that it becomes aligned. This is a little sad because it means we have even fewer guarantees on good buffer sizes, but we already kind of gave up on that so eh.

### DIFF
--- a/src/fs/raw_dir.rs
+++ b/src/fs/raw_dir.rs
@@ -1,7 +1,7 @@
 //! `RawDir` and `RawDirEntry`.
 
 use core::fmt;
-use core::mem::MaybeUninit;
+use core::mem::{align_of, MaybeUninit};
 use linux_raw_sys::general::linux_dirent64;
 
 use crate::backend::fs::syscalls::getdents_uninit;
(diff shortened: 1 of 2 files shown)


sunfishcode:
Could you add a testcase that passes in a deliberately misaligned buffer? Perhaps it could allocate a buffer and call the API with `&buffer[..]` and then again with `&buffer[1..]`, so that at least one of those is misaligned?

sunfishcode:
Also, don't worry about the macos-10.15 failure; that's unrelated.

SUPERCILEX:
Sounds good, done! I see you guys have a risc-v test, so I undid my changes to hopefully make that test segfault.

SUPERCILEX:
Nice, several arches faulted. Dropping the undo commit.

sunfishcode:
Thanks!

PR #465, opened 2022-11-29
Run the tests on Rust 1.48, and check compilation of tests and examples on more targets

sunfishcode:


cgwalters (approved):
Only skimmed but seems sane.  I'm sure there is a reason that 1.48 is targeted for some reverse dependency, if it can be named it seems useful to do so so that in the future we know when it can be bumped.

notgull:
> Only skimmed but seems sane. I'm sure there is a reason that 1.48 is targeted for some reverse dependency, if it can be named it seems useful to do so so that in the future we know when it can be bumped.

The README says:

> This crate currently works on the version of [Rust on Debian stable](https://packages.debian.org/stable/rust/rustc), which is currently Rust 1.48. 

In addition, this also lets this crate be used as a dependency of [`async-io`](https://github.com/smol-rs/async-io), which has a similar MSRV policy.

cgwalters:
Ah, OK that seems useful to include in the PR description or initial commit messages.

I mentioned this in a recent PR too, I think it'd make sense to codify more explicitly by using `rust-version` in `Cargo.toml`, then testing *that* in CI like https://github.com/ostreedev/ostree-rs-ext/blob/3db50b0001eaba208b0943dd3c6c3074a03a432e/.github/workflows/rust.yml#L74

sunfishcode:
I hadn't added rust-version yet because it's only recognized in Rust >= 1.56; older versions issue a warning: "warning: unused manifest key: package.rust-version". But maybe that's something we can live with at this point. It looks like the warning is only printed when rustix is built at the top level; when it's a dependency, I don't see the warning. So I've now added it.

I didn't make the CI script extract it because Rustix's CI runs lots of different configurations in a matrix and I'm not familiar enough with how that works to make this kind of change, but if someone else wanted to figure it out, I wouldn't object.

sunfishcode:
I've also now added a mention of async-io to the commit message.

PR #466, opened 2022-11-29
Deduplicate the "context" set of structures

notgull:
The `Context`, `Ref`, `Borrowing` and `Owned` structures are written twice; once for the `libc` `epoll` backend, and once for the one that uses raw Linux system calls. I'd also like to use this set of traits when implementing the `kqueue` system. This PR moves them to the `utils` module.


sunfishcode:
Cool, this is good to do! Instead of src/utils.rs, would it work to add a new src/io/epoll.rs module, which re-exports the public things from the backend epoll modules, as well as defining the common context types?

As a separate matter, I'm not super happy with rustix's current epoll API. It is usable in some use cases, but it's turned out to be awkward in others, so I'm open to ideas for what to do with it. One option I've been thinking about would be to make it `unsafe` and have it operate on `RawFd`'s. That would avoid the need for "Owning" and "Borrowing" and related things. But I'm open to other ideas too.

notgull:
A couple of ideas:

- Add a new context that operates on `RawFd` and make it unsafe to instantiate.
- Add a context that forces the `Epoll` to be `'static`, and only takes a wrapper that deletes the source from the poller once it's dropped.

notgull:
> Cool, this is good to do! Instead of src/utils.rs, would it work to add a new src/io/epoll.rs module, which re-exports the public things from the backend epoll modules, as well as defining the common context types?

I decided to move it to `src/io/context.rs` instead, since I plan on reusing parts of it for `kqueue`.

PR #468, opened 2022-11-30
Don't check the `errno` value from `isatty`.

### LINKED ISSUE #467: `isatty` is not required to set errno
Hi! We noticed that the nightly version of `cargo` is panicking inside rustix on illumos:

```
$ RUST_BACKTRACE=1 cargo +nightly --version | cat
thread 'main' panicked at 'unexpected error from isatty: 0', /cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.36.3/src/backend/libc/termios/syscalls.rs:141:20
stack backtrace:
   0: rust_begin_unwind
             at /rustc/e0098a5cc3a87d857e597af824d0ce1ed1ad85e0/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/e0098a5cc3a87d857e597af824d0ce1ed1ad85e0/library/core/src/panicking.rs:65:14
   2: rustix::backend::termios::syscalls::isatty
   3: <cargo::core::shell::Shell>::new
   4: <cargo::util::config::Config>::default
   5: <cargo::cli::LazyConfig>::get_mut
   6: cargo::cli::main
   7: cargo::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Digging in a bit, we found that the issue is that illumos currently chooses not to set errno in `isatty()`, but rustix [expects errno to be set to ENOTTY](https://github.com/bytecodealliance/rustix/blob/main/src/backend/libc/termios/syscalls.rs#L148-L149) if `isatty()` returns 0, and panics otherwise. I think illumos's behavior here is not ideal (and we're also filing a bug there), but is technically allowed by the wording on the [POSIX manpage](https://pubs.opengroup.org/onlinepubs/9699919799/functions/isatty.html) (emphasis mine):

> The isatty() function shall return 1 if fildes is associated with a terminal; otherwise, it shall return 0 and **may** set errno to indicate the error.

Would you be opposed to changing the implementation of `isatty()` to only look at the return value and not errno?

sunfishcode:
We don't actually do anything different for different errno values, other than panic on unknown ones, and that isn't that isn't adding much value compared to the cost of being an extra surprise when porting to new OS's.

Fixes #467.

### DIFF
--- a/src/backend/libc/termios/syscalls.rs
+++ b/src/backend/libc/termios/syscalls.rs
@@ -10,13 +10,13 @@ use crate::fd::BorrowedFd;
 #[cfg(feature = "procfs")]
 #[cfg(not(any(target_os = "fuchsia", target_os = "wasi")))]
 use crate::ffi::CStr;
+#[cfg(not(target_os = "wasi"))]
 use crate::io;
 #[cfg(not(target_os = "wasi"))]
 use crate::process::{Pid, RawNonZeroPid};
 #[cfg(not(target_os = "wasi"))]
 use crate::termios::{Action, OptionalActions, QueueSelector, Speed, Termios, Winsize};
 use core::mem::MaybeUninit;
-use libc_errno::errno;
 
 #[cfg(not(target_os = "wasi"))]
 pub(crate) fn tcgetattr(fd: BorrowedFd<'_>) -> io::Result<Termios> {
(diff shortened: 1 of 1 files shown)


ahl:
Any chance we could get a new release with this PR included? Thanks!

sunfishcode:
This is now released in 0.36.5!

sunshowers:
Thanks @sunfishcode!

PR #469, opened 2022-12-03
Future-proof for FreeBSD 12

asomers:
FreeBSD 12 changes the dirent structure, among other things.  libc currently binds a FreeBSD 11 ABI, but that will change some day.  Tweak rustix's dirent initialization code to work with either FreeBSD 11 or 12.


sunfishcode:
Thanks!

Issue #472, opened 2022-12-05
URGENT: patch out RawDir for the next release too


SUPERCILEX:
@sunfishcode pretty sure I done goofed again. Was listening to the ghost cell podcast and Ralf talked about how you can access all of the elements of the Iterator at the same time thanks to the non-aliasing references. Anyway, I was like, "doesn't raw dir work that way since the lifetime is bound to the buffer?" And then: "Oh shit, we reuse that buffer."

TL;DR: pretty sure people can accumulate entry references into some other structure, but they'll be garbage once the buffer restarts. Need to double check tomorrow. Somehow need lifetime that is bound to the buffer so it can't be dropped but also blocks aliasing so every call to next requires dropping the previous entry.

SUPERCILEX:
Preliminary research indicates that GATs are required. Need to investigate further.

sunfishcode:
Thanks for the heads up! I've now created a 0.36 branch, so we can do 0.36 releases without pulling in the `RawDir` changes for now.

SUPERCILEX:
Alrighty, I put up a patch.

I think we finally got this dang thing locked down. Safety is so hard lol. :sweat_smile:

PR #474, opened 2022-12-05
Fix another UB in raw dir

SUPERCILEX:
This kind of sucks, but we can't use for loops until `LendingIterator` becomes a thing in the stdlib. For now, I elected to simply not implement the `Iterator` interface and make people manually call next.

Code that compiled before:
```rust
let mut buf = Vec::with_capacity(1000);
let mut other = Vec::new();
for x in RawDir::new(cwd(), buf.spare_capacity_mut()) {
    other.push(x.unwrap());
}
```

Code that no longer compiles:
```rust
let mut buf = Vec::with_capacity(1000);
let mut other = Vec::new();
let mut iter = RawDir::new(crate::fs::cwd(), buf.spare_capacity_mut());
while let Some(entry) = iter.next() {
    other.push(entry.unwrap());
}
```

Closes https://github.com/bytecodealliance/rustix/issues/472

### DIFF
--- a/src/fs/raw_dir.rs
+++ b/src/fs/raw_dir.rs
@@ -45,7 +45,8 @@ impl<'buf, Fd: AsFd> RawDir<'buf, Fd> {
     /// let fd = openat(cwd(), ".", OFlags::RDONLY | OFlags::DIRECTORY, Mode::empty()).unwrap();
     ///
     /// let mut buf = Vec::with_capacity(8192);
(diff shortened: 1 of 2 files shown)


SUPERCILEX:
> Do you expect the API will be evolving beyond this?

Beyond being usable in a for loop when that becomes possible? No. And I'm hopeful that can actually be backwards compatible.

sunfishcode:
Makes sense, thanks!

PR #476, opened 2022-12-06
Make termios constants support AIX

ecnelises:
Mark those related constants as unsupported on AIX.


sunfishcode:
Thanks! I know those `cfg`s are really tricky to get right. I think this is a case where having src/termios/mod.rs use `use constants::*` and having constants.rs do `use backend::termios::types::*` would significantly simplify this. I'll do this hopefully soon, but alternatively feel free to make this change yourself if you'd like.

ecnelises:
I posted https://github.com/bytecodealliance/rustix/pull/543

sunfishcode:
#543 is now merged, so this PR can now be simplified accordingly :-).

sunfishcode:
I took the liberty of resolving the merge conflicts here.

ecnelises:
Thanks!

PR #479, opened 2022-12-08
Implement inotify API

carbotaniuman:
More of a Draft PR to test in CI right now, as it's missing documentation and tests. It's also waiting on a libc PR to add some more constants so I can grab them.

### DIFF
--- a/src/backend/libc/fs/inotify.rs
+++ b/src/backend/libc/fs/inotify.rs
@@ -1,0 +1,116 @@
+//! inotify support for working with inotifies
+
+use super::super::c;
+use super::super::conv::{borrowed_fd, c_str, ret, ret_c_int, ret_owned_fd};
+use crate::fd::{BorrowedFd, OwnedFd};
+use crate::io;
+use bitflags::bitflags;
+
+bitflags! {
+    /// `IN_*` for use with [`Inotify::new`].
+    pub struct CreateFlags: c::c_int {
+        /// `IN_CLOEXEC`
+        const CLOEXEC = c::IN_CLOEXEC;
+        /// `IN_NONBLOCK`
+        const NONBLOCK = c::IN_NONBLOCK;
+    }
+}
+
+bitflags! {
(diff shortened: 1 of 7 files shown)


sunfishcode:
Looks like ubuntu-latest just updated, and the prebuilt archives need to be updated. I'll fix this.

sunfishcode:
Closing and reopening to restart CI now that the CI issues are fixed.

src/backend/libc/fs/inotify.rs
```diff
@@ -0,0 +1,169 @@
+//! inotify support
+//! todo
```
sunfishcode, line 2: Can you elaborate on this "todo" comment?

src/backend/libc/fs/inotify.rs
```diff
@@ -0,0 +1,169 @@
...
+        /// `IN_ONLYDIR`
+        const ONLYDIR = c::IN_ONLYDIR;
+    }
+}
+
+/// An inotify, an interface to an OS object allowing one to
+/// watch a filesystem for changes.
+pub struct Inotify {
+    inotify_fd: OwnedFd,
+}
```
sunfishcode, line 76: Rustix's current API style is to expose functions that operate on `OwnedFd`/`BorrowedFd`/`AsFd` directly, rather than wrapping them in structs like this. The current theory is that this gives users more theoretical flexibility at the rustix level, and that most use cases will end up wrapping rustix with a higher-level API which tends to be a more advantageous abstraction level at which to have type distinctions. For example, if someone wanted to `dup2` something over an inotify file descriptor, having it wrapped up in a struct like this would make that awkward.

carbotaniuman:
Seems like CI has some rate-limit issues. I've modified the API to remove the struct and styled the organization similarly to epoll, but not sure if that's the correct move.

sunfishcode:
I reran the CI and it now passes. The FreebBSD failure is unrelated.

PR #483, opened 2022-12-09
Update archives for Ubuntu 22.04.

sunfishcode:
Github's ubuntu-latest is now on Ubuntu 22.04, which has a newer binutils, which produces slightly different prebuilt libraries. Update the checked-in libraries to match the builds produced in CI.


sunfishcode:
The problem here turned out to be a mismatch in the dwarf version, which the cc crate overrides to attempt to match the Rust version. I fixed it by pinning the builds to -gdwarf-3, which is the version used in Rust 1.48.

PR #486, opened 2022-12-13
Support `SEEK_DATA` and `SEEK_HOLE` for `lseek`

### LINKED ISSUE #480: Sparse file support for `lseek`?
On Linux and some other POSIX systems, [`lseek`](https://man7.org/linux/man-pages/man2/lseek.2.html) supports two additional values for `whence` relating to sparse files, `SEEK_HOLE` and `SEEK_DATA`. I believe these should be supported by `rustix`.

For reference, some filesystems have files with large amounts of zero data that can be optionally marked "sparse", and these sections won't be allocated on disk. This feature can be exposed to userspace applications to improve the performance of some operations, like `cp`.

Although this is technically not a part of the POSIX standard, it has been proposed for issue 8.

As an example of prior art, [`nix` already exposes this ability](https://docs.rs/nix/latest/nix/unistd/fn.lseek.html).

sharifhsn:
Fixes #480 

Adds two new variants to the `SeekFrom` API, `Data` and `Hole`.

Some questions:
- Are the docs fine? They're basically copied from [the man page](https://man7.org/linux/man-pages/man2/lseek.2.html).
- Should the `staged_api` attribute and other aspects be used? I'm not sure what their purpose is but I included it just in case.

### DIFF
--- a/src/io/seek_from.rs
+++ b/src/io/seek_from.rs
@@ -32,8 +32,7 @@ pub enum SeekFrom {
     /// specified number of bytes that contains data.
     ///
     /// If the offset points to data, the file offset is set to it.
-    #[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))]
-    Data(#[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))] i64),
+    Data(i64),
 
     /// Sets the offset to the next hole in the file greater than or equal to the
     /// specified number of bytes.
@@ -43,6 +42,5 @@ pub enum SeekFrom {
     ///
     /// If there is no hole past offset, then the file offset is adjusted to the end
     /// of the file (i.e., there is an implicit hole at the end of any file).
-    #[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))]
-    Hole(#[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))] i64),
+    Hole(i64),
 }
--- a/src/io/seek_from.rs
+++ b/src/io/seek_from.rs
@@ -27,4 +27,22 @@ pub enum SeekFrom {
     /// to seek before byte 0.
     #[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))]
     Current(#[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))] i64),
+
+    /// Sets the offset to next location in the file greater than or equal to the
+    /// specified number of bytes that contains data.
+    ///
+    /// If the offset points to data, the file offset is set to it.
+    #[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))]
+    Data(#[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))] i64),
+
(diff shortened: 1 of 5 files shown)


src/io/seek_from.rs
```diff
@@ -27,4 +27,22 @@ pub enum SeekFrom {
     /// to seek before byte 0.
     #[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))]
     Current(#[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))] i64),
+
+    /// Sets the offset to next location in the file greater than or equal to the
+    /// specified number of bytes that contains data.
+    ///
+    /// If the offset points to data, the file offset is set to it.
+    #[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))]
+    Data(#[cfg_attr(staged_api, stable(feature = "rust1", since = "1.0.0"))] i64),
```
sunfishcode, line 36: The `cfg_attr` things in the other arms are there because this is code derived from std, which has these stability feature attributes. For our own arms, `Data` and `Hole`, we can leave of these `#[cfg_attr(...*)]` attributes entirely.

sunfishcode (commented):
>     * Are the docs fine? They're basically copied from [the man page](https://man7.org/linux/man-pages/man2/lseek.2.html).

I'd like to avoid just copying the text like this; could you instead either write your own text, or make it a quote with a citation?

And actually, looking at the text, I'm having trouble parsing "Sets the offset to next location in the file greater than or equal to the specified number of bytes that contains data." Would it be correct to rephrase it along the lines of "seeks forward from the current position by the specified number of bytes, and then if that's in a hole, seeks forward until it reaches data" or so?

>     * Should the `staged_api` attribute and other aspects be used? I'm not sure what their purpose is but I included it just in case.

Those are there because that file is derived from a file in std, and it's easier to keep up to date if we keep the diffs minimal. For these new values, we don't need to add them.

sharifhsn:
> Would it be correct to rephrase it along the lines of "seeks forward from the current position by the specified number of bytes, and then if that's in a hole, seeks forward until it reaches data" or so?

Yes, I think this would be much clearer. My only concern is if it's preferable that the wording of the documentation remain consistent across `SeekFrom` variants?

sunfishcode:
How about, "Sets the offset to the current position plus the specified number of bytes, plus the distance to the next byte which is not in a hole."?

sunfishcode:
The change here looks good to me.

That said, it looks like this change needs a semver bump, because users using `std::io::SeekFrom` will need to switch to rustix's. I'm hoping to avoid doing a semver bump in rustix for a while. If you'd like this feature to show up in a release sooner, what would you think of renaming `seek` to `seek_ext` or something, and re-introducing the old `seek` that takes a `std::io::SeekFrom`, so that it's semver-compatible?

sharifhsn:
I'm not pressed for this feature currently, I just was interested in its absence in the API. My use for it is in an experimental PR which will not be merged soon, if ever. This should probably be placed in some kind of `0.36` milestone or something.

sunfishcode:
Sounds good. I've now added a 0.37 label and labelled this PR with it.

sunfishcode:
#487 landed, which will need a sember bump, so let's land this too.

PR #487, opened 2022-12-19
Remove the context API from epoll

### LINKED ISSUE #477: Rework the epoll API to prepare for future polling APIs
The current `Epoll` API is *good*; however, it's not compatible with the idea of a [static reactor](https://github.com/smol-rs/async-io/blob/8e1fa4346382d9b23f59122a0c79fa33eb75c48b/src/reactor.rs#L70). With the current "context" system, you have to choose between a `BorrowedFd` (which is difficult to do for `'static` without leakage) or an `OwnedFd` (which makes the source inaccessible for the duration of the wait). I filed #471; however, that's a band-aid solution at best. It would be nice to be able to rework the API so that it can be stored in a `'static` reactor, while also not forcing every source that interacts with it to be `'static` or owned. Given that #472 will probably lead to a breaking change, it's a good time to talk about this.

The idea I had was something to the tune of a `Source` structure, like this:

```rust
pub struct Source<T> {
    /// The actual data.
    data: T,

    /// Reference to the `Epoll` file descriptor.
    epoll: &'static EpollRef,
}
```

An object can become a `Source` by registering itself into the `Epoll`/`Kqueue`/whatever, and dropping the `Source` or moving the data out of the `Source` causes it to be removed from the poller. This forces the poller to be static, but with the `Context` system we can make it optional in this case.

Then we can rework the `Context` trait to take a lifetime, like so:

```rust
pub trait Context<'a> {
    fn encode(&'a self, ...) { ... }
}

impl<'a> Context<'a> for Borrowed<'a> { ... }
impl Context<'static> for SourceContext { ... }
```

...so that we can be sure that our context that uses `Source`s is always `'static`.

At this point, I'm not sure where to go from here. I tried to implement this pattern on top of the current system, but I ran into issues on how to `acquire` a `Source` without taking ownership of it or constricting it to a lifetime.

notgull:
It looks like we were overthinking it; from a brief glance at the documentation, it looks like deleting a file descriptor registered in `epoll` is not undefined behavior, at least by Rust's specifications. This PR removes the `Context` API and just uses the `BorrowedFd` API.

Closes #477


notgull:
From a semantic point of view, I like to think of a file descriptor as an `Arc<dyn SystemResource>`. And, in this call, the `Arc` is cloned. In this case, it would be valid under the Rust borrow system.

sunfishcode:
The equivalent of an Arc clone is a dup, but EPOLL_CTL_ADD isn't a dup, and I expect this is observable in various ways. For example, events will contain the file descriptors' numbers, so it's observable that they "escaped" the borrow without a dup.

We may be able to do this, but I want to think through the consequences more.

notgull:
The events themselves actually don't contain the file descriptor; it's just the event number that was passed in to the event.

(This may be an issue for `kqueue`, however.)

src/backend/linux_raw/io/epoll.rs
```diff
@@ -229,84 +211,68 @@ impl<Context: self::Context> Epoll<Context> {
...
-    context: *const Context,
-    _phantom: PhantomData<&'context Context>,
+pub struct Iter<'a> {
+    iter: core::slice::Iter<'a, Event>,
 }
 
-impl<'context, Context: self::Context> Iterator for Iter<'context, Context> {
-    type Item = (EventFlags, Ref<'context, Context::Target>);
+impl<'a> Iterator for Iter<'a> {
+    type Item = (EventFlags, u64);
```
sunfishcode, line 270: If we're going to hand the raw `u64` back to the user, we should rename the `encoded` field to just `data`, and drop the `encode`/`decode` functions in the `Context` abstraction.

sunfishcode:
Rustix's epoll API currently does store the file descriptor number in the epoll user data, and provides it to users as a file descriptor along with events. This way users that just need the file descriptor can get it, without having to keep a separate `HashMap` or using raw file descriptors and unsafe.

That said, I have been thinking about whether it's better to just letting users manage the `data` field, as this PR does, even if it does push people towards `unsafe` in some use cases. The extra flexibility and simplicity may be worth it.

notgull:
I replaced `Epoll` with `OwnedFd`, as mentioned in https://github.com/bytecodealliance/rustix/pull/488#discussion_r1054997533

notgull:
@sunfishcode Is there any chance you can review this?

> Rustix's epoll API currently does store the file descriptor number in the epoll user data, and provides it to users as a file descriptor along with events. This way users that just need the file descriptor can get it, without having to keep a separate `HashMap` or using raw file descriptors and unsafe.

In my opinion, this abstraction is too high level for `rustix`. If users want to keep track of data associated with the `data` field, it should be their decision to do so.

.gitignore
```diff
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+/.vscode
```
sunfishcode, line 3: Please don't add editor-specific entries to `.gitignore`; users of VSCode should add this to their global git config instead.

sunfishcode (commented):
> In my opinion, this abstraction is too high level for `rustix`. If users want to keep track of data associated with the `data` field, it should be their decision to do so.

I agree. I think this PR sounds like the right approach overall. The current Epoll abstraction was an attempt to provide an API that wouldn't require users to use unsafe, but it ended up getting too complex. This does appear to be a disadvantageous abstraction level for safety.

I'm still concerned about epoll holding onto file descriptors and reporting events for them after the scope of their borrow ends. There's no direct use-after-close hazard, but it seems error-prone. What would you think about making `epoll_add` unsafe, having it take an `AsRawFd`, and documentingbthat it behaves as if the raw fd is retained?

Also, tagging with 0.37, as this is a semver bump.

notgull:
> I'm still concerned about epoll holding onto file descriptors and reporting events for them after the scope of their borrow ends. There's no direct use-after-close hazard, but it seems error-prone. What would you think about making `epoll_add` unsafe, having it take an `AsRawFd`, and documentingbthat it behaves as if the raw fd is retained?

I don't think there's much danger here. The worst thing that can happen is a spurious event (which is pretty much expected to happen with leaky behavior like this anyways). I've added documentation to `epoll_add` indicating this, but I don't personally think it's enough to mark this as `unsafe`.

sunfishcode:
I started a thread about this in Zulip: https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/I.2FO.20safety.20for.20epoll . @bjorn3 suggested thinking of epoll as doing a "weak dup", which I'm thinking makes sense. If you want to write some documentation along those lines feel free, though I'd be happy to do a follow-up PR myself.

src/backend/libc/io/epoll.rs
```diff
@@ -337,24 +264,20 @@ struct Event {
     // the full union; `Context` implementations will simply need to deal with
     // casting the value into and out of the `u64` themselves.
     event_flags: EventFlags,
-    encoded: u64,
+    data: u64,
```
sunfishcode, line 267: Could we make `Event` public now and remove `EventVec` entirely, and have users use `Vec<Event>`?

sunfishcode (commented):
This needs a rustfmt for.CI, and I notice one other simplification we can make now:

src/backend/libc/io/epoll.rs
notgull: I don't think we can, since `Event` is `repr(packed)` and that's hard to deal with in safe Rust.

sunfishcode: Ah, makes sense.

sunfishcode:
Sounds good!

PR #488, opened 2022-12-19
Implement the kqueue API

notgull:
This implements the `kqueue` API, available on BSD and MacOS.

The event API is something I'm not entirely happy with; I am very open to suggestions here.

### DIFF
--- a/src/backend/libc/io/syscalls.rs
+++ b/src/backend/libc/io/syscalls.rs
@@ -33,6 +33,18 @@ use core::ptr;
 #[cfg(all(feature = "fs", feature = "net"))]
 use libc_errno::errno;
 
+#[cfg(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "watchos",
+    target_os = "freebsd",
+    target_os = "dragonfly",
+    target_os = "openbsd",
+    target_os = "netbsd"
+))]
+use crate::io::kqueue::Event;
+
 pub(crate) fn read(fd: BorrowedFd<'_>, buf: &mut [u8]) -> io::Result<usize> {
     let nread = unsafe {
         ret_ssize_t(c::read(
@@ -468,6 +480,46 @@ pub(crate) fn ioctl_tiocexcl(fd: BorrowedFd) -> io::Result<()> {
 #[cfg(not(any(target_os = "haiku", target_os = "redox", target_os = "wasi")))]
 pub(crate) fn ioctl_tiocnxcl(fd: BorrowedFd) -> io::Result<()> {
     unsafe { ret(c::ioctl(borrowed_fd(fd), c::TIOCNXCL as _)) }
+}
+
+#[cfg(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "watchos",
+    target_os = "freebsd",
+    target_os = "dragonfly",
+    target_os = "openbsd",
+    target_os = "netbsd"
+))]
+pub(crate) fn kqueue() -> io::Result<OwnedFd> {
+    unsafe { ret_owned_fd(c::kqueue()) }
+}
+
+#[cfg(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "watchos",
+    target_os = "freebsd",
+    target_os = "dragonfly",
+    target_os = "openbsd",
+    target_os = "netbsd"
+))]
+pub(crate) unsafe fn kevent(
+    kq: BorrowedFd<'_>,
+    changelist: &[Event],
+    eventlist: &mut [MaybeUninit<Event>],
+    timeout: Option<&c::timespec>,
+) -> io::Result<c::c_int> {
+    ret_c_int(c::kevent(
+        borrowed_fd(kq),
+        changelist.as_ptr() as *const _,
+        changelist.len() as _,
+        eventlist.as_mut_ptr() as *mut _,
+        eventlist.len() as _,
+        timeout.map_or(core::ptr::null(), |t| t as *const _),
+    ))
 }
 
 #[cfg(not(target_os = "wasi"))]
--- a/src/io/kqueue.rs
+++ b/src/io/kqueue.rs
@@ -1,0 +1,448 @@
+//! An API for interfacing with `kqueue`.
+
+use crate::backend;
+use crate::fd::{AsFd, AsRawFd, OwnedFd, RawFd};
+use crate::io;
+
+use backend::c::{self, kevent as kevent_t, uintptr_t};
+use backend::io::syscalls;
+
+#[cfg(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "watchos",
+    target_os = "dragonfly",
+    target_os = "freebsd",
+))]
+use backend::c::intptr_t;
+
+use alloc::vec::Vec;
+use core::ptr::slice_from_raw_parts_mut;
+use core::time::Duration;
+
+/// A `kqueue` event.
+#[repr(transparent)]
+#[derive(Copy, Clone)]
+pub struct Event {
+    // The layout varies between BSDs and macOS.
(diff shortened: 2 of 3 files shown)


src/io/kqueue.rs
```diff
@@ -0,0 +1,566 @@
...
+impl Kqueue {
+    /// Create a new `Kqueue`.
+    pub fn new() -> io::Result<Self> {
+        syscalls::kqueue().map(|fd| Self { fd })
+    }
+
+    /// Wait for events on this queue.
+    pub fn kevent(
+        &self,
+        changelist: &[InEvent<'_>],
```
sunfishcode, line 448: I'm not super familiar with the `kqueue` API, but as I understand it, the changelist registers file descriptors which are associated with the kqueue file descriptor, which may outlive the `&self` and ``` `_ ``` lifetimes here, so this lifetime description doesn't fully describe the possible file-descriptor lifetimes.

sunfishcode (commented):
Thanks for working on this! Here are some initial comments from a first readthrough.

src/io/kqueue.rs
notgull: I believe that, if a file descriptor is `close`d while it's registered in `kqueue`, it's also automatically deregistered from `kqueue`, so it shouldn't be a soundness issue, at the very least.

src/io/kqueue.rs
```diff
@@ -0,0 +1,566 @@
...
+    pub fn get(self) -> u32 {
+        self.0
+    }
+}
+
+/// A `kqueue`.
+pub struct Kqueue {
+    /// The underlying `kqueue` file descriptor.
+    fd: OwnedFd,
+}
```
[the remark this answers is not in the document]
notgull: I think it's more idiomatic to have a `Kqueue` object you can call `new()` and `kevent()` on rather than a freestanding `kevent` function, if not also less of a footgun.

src/backend/libc/io/syscalls.rs
```diff
@@ -33,6 +33,18 @@ use core::ptr;
...
+#[cfg(any(
+    target_os = "macos",
+    target_os = "ios",
+    target_os = "tvos",
+    target_os = "watchos",
+    target_os = "freebsd",
+    target_os = "dragonfly",
+    target_os = "openbsd",
+    target_os = "netbsd"
+))]
```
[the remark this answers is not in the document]
notgull: I'll write a future PR for this

src/io/kqueue.rs
sunfishcode: The same thing could be said about directories, socket listeners, socket connections, udp sockets, signalfd, eventfd, and other kinds of file descriptors too though. Should rustix strive to have static types for every "type" of file descriptor?

Currently, rustix's approach is to assume that the bulk of the Rust code in the world won't be using rustix directly. Abstractions like std, cap-std, etc. provide these kinds of static type distinctions, in `File`, `ReadDir`, `TcpListener`, and so on. With that, rustix then focuses on just being a simple low-level API that isn't opinionated about which file descriptors can be used with which system calls. Maybe someone wants to `dup2` a file descriptor over a `kqueue` file descriptor. That's awkward to do if there's a `Kqueue` wrapper struct. And perhaps there are other things users want to do, that we can't anticipate. So rustix so far has erred on the side of letting users pass any file descriptor to any OS system call, because it isn't memory-unsafe, and in practice it hasn't been observed to be a problem in practice, and maybe there's a use case that rustix doesn't anticipate which users *do* want to do.

sunfishcode: (Oops, that comment should have been in reply to the other thread about the `Kqueue` type.)

notgull: Thanks for the explanation! I've changed it so that there is just a freestanding `kqueue` and `kevent` function pair.

notgull:
@sunfishcode Could you take another look at this? I've gotten the event API to a place where I like it.

sunfishcode:
Yes, I apologize for being slow here; I am looking forward to reviewing this! I'm not familar with the kqueue API so I need some time to read up on that, and I've been busy with holiday and real-life stuff, but I hope to get to this soon.

src/io/kqueue.rs
```diff
@@ -0,0 +1,558 @@
...
+
+/// `kqueue()`- Create a new `kqueue` file descriptor.
+pub fn kqueue() -> io::Result<OwnedFd> {
+    syscalls::kqueue()
+}
+
+/// `kevent()`- Wait for events on a `kqueue`.
+pub fn kevent(
+    kqueue: impl AsFd,
+    changelist: &[InEvent<'_>],
```
sunfishcode, line 472: This changelist contains borrowed fds that get registered with the kqueue. What happens if they are closed while they're still registered? And, is there any way to read the fd values back from the kqueue?

notgull: > What happens if they are closed while they're still registered?

The events will be removed, similarly to how `epoll` does it. From the [man page](https://man.openbsd.org/kqueue.2):

> Calling [close(2)](https://man.openbsd.org/close.2) on a file descriptor will remove any kevents that reference the descriptor.

> And, is there any way to read the fd values back from the kqueue?

In the `FilterKind` structure (accessible via the `OutEvent` structure), the `RawFd` is accessible. I'm not sure if this was the best decision, and I am willing to remove this method from `OutEvent` if you think that it's too much of a footgun.

sunfishcode: Does it remove the fd from the kqueue on the close of the file descriptor or the file description? The man page says:

> Calling close() on a file descriptor will remove any kevents that reference the descriptor.

makes it sound like it's the file descriptor, but later it says:

> Events which are attached to file descriptors are automatically deleted on the last close of the descriptor.

and the phrase "last close" doesn't make sense for file descriptors because they can only be closed once, but it could make sense for file descriptions.

I ask because on Linux epoll we're using a kind of "weak dup" concept to describe what it does, which works because Linux explicitly works in terms of the file descriptions, making a "weak dup" make sense. But if `kqueue` here is working in terms of the file descriptor, we may need to say something different.

> In the FilterKind structure (accessible via the OutEvent structure), the RawFd is accessible. I'm not sure if this was the best decision, and I am willing to remove this method from OutEvent if you think that it's too much of a footgun.

If the underlying `kqueue` API provides the fd there, then I think Rustix should too. I wasn't happy with the old epoll API being too restrictive, and would like to move toward having APIs providing the full expressivity of the underlying system, even if that requires being `unsafe` in some places.

If `kevent` is causing the queue to hold on to the file *descriptor*, then I don't think we can call it a "weak dup", and my concern is that it's holding onto the file descriptor past the lifetime of the `BorrowedFd`. My initial reaction here is that `EventFilter` should switch from `BorrowedFd` to `RawFd` to reflect the dynamic lifetime of these fds. That would require `kevent` to be `unsafe`; is that too burdensome?

notgull: > My initial reaction here is that `EventFilter` should switch from `BorrowedFd` to `RawFd` to reflect the dynamic lifetime of these fds. That would require `kevent` to be `unsafe`; is that too burdensome?

I'd be fine with this; my motivating use case (`polling`) is currently stuck on `RawFd` anyways. I agree with being rigorous about borrow checking. Although, in terms of future proofing, if `kqueue` is holding on to the file descriptor, do we want to add a lifetime parameter somewhere in this process? If we add a function that takes `RawFd` we may want to add one that takes `BorrowedFd` in the future, even in the latter is restrictive on lifetimes.

notgull: From a look at the [`kqueue` source code](https://github.com/freebsd/freebsd-src/blob/dcfa3ee44da2b139f51a8aedb0f55735c6dfe3f3/sys/kern/kern_event.c#L1595-L1598), it looks like it holds the file descriptor itself.

sunfishcode: Great, thanks for looking into that!

sunfishcode (commented):
I apologize again for being slow here. Thanks a lot for working on this!

Issue #493, opened 2022-12-28
Capabilities functions names are super inconsistent


SUPERCILEX:
I know deprecating stuff just for renames sucks, but the current names are way too inconsistent IMO. I'd propose the following renames (with unchanged APIs included for comparison):
- `get_keep_capabilities`
- `set_keep_capabilities`
- `capabilities_secure_bits` -> `get_capabilities_secure_bits`
- `set_capabilities_secure_bits`
- `capability_is_in_ambient_capability_set` -> `is_in_ambient_capability_set`
- `configure_capability_in_ambient_capability_set` -> `configure_ambient_capability_set`
- `remove_capability_from_capability_bounding_set` -> `remove_from_capability_bounding_set`
- `is_in_capability_bounding_set`
- `clear_ambient_capability_set`

sunfishcode:
These sound reasonable to me. @koutheir, do you have any preferences about these?

If we like these names, I think it's fine to do the deprecation dance.

koutheir:
I disagree about the renames:
- I like expressive names, because we read code more than we write code, and more than we read documentation.
- I don't add `get_` prefixes when they're obvious from the name and the prototype.
- Some of the renames above make it unclear what is being checked or manipulated, and that hurts readability.

If you think the naming is inconsistent, then think about it as an opportunity to make the rest of the names *better* (yes, that is my subjective opinion), instead of lowering to a common denominator.

SUPERCILEX:
@koutheir I thought these names were an improvement (writing capability 3 times is pretty unnecessary IMO), but I won't complain as long as they're consistent. What do you think of these?
- `get_keep_capabilities` -> `keep_capabilities`
- `set_keep_capabilities`
- `capabilities_secure_bits`
- `set_capabilities_secure_bits`
- `capability_is_in_ambient_capability_set`
- `configure_capability_in_ambient_capability_set`
- `remove_capability_from_capability_bounding_set` -> `remove_capability_from_bounding_capability_set`
- `is_in_capability_bounding_set` -> `capability_is_in_bounding_capability_set`
- `clear_ambient_capability_set`

koutheir:
> `get_keep_capabilities` -> `keep_capabilities`

Removing the `get_` prefix from `get_keep_capabilities` changes the meaning of the function name because `keep` is a verb. A function named `keep_capabilities` seems to change something in order to keep capabilities, which is **not** what `get_keep_capabilities` does. I thus disagree with this renaming.

[...]

PR #500, opened 2023-01-06
Rename capabilities functions to be consistent

SUPERCILEX:
Partial https://github.com/bytecodealliance/rustix/issues/493

I didn't rename `capabilities_secure_bits` yet because that's the one we don't seem to have agreement on.

Also did we say we actually wanted to go through a round of deprecation or just pull the plug? I just renamed things, though I can easily add back the old names as deprecated functions.

### DIFF
--- a/src/thread/prctl.rs
+++ b/src/thread/prctl.rs
@@ -387,7 +387,7 @@ pub enum Capability {
 ///
 /// [`prctl(PR_CAPBSET_READ,...)`]: https://man7.org/linux/man-pages/man2/prctl.2.html
 #[inline]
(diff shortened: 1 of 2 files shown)


sunfishcode:
I don't have an opinion about these names myself. @koutheir do these names look good?

koutheir:
> do these names look good?

The renames I see now seem reasonable to me.

> Also did we say we actually wanted to go through a round of deprecation or just pull the plug? I just renamed things, though I can easily add back the old names as deprecated functions.

See: https://github.com/bytecodealliance/rustix/issues/493#issuecomment-1366989927

Issue #496, opened 2023-01-03
0.37 release planning


sunfishcode:
With #487 and #486 landed, we have an upcoming need for a semver bump to 0.37. Please post any questions or concerns here.

If they're fixed soon, it'd be nice to pull on the changes for #495 and/or #493 here too.

notgull:
I was hoping that I could get #488 into shape as well, as well as implement a passable API for Solarish's event ports API so that this crate can be used in `polling` for all of the *nix backends.

Also, if we're doing breaking changes, is there a way that we can adjust [`dup2`](https://docs.rs/rustix/latest/rustix/io/fn.dup2.html) and [`dup3`](https://docs.rs/rustix/latest/rustix/io/fn.dup3.html)'s APIs? Having to take an `OwnedFd` as a parameter hurts them, especially in the case of `dup3`, where having an option for `CLOEXEC` is beneficial.

SUPERCILEX:
Regarding dup, Alan noticed this too: https://github.com/nix-rust/nix/pull/1863#discussion_r1039129730. I couldn't figure out a good way to both require ownership and not force you to pass in the `OwnedFd`, so I ended up adding
```rust
fn dup2_raw<Fd1: AsFd, Fd2: AsRawFd + IntoRawFd>(oldfd: Fd1, newfd: Fd2) -> Result<OwnedFd> {
    let res = unsafe { libc::dup2(oldfd.as_fd().as_raw_fd(), newfd.as_raw_fd()) };
    Errno::result(res).map(|_| unsafe { OwnedFd::from_raw_fd(newfd.into_raw_fd()) })
}
```
on top of the existing dup2 (and same for dup3). That kind of sucks, but it lets us have two safe interfaces that can be used at different levels of abstraction. Would be cool if there was a better way though!

---

Regarding #493, I just need a final sign-off and I'll send a PR.

sunfishcode:
By I/O safety, that dup2_raw should be unsafe, because it operates on raw fds (AsRawFd/IntoRawFd are not trustable).

SUPERCILEX:
Even though it will handle closing the FD for you if it's owned? @ids1024 argued the same position, so I'm probably in the minority and would be fine with it being unsafe. I don't see a way to let you create create FDs out of thin air without raw FDs (well you could indirectly with BorrowedFd, but that would be icky), so that's why I landed on that solution.

sunfishcode:
[...]

PR #501, opened 2023-01-06
Gate statx syscall compat wrapper with feature flag

SUPERCILEX:
See discussion: https://github.com/bytecodealliance/rustix/issues/496#issuecomment-1371453474

Extensibility plan:
- Not enabled by default, so users get compat if they don't make any choices.
- Future `linux_x` flags simply depend on the `linux_{x-1}` flag so that we get easy "or later" behavior.


sunfishcode:
I like the idea of making this a compile-time flag, though my understanding of cargo features is that they should [be additive](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification). If a rustix appears multiple times in a programs' dependency tree, and one enables linux_4_11 and the other doesn't, the feature is enabled for both, potentially breaking the one that didn't enable it.

SUPERCILEX:
Yeah, I'm not super familiar with cargo features, but I think `use-libc` is broken in the same way. If one library enables it, everyone will be stuck on libc. AFAIK, libraries are supposed to use the minimal set of features they need and let binaries decided the final feature mix. Unless you were suggesting `linux_x_compat` features? Though I think you run into the same issues.

sunfishcode:
The idea with `use-libc` is that it should be ok if one crate enables it and it gets enabled for all other crates, because the behavior should be the same. It's "just" an optimization. (In practice, there are some observable differences, but they're minor and not things that users are meant to depend on).

Would it work to reverse the sense of the feature here? Instead of disabled-by-default features that disable Linux version support, what about an enabled-by-default feature that enables Linux version support? Users would use `default-features = false, features = ["std", "linux_5_18"]` or so to drop support for old versions. And maybe we could have some checking that at least one Linux version must be enabled, on Linux targets. Would that work?

SUPERCILEX:
> It's "just" an optimization. 

I might be missing something, but that's exactly what this PR is doing. If someone adds the `linux_4_11`, it just means pre-4.11 kernels will take longer to realize the statx call failed, but there are no semantic differences that I can tell.

> Would it work to reverse the sense of the feature here?

I think that would work, but I want to make sure we're on the same page about `linux_4_11` just being an optimization first.

Cargo.toml
```diff
@@ -194,3 +194,11 @@ os_pipe = ["io-lifetimes/os_pipe"]
 fs-err = ["io-lifetimes/fs-err"]
 #all-impls = ["async-std", "tokio", "os_pipe", "socket2", "mio", "fs-err"]
 all-impls = ["os_pipe", "fs-err"]
+
+# OS compatability features
+
+# Build for linux 4.11 or later
```
sunfishcode, line 200: Could you change this comment to say "Optimize for Linux 4.11 or later"? I think that would help clarify that it's still additive in terms of functionality, and it's just about tuning.

Cargo.toml
```diff
@@ -194,3 +194,11 @@ os_pipe = ["io-lifetimes/os_pipe"]
...
+linux_4_11 = []
+
+# Build without backcompat
```
sunfishcode, line 203: Similarly, could this say "Enable all optimizations for the latest Linux versions" or so?

sunfishcode (commented):
Sorry for the slow review here. I think the thing I was missing is that this behaves the same both with and without the feature, but the feature just controls how fast it fails in different situations. So I think this looks good, with just some minor comment tweaks:

SUPERCILEX:
Done!

sunfishcode:
Thanks!

PR #502, opened 2023-01-06
Improve perf of all buffer consuming APIs

SUPERCILEX:
Rustix being a low-level library, it should not have worse perf than the stdlib: https://github.com/rust-lang/rust/blob/0fb8b72ce49997d60a631e921d2cf5be9ca229e6/library/std/src/sys/unix/fs.rs#L1396-L1423

### DIFF
--- a/src/backend/linux_raw/termios/syscalls.rs
+++ b/src/backend/linux_raw/termios/syscalls.rs
@@ -273,7 +273,8 @@ pub(crate) fn isatty(fd: BorrowedFd<'_>) -> bool {
 }
 
 #[cfg(feature = "procfs")]
-pub(crate) fn ttyname(fd: BorrowedFd<'_>, buf: &mut [u8]) -> io::Result<usize> {
+#[allow(unsafe_code)]
+pub(crate) fn ttyname(fd: BorrowedFd<'_>, buf: &mut [MaybeUninit<u8>]) -> io::Result<usize> {
     let fd_stat = crate::backend::fs::syscalls::fstat(fd)?;
 
     // Quick check: if `fd` isn't a character device, it's not a tty.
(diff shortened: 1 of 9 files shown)


SUPERCILEX:
Turns out ttyname depends on readlink, so I just converted all the buffer consuming APIs to the new approach. `getcwd` from the stdlib: https://github.com/rust-lang/rust/blob/0fb8b72ce49997d60a631e921d2cf5be9ca229e6/library/std/src/sys/unix/os.rs#L142-L166

SUPERCILEX:
Well shucks, all of this stuff requires Rust 1.60. I guess we can just let this sit for a few months until Bookworm is released? It currently has 1.63: https://packages.debian.org/testing/rust/rustc

src/backend/linux_raw/termios/syscalls.rs
```diff
@@ -238,14 +239,19 @@ pub(crate) fn ttyname(fd: BorrowedFd<'_>, buf: &mut [u8]) -> io::Result<usize> {
...
     }
-    buf[r] = b'\0';
+    buf[r].write(b'\0');
 
     // Check that the path we read refers to the same file as `fd`.
-    let path = CStr::from_bytes_with_nul(&buf[..=r]).unwrap();
+    {
+        // SAFETY:
+        // - We just wrote the NUL byte above
+        // - readlinkat() returns the number of bytes placed in the buffer
```
sunfishcode, line 248: As a minor style nit, I prefer to see function names code-quoted and not suffixed with `()` unless they actually have zero arguments.  The old `()` convention reminds me too much of old-school K&R C where you could just `()` even if the function has args :-}.

sunfishcode (commented):
Nice! And yeah, I'm really looking forward to bumping the MSRV too, as 1.48 is getting increasingly burdensome, but for our reverse dependencies it would be good to wait a little longer.

src/backend/linux_raw/termios/syscalls.rs
SUPERCILEX: Lol, fixed.

SUPERCILEX:
> for our reverse dependencies it would be good to wait a little longer

Sounds good!

SUPERCILEX:
@sunfishcode Lemme know when master stabilizes and I'll resolve the conflicts here.

sunfishcode:
@SUPERCILEX The 0.38-planning branch is now landed on main, with the MSRV bump, so we're now ready!

sunfishcode:
Thanks!

PR #503, opened 2023-01-07
feat: Expose the Solarish event ports API

notgull:
This PR exposes the Solarish event ports API, which is Solaris' way of handling polling.

### DIFF
--- a/src/backend/libc/io/syscalls.rs
+++ b/src/backend/libc/io/syscalls.rs
@@ -553,3 +553,42 @@ pub unsafe fn vmsplice(
     ))
     .map(|spliced| spliced as usize)
 }
+
+#[cfg(any(target_os = "solaris", target_os = "illumos",))]
+pub fn port_create() -> io::Result<OwnedFd> {
+    unsafe { ret_owned_fd(c::port_create()) }
+}
+
+#[cfg(any(target_os = "solaris", target_os = "illumos",))]
+pub unsafe fn port_associate(
+    port: BorrowedFd<'_>,
+    source: c::c_int,
+    object: c::uintptr_t,
+    events: c::c_int,
+    user: *mut c::c_void,
+) -> io::Result<()> {
+    unsafe {
+        ret(c::port_associate(
+            borrowed_fd(port),
+            source,
+            object,
+            events,
+            user,
+        ))
+    }
+}
+
+#[cfg(any(target_os = "solaris", target_os = "illumos",))]
+pub unsafe fn port_dissociate(
+    port: BorrowedFd<'_>,
+    source: c::c_int,
+    object: c::uintptr_t,
+) -> io::Result<()> {
+    unsafe { ret(c::port_dissociate(borrowed_fd(port), source, object)) }
+}
+
+#[cfg(any(target_os = "solaris", target_os = "illumos",))]
+pub unsafe fn port_get(
+    port: BorrowedFd<'_>,
+    timeout: Option<&c::timespec>,
+) -> io::Result<Option<
--- a/src/backend/libc/io/syscalls.rs
+++ b/src/backend/libc/io/syscalls.rs
@@ -12,6 +12,8 @@ use super::super::offset::{libc_preadv, libc_pwritev};
 #[cfg(all(target_os = "linux", target_env = "gnu"))]
 use super::super::offset::{libc_preadv2, libc_pwritev2};
 use crate::fd::{AsFd, BorrowedFd, OwnedFd, RawFd};
+#[cfg(any(target_os = "solaris", target_os = "illumos"))]
+use crate::io::port::Event;
 #[cfg(not(any(target_os = "aix", target_os = "wasi")))]
 use crate::io::DupFlags;
 #[cfg(not(any(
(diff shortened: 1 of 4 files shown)


src/backend/libc/io/syscalls.rs
```diff
@@ -12,6 +12,8 @@ use super::super::offset::{libc_preadv, libc_pwritev};
 #[cfg(all(target_os = "linux", target_env = "gnu"))]
 use super::super::offset::{libc_preadv2, libc_pwritev2};
 use crate::fd::{AsFd, BorrowedFd, OwnedFd, RawFd};
+#[cfg(any(target_os = "solaris", target_os = "illumos"))]
```
sunfishcode, line 15: To make the cfg's slightly easier to manage, we keep the contents of any/all alphabetized, so illumos should come before solaris.

Another option would be to do a test in build.rs and do `println!("cargo:rustc-cfg=solarish");` so that the code in rustix can just be `cfg(solarish)`.

src/io/port.rs
```diff
@@ -0,0 +1,94 @@
...
+        self.0.portev_object
+    }
+
+    /// Get the userdata associated with this event.
+    pub fn userdata(&self) -> *mut () {
+        self.0.portev_user.cast()
+    }
+}
+
+/// `port_create()`- Creates a new port.
```
sunfishcode, line 32: Here and for the other public functions, can you add References sections to the doc comments, with links to https://illumos.org/man/3C/port_create and similar?

src/io/port.rs
```diff
@@ -0,0 +1,94 @@
...
+pub fn port_create() -> io::Result<OwnedFd> {
+    syscalls::port_create()
+}
+
+/// `port_associate()`- Associates a file descriptor with a port.
```
sunfishcode, line 37: The current comment convention is to include the argument names in the pseudocode comments.

src/io/port.rs
```diff
@@ -0,0 +1,94 @@
...
+
+/// `port_getn()`- Gets multiple events from a port.
+pub fn port_getn(
+    port: &impl AsFd,
+    events: &mut Vec<Event>,
+    min_events: usize,
+    timeout: Option<Duration>,
+) -> io::Result<()> {
+    let mut timeout = timeout.map(|timeout| c::timespec {
+        tv_sec: timeout.as_secs() as _,
```
sunfishcode, line 84: Can the seconds conversion overflow? If so, please use `try_into()` and fail as needed.

src/io/port.rs
```diff
@@ -0,0 +1,94 @@
...
+
+/// `port_create()`- Creates a new port.
+pub fn port_create() -> io::Result<OwnedFd> {
+    syscalls::port_create()
+}
+
+/// `port_associate()`- Associates a file descriptor with a port.
+pub fn port_associate_fd(
+    port: &impl AsFd,
+    object: &impl AsFd,
```
sunfishcode, line 40: Do you know what the behavior is if object is closed while it's associated with a port?

Also, is it possible to get the fd back from the port after this call?

src/io/port.rs
```diff
@@ -0,0 +1,94 @@
...
+        self.0.portev_events
+    }
+
+    /// Get the event source associated with this event.
+    pub fn object(&self) -> usize {
+        self.0.portev_object
+    }
+
+    /// Get the userdata associated with this event.
+    pub fn userdata(&self) -> *mut () {
```
sunfishcode, line 27: I have a mild preference for using `*mut c_void` for this, as it conveys the sense of "pointer to unknown thing" better than `*mut ()` which is a dereferenceable type.

src/backend/libc/io/syscalls.rs
```diff
@@ -531,3 +533,98 @@ pub unsafe fn vmsplice(
     ))
     .map(|spliced| spliced as usize)
 }
+
+#[cfg(any(target_os = "solaris", target_os = "illumos",))]
+pub(crate) fn port_create() -> io::Result<OwnedFd> {
+    unsafe { ret_owned_fd(c::port_create()) }
+}
+
+#[allow(unused_unsafe)]
```
sunfishcode, line 542: Could this omit the unsafe block and obviate the allow(unused_unsafe)?

src/backend/libc/io/syscalls.rs
```diff
@@ -531,3 +533,98 @@ pub unsafe fn vmsplice(
...
+    timeout: Option<&mut c::timespec>,
+) -> io::Result<Option<Event>> {
+    let mut event = MaybeUninit::<c::port_event>::uninit();
+    let timeout = timeout.map_or(core::ptr::null_mut(), |t| t as *mut _);
+    let result = unsafe { ret(c::port_get(borrowed_fd(port), event.as_mut_ptr(), timeout)) };
+
+    // If we timed out, return None.
+    match result {
+        Ok(_) => Ok(Some(Event(unsafe { event.assume_init() }))),
+        Err(e) if e == crate::io::Errno::TIME => Ok(None),
```
sunfishcode, line 584: Is it rustix's job to handle `TIME` here, or should it just let that return to the user to handle?

src/backend/libc/io/syscalls.rs
```diff
@@ -531,3 +533,98 @@ pub unsafe fn vmsplice(
...
+    events: &mut Vec<Event>,
+    nget: usize,
+) -> io::Result<()> {
+    let timeout = timeout.map_or(core::ptr::null_mut(), |t| t as *mut _);
+    let mut nget = nget as _;
+    let result = unsafe {
+        ret(c::port_getn(
+            borrowed_fd(port),
+            events.as_mut_ptr().cast(),
+            events.len() as _,
```
sunfishcode, line 602: It looks like this as cast could overflow.

src/backend/libc/io/syscalls.rs
```diff
@@ -531,3 +533,98 @@ pub unsafe fn vmsplice(
...
+
+#[cfg(any(target_os = "solaris", target_os = "illumos"))]
+pub(crate) fn port_getn(
+    port: BorrowedFd<'_>,
+    timeout: Option<&mut c::timespec>,
+    events: &mut Vec<Event>,
+    nget: usize,
+) -> io::Result<()> {
+    let timeout = timeout.map_or(core::ptr::null_mut(), |t| t as *mut _);
+    let mut nget = nget as _;
```
sunfishcode, line 597: It looks like this as cast could overflow.

src/io/port.rs
notgull: > Do you know what the behavior is if object is closed while it's associated with a port?

Hmm, going through the man pages for OpenSolaris and illumos, I can't find any documentation for what happens. I'm assuming it's the same for `epoll` and `kqueue`, but you know that they say about assumptions.

> Also, is it possible to get the fd back from the port after this call?

The `object` method of the `Event` structure contains the file descriptor. This can't be explicitly the FD because for `EVENT_SOURCE_FILE` it may also be a pointer to a structure describing the file.

src/backend/libc/io/syscalls.rs
notgull: You're right, I've fixed it

src/io/port.rs
sunfishcode: If it's the file descriptor, then I'm nervous about `AsFd`/`BorrowedFd` in the same I am about them for `kqueue`. It's important to me to be rigorous about borrows, even if there's no actual use-after-close hazard, because I'd like to eventually have things like I/O safety checking in miri. As in `kqueue`, what would you think about using `RawFd` and unsafe here?

The thing that hopefully makes this ok is that most users don't use these APIs directly; they're most often used in async runtimes and similar, which have to be carefully written, but then they can offer a higher-level abstraction to their users where it's easier to have safety invariants.

notgull: `polling` uses `RawFd` and `unsafe` for now, and getting it to the point where it wouldn't would require some serious chicanery. I'll edit this PR and the `kqueue` one to use `RawFd`.

notgull:
Spurious network failure :-/ I'll wait a few hours and then push an empty commit to hopefully kick the CI back into gear

sunfishcode:
Looks good, thanks!

PR #504, opened 2023-01-07
Fix Fuchsia target

SUPERCILEX:
Actually it turns out the fuchsia breakage is expected: https://github.com/rust-lang/rust/pull/106429


SUPERCILEX:
Ok, well dunno why this doesn't work.

SUPERCILEX:
Nvm, I realized I'm a dum dum and the triple change is only applicable to nightly for now.

sunfishcode:
Thanks!

PR #505, opened 2023-01-08
feat: Add sendmsg/recvmsg

notgull:
This PR adds the `sendmsg_*`/`recvmsg` family of functions. It also adds structs for dealing with ancillary data. Some of the code here is taken from #159.

Very WIP for now


chrysn:
Great to see this come, recvmsg would make this a potential backend for the async variety of the std-embedded-nal crate.

Given that this is in `net`, and that advertises Winsock2 support, and that there is some recvmsg-like stuff in the Windows APIs (eg. IN_RECVERR, IN_PKTINFO): Do you figure this could support recvmsg even on Windows?

notgull:
Re: both comments above, my motivating use case for this PR is using `rustix` in `x11rb` and `wayland-rs`, which both use `sendmsg`/`recvmsg` as a transport for sending/receiving file descriptors over a Unix socket. Therefore, I'm only really familiar with `SCM_RIGHTS`; I'm not familiar with other types of ancillary data.

Since the `SendAncillaryMessage` and `RecvAncillaryMessage` enums are both `#[non_exhaustive]`, it shouldn't be too hard to add new ancillary data to the enum.

Cargo.toml
```diff
@@ -78,6 +78,7 @@ io-lifetimes = { version = "1.0.10", default-features = false, features = ["clos
 serial_test = "0.6"
 memoffset = "0.8.0"
 flate2 = "1.0"
+cfg-if = "1.0.0"
```
sunfishcode, line 81: I don't see `cfg_if` in use in the code; can this dependency be removed?

Cargo.toml
```diff
@@ -124,13 +125,7 @@ default = ["std", "use-libc-auxv"]
...
 
 # This is used in the port of std to rustix.
-rustc-dep-of-std = [
-    "core",
-    "alloc",
-    "compiler_builtins",
-    "linux-raw-sys/rustc-dep-of-std",
-    "bitflags/rustc-dep-of-std",
-]
+rustc-dep-of-std = ["core", "alloc", "compiler_builtins", "linux-raw-sys/rustc-dep-of-std", "bitflags/rustc-dep-of-std"]
```
sunfishcode, line 128: The resulting line here is 120 columns wide, which is wider than Rust's typical line length limit, making it awkward to read. Could you revert this formatting change?

Cargo.toml
```diff
@@ -172,20 +167,7 @@ rand = []
...
-    "param",
-    "process",
-    "procfs",
-    "rand",
-    "runtime",
-    "termios",
-    "thread",
-    "time",
-]
+all-apis = ["fs", "io_uring", "mm", "net", "param", "process", "procfs", "rand", "runtime", "termios", "thread", "time"]
```
sunfishcode, line 170: Similarly, this creates a very long line, and this line changes over time, and keeping each feature on its own line makes it easier to read diffs.

src/net/send_recv/msg.rs
```diff
@@ -0,0 +1,662 @@
...
+    type Item = T;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // See if there is a next item.
+        if self.data.len() < size_of::<T>() {
+            return None;
+        }
+
+        // Get the next item.
+        let item = unsafe { ptr::read_unaligned(self.data.as_ptr() as *const T) };
```
sunfishcode, line 527: Can this use a `.cast::<T>()` to avoid an `as` cast?

src/net/send_recv/msg.rs
```diff
@@ -0,0 +1,662 @@
...
+    /// Drop `n` items from this iterator.
+    unsafe fn drop_items(&mut self, n: usize) {
+        /// On drop, move the slice forward by `size_of<T>()`.
+        struct MoveForward<'a, 'b, T>(&'a mut &'b mut [u8], PhantomData<T>);
+
+        impl<T> Drop for MoveForward<'_, '_, T> {
+            fn drop(&mut self) {
+                // Move slice forward.
+                let slice = mem::take(self.0);
+                *self.0 = &mut slice[size_of::<T>()..];
```
sunfishcode, line 484: Can this be written as `*self.0 = &mut self.0[size_of::<T>()..];`?

src/net/send_recv/msg.rs
```diff
@@ -0,0 +1,662 @@
...
+        if self.data.len() < size_of::<T>() {
+            return None;
+        }
+
+        // Get the next item.
+        let item = unsafe { ptr::read_unaligned(self.data.as_ptr() as *const T) };
+
+        // Move forward.
+        let data = mem::take(&mut self.data);
+        self.data = &mut data[size_of::<T>()..];
```
sunfishcode, line 531: As above, can this be simplified to a single line?

src/net/send_recv/msg.rs
```diff
@@ -0,0 +1,662 @@
...
+
+    /// Increment the number of messages we've read.
+    read: &'buf mut usize,
+}
+
+impl<'buf> AncillaryDrain<'buf> {
+    /// A closure that converts a message into a `RecvAncillaryMessage`.
+    #[allow(unsafe_code)]
+    fn cvt_msg(read: &mut usize, msg: NonNull<c::cmsghdr>) -> Option<RecvAncillaryMessage<'buf>> {
+        unsafe {
```
sunfishcode, line 276: Ordinarily, a function that takes a raw pointer like this should be an `unsafe` function. But since this is not public, and it would add a fair amount of clutter to make it unsafe, I think it's ok it is. But it would be good to have more comments about what unsafe things are being done here.

src/net/send_recv/msg.rs
```diff
@@ -0,0 +1,662 @@
...
+            Self {
+                msghdr,
+                header,
+                _buffer: PhantomData,
+            }
+        }
+    }
+
+    impl Iterator for Messages<'_> {
+        type Item = NonNull<c::cmsghdr>;
```
sunfishcode, line 631: Would it be possible make this a `&mut c::cmsghdr` instead of a raw pointer? It seems like that might simplify the uses. If not, could you add a comment mentioning what the lifetime of the pointee is?

sunfishcode (commented):
Thanks for working on this! I've now made a preliminary pass with a few comments.

src/net/send_recv/msg.rs
notgull: That sounds like a better idea! Fixed to use that.

src/net/send_recv/msg.rs
notgull: Not needed after I changed it to `&c::msghdr`.

src/net/send_recv/msg.rs
notgull: Since `&mut [u8]` isn't `Clone`, we need to move it out of the structure before we set it back. It may seem dump, but the borrow checker complains if I do it the normal way.

Cargo.toml
notgull: Oopsy daisy, fixed

Cargo.toml
notgull: `cargo add` just does that for some reason. Fixed!

src/backend/libc/conv.rs
```diff
@@ -222,3 +222,73 @@ pub(super) fn ret_send_recv(len: isize) -> io::Result<usize> {
 pub(super) fn ret_send_recv(len: i32) -> io::Result<usize> {
     ret_usize(len as isize)
 }
+
+/// Convert the value to the `msg_iovlen` field of a `msghdr` struct.
+#[cfg(all(
+    not(any(windows, target_os = "wasm32")),
```
sunfishcode, line 228: Here and below, "wasm32" is a target_arch rather than a target_os.

sunfishcode (commented):
Ok, I've now read through the whole PR. Overall this looks great!

src/net/send_recv/msg.rs
```diff
@@ -0,0 +1,687 @@
...
+        assert_eq!(data.len() % size_of::<T>(), 0);
+
+        Self {
+            data,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Drop `n` items from this iterator.
+    unsafe fn drop_items(&mut self, n: usize) {
```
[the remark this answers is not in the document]
notgull: I removed this function.

src/net/send_recv/msg.rs
```diff
@@ -0,0 +1,687 @@
...
+            // See if there is a `T` left.
+            if self.data.len() < size_of::<T>() {
+                return;
+            }
+
+            // Move forward by one after drop, even on panic.
+            let move_forward = MoveForward::<'_, '_, T>(&mut self.data, PhantomData);
+
+            // Drop the `T`.
+            drop(ptr::read_unaligned(move_forward.0.as_ptr() as *const T));
```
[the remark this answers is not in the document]
notgull: Works by me. Originally it used `drop_in_place` to avoid a read, but eventually I realized I needed to do `read_unaligned` and had to rework it around that. Probably best to just remove it and use `self.for_each(drop)` to drop.

src/backend/linux_raw/net/syscalls.rs
```diff
@@ -239,6 +242,196 @@ pub(crate) fn acceptfrom_with(
     }
 }
 
+#[inline]
```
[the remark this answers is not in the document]
notgull: I've inlined all of the public definitions.

src/net/send_recv/msg.rs
```diff
@@ -0,0 +1,687 @@
...
+    iov: &mut [io::IoSliceMut<'_>],
+    control: &mut RecvAncillaryBuffer<'_>,
+    flags: RecvFlags,
+) -> io::Result<RecvMsgResult> {
+    backend::net::syscalls::recvmsg(socket.as_fd(), iov, control, flags)
+}
+
+/// The result of a `recvmsg` call.
+#[non_exhaustive]
+pub struct RecvMsgResult {
```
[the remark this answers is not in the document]
notgull: It is `non_exhaustive` so that, if we want to add fields in the future, it isn't a breaking change.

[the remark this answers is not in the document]
sunfishcode: Do we anticipate additional fields here? `recvmsg`'s API hasn't fundamentally changed in decades.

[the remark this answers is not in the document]
notgull: Good point, fixed!

sunfishcode (commented):
I apologize for one more round of review here, but I did find a few more things here.

notgull:
> I apologize for one more round of review here, but I did find a few more things here.

No problem! Especially for a crate like `rustix` it's good to dot all our I's and cross all our T's.

src/net/send_recv/msg.rs
```diff
@@ -583,11 +584,11 @@ mod messages {
 
     impl<'buf> Messages<'buf> {
         /// Create a new iterator over messages from a byte buffer.
-        pub(super) fn new(buf: &'buf mut [u8], len: usize) -> Self {
+        pub(super) fn new(buf: &'buf mut [u8]) -> Self {
             let msghdr = {
                 let mut h: c::msghdr = unsafe { core::mem::zeroed() };
                 h.msg_control = buf.as_mut_ptr().cast();
-                h.msg_controllen = len as _;
+                h.msg_controllen = crate::backend::msg_control_len(buf.len());
```
sunfishcode, line 591: I was expecting this would want to fail on overflow, but `msg_control_len` does a saturation. Is that intended?

notgull: Good catch, thanks!

sunfishcode:
Thanks!

sunfishcode:
This is now released in rustix 0.37.16

PR #508, opened 2023-01-11
Update to Linux 6.1 io_uring features.

sunfishcode:
Update to linux-raw-sys 0.2 which adds Linux 6.1 features, and add io_uring bindings for new io_uring features.

### DIFF
--- a/src/io_uring.rs
+++ b/src/io_uring.rs
@@ -178,6 +178,24 @@ pub enum IoringRegisterOp {
 
     /// `IORING_REGISTER_IOWQ_MAX_WORKERS`
     RegisterIowqMaxWorkers = sys::IORING_REGISTER_IOWQ_MAX_WORKERS as _,
+
+    /// `IORING_REGISTER_RING_FDS`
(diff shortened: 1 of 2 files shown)


src/io_uring.rs
```diff
@@ -947,6 +1071,18 @@ impl Default for addr_or_splice_off_in_union {
     }
 }
 
+impl Default for addr3_or_cmd_union {
+    #[inline]
+    fn default() -> Self {
+        let mut s = ::core::mem::MaybeUninit::<Self>::uninit();
+        // Safety: All of Linux's io_uring structs may be zero-initialized.
+        unsafe {
+            ::core::ptr::write_bytes(s.as_mut_ptr(), 0, 1);
```
cgwalters, line 1080: (Hmm, does LLVM really not optimize the case of the "default `default()`" function to this zero memset?)

cgwalters (approved):
(Mostly superficial) lgtm

src/io_uring.rs
sunfishcode: The `Default` derive doesn't work for union types, so we need to provide our own implementation.

Issue #509, opened 2023-01-11
Add setuid style syscalls?


SUPERCILEX:
Any reason the set variants were left out?

sunfishcode:
No reason; I just didn't have a need for them myself when I was adding the get variants.

SUPERCILEX:
Lol sounds good. Need em for my current project, so I'll send a PR.

PR #510, opened 2023-01-12
Add set{u,g}id syscalls

SUPERCILEX:
Closes https://github.com/bytecodealliance/rustix/issues/509


sunfishcode:
It looks like [setuid](https://man7.org/linux/man-pages/man2/setuid.2.html) and [setgid](https://man7.org/linux/man-pages/man2/setgid.2.html) have the same note about id's being per-thread that `setresuid` has.

SUPERCILEX:
Hmmm, so then should we move getuid and friends to `thread` instead of `process`?

sunfishcode:
This whole issue with per-thread ids is news to me. But at first glance, I don't think we want to be giving users an API to set the uid/gid of one thread of a process, leaving the other threads with their old uid/gids. Or if we do, I expect it's `unsafe` and maybe even `doc(hidden)`. Mixed-id processes are something you can't get with libc, and they probably have security implications.

That said, I also don't know of any way to implement setting the id on all threads in rustix. Rustix isn't the libc/libpthread, so it doesn't have a list of all the threads it can iterate over.

SUPERCILEX:
> Or if we do, I expect it's unsafe and maybe even doc(hidden).

unsafe seems reasonable, though I do think it's a little weird to say "not following the posix spec is unsafe." If we do go that route, then it means we can keep `setuid` in `process` and have the invariant be "All threads much change uid in tandem."

sunfishcode:
[This blog post](https://ewontfix.com/17/) has a good analysis. If a process calls setuid to drop privs, security might depend on the entire process dropping privs.

But, there could also be legitimate uses for per-thread ids, as long as the application knows what it's asking for.

I haven't thought deeply about this yet, but maybe what this means is: We name these `set_thread_uid` or something, put them in `rustix::thread`, and give them a warning in their doc comment, to make it super clear what these do. Then I don't think they need to be unsafe or doc(hidden). And in the libc backend, we use `libc:: syscall` to implement these instead of `libc::setuid`.

How does that sound?

SUPERCILEX:
> And in the libc backend, we use libc:: syscall to implement these instead of libc::setuid.

I believe that requires also changing getuid to be a raw syscall in libc (or at least I don't know what semantics we can expect from calling getuid but not setuid in libc).

This boils down to one key question for me: do we want posix semantics or linux semantics. If we do a mix of the two, trying to explain that will be deeply unpleasant (or we can make all the linux-raw-sys *id functions live in thread:: while the use-libc functions live in process::, but now portability is a mess).

The problem I see with taking on linux semantics is that people who choose the `use-libc` backend will expect libc semantics. Because of that, I'm leaning towards requiring posix semantics, but I also see your point about maybe not wanting that sometimes. I also don't like the fact that we'd be making the libc variant unsafe too even though it isn't.

So then maybe the right answer is to put all the linux-raw-sys *id functions in thread:: and tell people to cfg between process:: and thread::? I don't like that either.

Maybe we just throw up our hands, leave setuid in process::, and put a note explaining the subtlety in the docs? I can't think of an easy win here. :(

SUPERCILEX:
Actually, I also just realized `getuid` is currently broken on linux-raw-sys: if some other library uses libc to call `setuid` and you're using `getuid`, there's no guarantee of an atomic transition which means the rustix `getuid` will see uid interleavings that are invalid posix. Not sure what that means for our discussion.

sunfishcode:
I don't think any libc on Linux guarantees an atomic transition here.

What if we do this:
 - Leave `rustix::process::getuid` as-is, but add documentation mentioning that on Linux the ID may be per-thread.
 - Add `rustix::thread::set_thread_uid` etc. on Linux which is just the syscall (in both linux_raw and libc backends).
 - If we ever add POSIX-semanics `setuid` etc., do so as `rustix::process::setuid`, and either don't provide it on Linux, or just use libc to do it, even in the linux_raw backend (similar to how the linux_raw backend uses libc for `getauxval` by default too, because there's no other good way to do it).

Does the "process" in `rustix::process::getuid` make it potentially confusing? Yes, but it's not different from having a C library linked in that calls `getuid` in C and expects it to have POSIX semantics. This is a problem for callers of `rustix::thread::set_thread_uid` to worry about, and Linux designers to worry about, not all extant C libraries that call `getuid` to worry about.

cgwalters:
> Add rustix::thread::set_thread_uid etc. on Linux which is just the syscall (in both linux_raw and libc backends).

It seems a bit odd to not expose `setuid` with `#[cfg(feature = "libc")]` at least though.

But, OTOH the number of codebases that should be calling `setuid()` is really quite small.  There are mature tools like systemd unit files and container runtimes that cover much more than setuid that if one is transitioning from a privileged path to lesser privileged, one also normally wants to do more than just `setuid` anyways - and these other tools should likely be used.

SUPERCILEX:
> I don't think any libc on Linux guarantees an atomic transition here.

Ok, I reread the article and I think you're right. It said, "Atomicity of changes: It needs to be impossible for the application to observe mid-call states where some threads have one set of ids and other threads have a different set of ids." Later: "waiting for all of them either to respond to signals, or exit." Those two statement contradict each other as far as I can tell. It seems like the real implementation is something along the lines of *don't return from my setuid call until all threads have changed their state*.

> What if we do this

I'm on board with that. I think it's as consistent as we're going to get.

---

> It seems a bit odd to not expose setuid with #[cfg(feature = "libc")] at least though.

Do you mean in `::process`? I think that's reasonable, except that now I'm not sure `setuid` will even work properly in Rust with regards to threading. The [setxid implementation](https://github.com/bminor/glibc/blob/7e31d166510ac4adbf53d5e8144c709a37dd8c7a/nptl/nptl_setxid.c) relies on pthreads, but I thought rust didn't use pthreads which would mean setxid won't find the threads to signal. Or maybe I'm wrong? Until someone digs into the detail, I think we should stick with just `::thread`.

SUPERCILEX:
Ok, non-linux libc doesn't let you make raw syscalls as far as I can tell? So I ended up only doing the syscall approach for linux (which should be fine since other unixes are supposed to comply with posix).

SUPERCILEX:
Also the freebsd builds keep failing with `curl: (60) SSL certificate problem: unable to get local issuer certificate`, not sure what to do about that.

src/thread/id.rs
```diff
@@ -12,6 +11,61 @@ use crate::process::Pid;
...
+/// # References
+///  - [POSIX]
+///  - [Linux]
+///
+/// [POSIX]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/setuid.html
+/// [Linux]: https://man7.org/linux/man-pages/man2/setuid.2.html
+#[inline]
+#[must_use]
+#[cfg(not(any(target_os = "redox", target_os = "wasi")))]
+pub fn setuid(uid: crate::process::Uid) -> io::Result<()> {
```
sunfishcode, line 35: I propose renaming this function to `set_thread_uid`, having it only be defined on `cfg(any(target_os = "android", target_os = "linux"))`, and rephrasing the warning from "On Linux, [it's different]" to "This is not `setuid`; here's now it's different".

sunfishcode: And similar for `setgid`.

cgwalters:
This debate about per-thread state and safety mirrors that of [setenv](https://internals.rust-lang.org/t/synchronized-ffi-access-to-posix-environment-variable-functions/15475) - if there's only one thread, then the distinction doesn't matter.  (And really everything invoking `setuid`, like `setenv` should be doing it extremely early in process startup before there are any other threads)

src/thread/id.rs
SUPERCILEX: So we leave other platforms out?

sunfishcode: If any other platform has a way to set a thread-specific uid, then we can implement it for them. I don't know of any others that do.

Does your use case need POSIX-semantics `setuid`? If so, then we add a `rustix::process::setuid` which always calls libc. (Or, we implement the POSIX semantics manually for linux_raw, but I'm assuming that's not worthwhile at this point.)

SUPERCILEX: Gotya, so we're going with the "linux is weird and gets special treatment, we'll deal with other platforms later" approach. Seems reasonable to me. And yeah, I only need linux support.

sunfishcode: The ideal is that any given name has at most one meaning. Rustix has tons of functions that are available on some OS's and not others. But we want to avoid the same name meaning different things on different platforms.

That said, there are gray areas. There are platform-specific variations in lots of mostly-portable functions. So we have to make a judgement call whether something is "different enough" to justify giving it a different name. In this case, setting the uid for just one thread instead of all threads feels definitely "different enough".

SUPERCILEX: Gotya, thanks for the explanation. :+1:

src/thread/id.rs
```diff
@@ -15,3 +15,45 @@ use crate::process::Pid;
 pub fn gettid() -> Pid {
     backend::thread::syscalls::gettid()
 }
+
+/// `setuid(uid)`—Sets the effective user ID of the calling process.
```
sunfishcode, line 19: Since this isn't the same as POSIX `setuid`, I think it makes sense to omit the code-quoted part here.

SUPERCILEX:
Awesome, done!

Issue #492, opened 2022-12-28
Way to retrieve current process's capabilities?


SUPERCILEX:
You can check the ambient and bounding set, but not the effective set? Granted, I think this requires using https://man7.org/linux/man-pages/man2/capget.2.html, but it'd be nice if this was an option.

PR #514, opened 2023-01-13
Add cap{get,set} syscalls

SUPERCILEX:
Closes https://github.com/bytecodealliance/rustix/issues/492

### DIFF
--- a/src/backend/libc/thread/syscalls.rs
+++ b/src/backend/libc/thread/syscalls.rs
@@ -3,7 +3,7 @@
 use super::super::c;
 use super::super::conv::ret;
 #[cfg(any(target_os = "android", target_os = "linux"))]
-use super::super::conv::{borrowed_fd, ret_c_int};
+use super::super::conv::{borrowed_fd, ret_c_int, syscall_ret};
 use super::super::time::types::LibcTimespec;
 #[cfg(any(target_os = "android", target_os = "linux"))]
 use crate::fd::BorrowedFd;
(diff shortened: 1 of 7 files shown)


src/backend/libc/thread/syscalls.rs
```diff
@@ -299,3 +299,21 @@ pub(crate) fn setns(fd: BorrowedFd, nstype: c::c_int) -> io::Result<c::c_int> {
 pub(crate) fn unshare(flags: crate::thread::UnshareFlags) -> io::Result<()> {
     unsafe { ret(c::unshare(flags.bits() as i32)) }
 }
+
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[inline]
+pub(crate) fn capget(
+    header: *mut linux_raw_sys::general::__user_cap_header_struct,
+    data: *const linux_raw_sys::general::__user_cap_data_struct,
```
sunfishcode, line 307: This function and several like it take raw pointers, so they should be `unsafe fn`.

Also, should `data` here be a `*mut` pointer?

src/thread/libcap.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+/// `capget(_LINUX_CAPABILITY_VERSION_3, pid)`
+///
+/// # References
+///  - [Linux]
+///
+/// [Linux]: https://man7.org/linux/man-pages/man2/capget.2.html
+#[inline]
+pub fn get_capabilities(pid: Option<Pid>) -> io::Result<CapabilitySets> {
```
sunfishcode, line 113: Could you add a `#[doc(alias = "capget"))]` for this?

src/thread/libcap.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+/// `capset(_LINUX_CAPABILITY_VERSION_3, pid, effective, permitted, inheritable)`
+///
+/// # References
+///  - [Linux]
+///
+/// [Linux]: https://man7.org/linux/man-pages/man2/capget.2.html
+#[inline]
+pub fn set_capabilities(pid: Option<Pid>, sets: CapabilitySets) -> io::Result<()> {
```
sunfishcode, line 124: And similarly, doc alias "capset" for this?

src/thread/libcap.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    let data = {
+        let mut header = linux_raw_sys::general::__user_cap_header_struct {
+            version: linux_raw_sys::general::_LINUX_CAPABILITY_VERSION_3,
+            pid: Pid::as_raw(pid) as backend::c::c_int,
+        };
+
+        backend::thread::syscalls::capget(
+            &mut header as *mut linux_raw_sys::general::__user_cap_header_struct,
+            &mut data as *mut _ as *mut linux_raw_sys::general::__user_cap_data_struct,
```
sunfishcode, line 141: Instead of doing a pointer cast, can this use [`as_mut_ptr()`](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.as_mut_ptr)?

src/thread/libcap.rs
```diff
@@ -0,0 +1,189 @@
...
+    let mut data = MaybeUninit::<[linux_raw_sys::general::__user_cap_data_struct; 2]>::uninit();
+
+    let data = {
+        let mut header = linux_raw_sys::general::__user_cap_header_struct {
+            version: linux_raw_sys::general::_LINUX_CAPABILITY_VERSION_3,
+            pid: Pid::as_raw(pid) as backend::c::c_int,
+        };
+
+        backend::thread::syscalls::capget(
+            &mut header as *mut linux_raw_sys::general::__user_cap_header_struct,
```
sunfishcode, line 140: I believe there's an implicit convertion from `&T` to `*const T`, so the `as` casts in these can be omitted.

src/backend/libc/thread/syscalls.rs
SUPERCILEX: > Also, should data here be a *mut pointer?

Really good catch, thank you!

Regarding unsafe, I realized this abstraction sucks anyway: these functions take in references now and are in charge of raw pointer shenanigans.

sunfishcode:
Thanks!

PR #515, opened 2023-01-15
Simplify RawDir iteration

SUPERCILEX:
Recently had to look at this again and unsurprisingly I have no idea what past me was thinking. :sweat_smile: 

Basic idea: get rid of the loop. It ain't needed.

New control flow:
- If we've read up to the initialized bytes (including `0 >= 0`), we need to refill the buffer.
- `initialized` and `offset` only ever need to be reset when we actually succeeded in filling the buffer. On an error, we return which means that if `next` gets called again, the `offset >= initialized` check will run again with the same result since nothing changed.

Note: the diff is overcomplicating things, the main change is just swapping where the syscall and buffer consumption occur.


sunfishcode:
Makes sense!

PR #520, opened 2023-01-17
0.36 cherrypicks

sunfishcode:
This cherry-picks several patches from main onto the 0.36 branch. main has several semver-incompatible changes that will need a 0.37 branch, so 0.36 will just be for the changes that are semver compatible.


sunfishcode:
Oops, this is based on the wrong branch.

PR #523, opened 2023-01-19
Add NO_AUTOMOUNT AtFlags,  STATX_DIOALIGN mask and reorder bitflag to follow numbering

ETKNeil:
The statx syscall accept the [AT_NO_AUTOMOUNT flag ](https://man.archlinux.org/man/statx.2#AT_NO_AUTOMOUNT), however this one was not listed inside the AtFlags bitflag. 
This PR add such flag from the underlying raw types, it also re-orders the bitflag to be in correlation with the underlying values.

The statx syscall accept the [STATX_DIOALIGN](https://elixir.bootlin.com/linux/v6.1/A/ident/STATX_DIOALIGN) mask since Linux 6.1 however this one was not listed inside the StatxFlags bitflag.
This PR add such flag from the underlying raw types.

Note, `can_compile` signature in build.rs was changed on the 13th of January and you can remove the reference to format (thx clippy)

### DIFF
--- a/src/backend/libc/fs/types.rs
+++ b/src/backend/libc/fs/types.rs
@@ -648,6 +648,9 @@ bitflags! {
         /// `STATX_MNT_ID` (since Linux 5.8)
         const MNT_ID = c::STATX_MNT_ID;
 
+        /// `STATX_DIOALIGN` (since Linux 6.1)
+        const DIOALIGN = c::STATX_DIOALIGN;
+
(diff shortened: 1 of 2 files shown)


sunfishcode:
This looks good to me, so now it's just waiting on the libc PR.

sunfishcode:
The libc PR here landed and is released as of libc 0.2.140, so now this just needs a rebase.

PR #524, opened 2023-01-22
Add method to know when `RawDir` buffer will be refilled

SUPERCILEX:
This is useful for knowing how long the file name pointers will be valid for. In my case, I'm using it to know when I need to perform an io_uring flush (since the io_uring ops use the file name pointers).


sunfishcode:
Sounds a little tricky, but I suppose this is `RawDir` after all :-).

PR #527, opened 2023-01-27
Add O_DIRECT flag

jordanisaacs:
Added the `O_DIRECT` flag. Not sure if there was a reason for not having it included.

### DIFF
--- a/src/backend/libc/fs/types.rs
+++ b/src/backend/libc/fs/types.rs
@@ -259,6 +259,17 @@ bitflags! {
             target_os = "linux",
         ))]
         const NOATIME = c::O_NOATIME;
+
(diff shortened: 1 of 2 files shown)


sunfishcode:
Thanks! Could you also add `O_DIRECT` to the libc backend, in src/backend/libc/fs/types.rs? It looks like it's available on at least Linux, Android, Emscripten, FreeBSD, NetBSD, and Fuchsia.

Also, the x86_64-unknown-freebsd-13 CI failure here is unrelated so we can ignore that here.

jordanisaacs:
Added in the libc backend @sunfishcode

sunfishcode:
Thanks!

PR #528, opened 2023-02-02
Add support for loongarch64

zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html

### DIFF
--- a/src/backend/linux_raw/fs/types.rs
+++ b/src/backend/linux_raw/fs/types.rs
@@ -573,7 +573,10 @@ pub struct Stat {
 ///
 /// [`statat`]: crate::fs::statat
 /// [`fstat`]: crate::fs::fstat
-#[cfg(all(target_pointer_width = "64", not(target_arch = "mips64")))]
+#[cfg(all(
+    target_pointer_width = "64",
+    not(any(target_arch = "mips64", target_arch = "loongarch64"))
+))]
 pub type Stat = linux_raw_sys::general::stat;
 
 /// `struct statfs` for use with [`statfs`] and [`fstatfs`].
--- a/src/backend/linux_raw/process/syscalls.rs
+++ b/src/backend/linux_raw/process/syscalls.rs
@@ -115,7 +115,11 @@ pub(crate) fn getpgid(pid: Option<Pid>) -> io::Result<Pid> {
 #[inline]
 pub(crate) fn getpgrp() -> Pid {
     // Use the `getpgrp` syscall if available.
-    #[cfg(not(any(target_arch = "aarch64", target_arch = "riscv64")))]
+    #[cfg(not(any(
+        target_arch = "aarch64",
+        target_arch = "loongarch64",
+        target_arch = "riscv64"
+    )))]
     unsafe {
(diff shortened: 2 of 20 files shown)


src/backend/linux_raw/vdso_wrappers.rs
```diff
@@ -361,6 +361,8 @@ fn init() {
         let ptr = vdso.sym(cstr!("LINUX_2.6"), cstr!("__vdso_clock_gettime64"));
         #[cfg(target_arch = "mips64")]
         let ptr = vdso.sym(cstr!("LINUX_2.6"), cstr!("__vdso_clock_gettime"));
+        #[cfg(target_arch = "loongarch64")]
+        let ptr = vdso.sym(cstr!("LINUX_5.19"), cstr!("__vdso_clock_gettime"));
```
xen0n, line 365: This is wrong, LoongArch vDSO symbol version is `LINUX_5.10` [according to the sources](https://github.com/torvalds/linux/blob/master/arch/loongarch/vdso/vdso.lds.S). (It's the version at the time LoongArch upstream work has started, not the time when the port was mainlined, IIUC.)

src/backend/linux_raw/process/syscalls.rs
```diff
@@ -328,6 +336,7 @@ pub(crate) fn setpriority_process(pid: Option<Pid>, priority: i32) -> io::Result
     }
 }
 
+#[cfg(not(target_arch = "loongarch64"))]
```
xen0n, line 339: I don't think we should deprive LoongArch from having old-style `{get,set}rlimit` API support. As you can see, this function is actually just a `prlimit64` wrapper.

src/backend/linux_raw/fs/types.rs
```diff
@@ -573,7 +573,10 @@ pub struct Stat {
 ///
 /// [`statat`]: crate::fs::statat
 /// [`fstat`]: crate::fs::fstat
-#[cfg(all(target_pointer_width = "64", not(target_arch = "mips64")))]
+#[cfg(all(
+    target_pointer_width = "64",
+    not(any(target_arch = "mips64", target_arch = "loongarch64"))
```
xen0n, line 578: Similarly we most definitely don't want to force everyone to migrate to `statx` if they ever want their code to run on LoongArch systems. We should keep the generic `stat` definitions here, but add code for translating between the two ABIs: see [my Go adaptation a while ago](https://go.dev/cl/407694) (check `src/syscall/syscall_linux_loong64.go`) or glibc (check `sysdeps/unix/sysv/linux/fstatat64.c`). We have to add the code because we're the first 64-bit architecture without `fstat` at all.

xen0n (changes requested):
Now I see why https://github.com/rust-lang/rust/pull/101069 is needed at all at such an early bring-up stage... this crate is being depended upon by Cargo. A lot have changed since I last ported Rust.

cc @heiher

PR #531, opened 2023-02-03
Backport patches to 0.36

sunfishcode:


sunfishcode:
I dropped #514 here because that depends on a new linux-raw-sys and rustix does currently have linux-raw-sys as a public dependency.

PR #532, opened 2023-02-03
Add support for pidfd_open and waitid

notgull:
This PR adds support for `pidfd_open` on Linux. I wasn't sure if it should go in `process` or `io`, so I just put it under `process`.

This PR also adds `waitid` for testing `pidfd_open`, although the implementation could probably use some expansion.


notgull:
Looks like the MacOS VM just timed out :-/

sunfishcode:
`process` is a good place for this. I think this looks good.

PR #536, opened 2023-02-13
Bump errno 0.3.0 and memoffset 0.8.0

Unpublished:
Gets rid of winapi as implicit dep (from [errno](https://github.com/lambda-fairy/rust-errno/blob/main/CHANGELOG.md#030---2023-02-12)).


sunfishcode:
Thanks! This also has https://github.com/lambda-fairy/rust-errno/pull/66 which will help us support WASI on Rust stable.

PR #540, opened 2023-02-18
Reenable FreeBSD CI with correct image family

valpackett:
The issue was using snapshots of already released versions, which is not supported.


sunfishcode:
Thanks!

PR #541, opened 2023-02-20
FreeBSD RESOLVE_BENEATH support and more

valpackett:
This is everything required to implement support for FreeBSD's `(O|AT)_RESOLVE_BENEATH` based lookup sandboxing in cap-std :) While here, also expose more Linux-compatible stuff that FreeBSD's libc exposes, also improve kqueue a bit.

The `chmodat` signature change is an API break :/ but the omission of `flags` feels like a mistake to me and I think it's better to fix it earlier rather than later. What do you think?

~~Waiting for: a libc release with https://github.com/rust-lang/libc/pull/3114 https://github.com/rust-lang/libc/pull/3122 https://github.com/rust-lang/libc/pull/3124~~ wheeee  
~~Includes: #540~~

### DIFF
--- a/src/backend/libc/fs/syscalls.rs
+++ b/src/backend/libc/fs/syscalls.rs
@@ -638,18 +638,41 @@ unsafe fn utimensat_old(
     target_os = "redox",
     target_os = "wasi",
 )))]
-pub(crate) fn chmodat(dirfd: BorrowedFd<'_>, path: &CStr, mode: Mode) -> io::Result<()> {
-    unsafe { ret(c::fchmodat(borrowed_fd(dirfd), c_str(path), mode.bits(), 0)) }
+pub(crate) fn chmodat(
+    dirfd: BorrowedFd<'_>,
+    path: &CStr,
+    mode: Mode,
+    flags: AtFlags,
+) -> io::Result<()> {
+    unsafe {
+        ret(c::fchmodat(
+            borrowed_fd(dirfd),
+            c_str(path),
+            mode.bits() as c::mode_t,
+            flags.bits(),
+        ))
+    }
 }
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
-pub(crate) fn chmodat(dirfd: BorrowedFd<'_>, path: &CStr, mode: Mode) -> io::Result<()> {
+pub(crate) fn chmodat(
+    dirfd: BorrowedFd<'_>,
+    path: &CStr,
+    mode: Mode,
+    flags: AtFlags,
+) -> io::Result<()> {
     // Linux's `fchmodat` does not have a flags argument.
     //
     // Use `c::syscall` rather than `c::fchmodat` because some libc
     // implementations, such as musl, add extra logic to `fchmod` to emulate
     // support for `O_PATH`, which uses `/proc` outside our control and
     // interferes with our own use of `O_PATH`.
+    if flags == AtFlags::SYMLINK_NOFOLLOW {
+        return Err(io::Errno::OPNOTSUPP);
+    }
+    if !flags.is_empty() {
+        return Err(io::Errno::INVAL);
+    }
     unsafe {
         // Pass `mode` as a `c_uint` even if `mode_t` is narrower, since
         // `libc_openat` is declared as a variadic function and narrower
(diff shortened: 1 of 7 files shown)


sunfishcode (commented):
Cc @notgull in case you're interested in the kqueue changes.

For chmodat, it feels like it might be error-prone to have the flag on Linux where it isn't supported. What would you think about adding a chmodat_with function with a flags argument, that's only available on platforms that support flags?

valpackett:
Hm, that can be done for not breaking backwards compatibility I guess.
But `AtFlags` is already not something that guarantees that every flag will be used. A lot of the AtFlags are function specific, for example `AT_REMOVEDIR` is for `unlinkat` and would do nothing for, say, `fstatat`. I guess the difference here is the kernel would return `EINVAL` for unsupported flags, but the ignoring implementation won't—but we can just make it do that.

sunfishcode:
Good point, having it fail with `EINVAL` sounds good. Or maybe `ENOTSUP` would be better?

The other concern here is that rustix recently had a semver bump with 0.37, and we hopefully won't need another for quite a while. Would you be ok introducing a different name for now, with a plan to change it once we start thinking about the next semver bump?

valpackett:
Nah, `EINVAL` is the correct one for unsupported arguments, that's what all the libcs return in that case.

Fun discovery btw: [musl implements `AT_SYMLINK_NOFOLLOW` there](https://git.musl-libc.org/cgit/musl/commit/?id=0dc4824479e357a3e23a) even though glibc doesn't.

Changed to `chmodat_with`, hopefully the doc stuff looks okay…

src/fs/at.rs
```diff
@@ -319,7 +315,31 @@ pub fn utimensat<P: path::Arg, Fd: AsFd>(
...
 pub fn chmodat<P: path::Arg, Fd: AsFd>(dirfd: Fd, path: P, mode: Mode) -> io::Result<()> {
-    path.into_with_c_str(|path| backend::fs::syscalls::chmodat(dirfd.as_fd(), path, mode))
+    chmodat_with(dirfd, path, mode, AtFlags::empty())
+}
+
+/// `fchmodat(dirfd, path, mode, flags)`—Sets file or directory permissions.
+///
+/// Platform support for flags varies widely, for example Linux does not yet
+/// implement `AT_SYMLINK_NOFOLLOW` (except when using the libc backend with musl libc)
+/// and will return `EINVAL` with non-empty flags.
```
sunfishcode, line 325: Glibc and musl use `EINVAL` for any flag other than `AT_SYMLINK_NOFOLLOW`, but they use `EOPNOTSUP` for `AT_SYMLINK_FOLLOW` if they can't emulate it, so it looks like `EOPNOTSUP` is the way to go here.

src/backend/libc/fs/syscalls.rs
```diff
@@ -638,12 +638,29 @@ unsafe fn utimensat_old(
...
+    }
 }
 
 #[cfg(any(target_os = "android", target_os = "linux"))]
-pub(crate) fn chmodat(dirfd: BorrowedFd<'_>, path: &CStr, mode: Mode) -> io::Result<()> {
+pub(crate) fn chmodat(
+    dirfd: BorrowedFd<'_>,
+    path: &CStr,
+    mode: Mode,
+    _flags: AtFlags,
```
sunfishcode, line 662: This version of `chmodat` should also check the flags and issue an error.

valpackett: Here the libc itself will return the error. It would be wrong to add our own checks, because at any point a future release of glibc should be free to implement the support!

valpackett: Ah, oops… Why did this `linux`/`android` version exist? `libc::fchmodat` is defined right in `src/unix/mod.rs` in libc. Removed to let the actual libc check the flags and issue the error (or emulate the behavior).

src/fs/at.rs
valpackett: Ah, interesting, [true](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/sysdeps/unix/sysv/linux/fchmodat.c#L36). Changed.

src/backend/libc/fs/syscalls.rs
valpackett: Interesting, avoiding `/proc` does sound like a good idea!

sunfishcode:
This looks good! I made a note about the naming of `chmodat_with` in #578.

valpackett:
Yaay :3 I'll have a little bit more coming today (adding more procctl's right now), hopefully it'd be possible to get that into the next minor release

PR #542, opened 2023-02-22
Support termios2

SUPERCILEX:
Needed to be able to set the baud rate to arbitrary values.

### DIFF
--- a/src/backend/libc/termios/syscalls.rs
+++ b/src/backend/libc/termios/syscalls.rs
@@ -27,6 +27,27 @@ pub(crate) fn tcgetattr(fd: BorrowedFd<'_>) -> io::Result<Termios> {
     }
 }
 
+#[cfg(all(
+    any(target_os = "android", target_os = "linux"),
+    any(
+        target_arch = "x86",
+        target_arch = "x86_64",
+        target_arch = "x32",
+        target_arch = "riscv64",
+        target_arch = "aarch64",
+        target_arch = "arm",
+        target_arch = "mips",
+        target_arch = "mips64",
+    )
+))]
+pub(crate) fn tcgetattr2(fd: BorrowedFd<'_>) -> io::Result<crate::termios::Termios2> {
+    let mut result = MaybeUninit::<crate::termios::Termios2>::uninit();
+    unsafe {
+        ret(c::ioctl(borrowed_fd(fd), c::TCGETS2, result.as_mut_ptr()))?;
+        Ok(result.assume_init())
+    }
+}
+
 #[cfg(not(target_os = "wasi"))]
 pub(crate) fn tcgetpgrp(fd: BorrowedFd<'_>) -> io::Result<Pid> {
     unsafe {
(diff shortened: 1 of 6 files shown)


src/backend/linux_raw/termios/types.rs
```diff
@@ -52,11 +52,18 @@ pub enum Action {
     IOn = linux_raw_sys::general::TCION,
 }
 
-/// `struct termios` for use with [`tcgetattr`].
+/// `struct termios` for use with [`tcgetattr`] and [`tcsetattr`].
 ///
 /// [`tcgetattr`]: crate::termios::tcgetattr
+/// [`tcsetattr`]: crate::termios::tcsetattr
 pub type Termios = linux_raw_sys::general::termios;
```
notgull, line 59: I think we should avoid making `libc` and `linux-raw-sys` public dependencies.

SUPERCILEX: Fair enough, but that's neither here nor there: termios is already implemented like that.

notgull: Mhmm, fair enough, I didn't see that.

src/backend/libc/termios/syscalls.rs
```diff
@@ -27,6 +27,27 @@ pub(crate) fn tcgetattr(fd: BorrowedFd<'_>) -> io::Result<Termios> {
...
+    any(target_os = "android", target_os = "linux"),
+    any(
+        target_arch = "x86",
+        target_arch = "x86_64",
+        target_arch = "x32",
+        target_arch = "riscv64",
+        target_arch = " aarch64",
+        target_arch = " arm",
+        target_arch = " mips",
+        target_arch = " mips64",
```
sunfishcode, line 40: Here and in the rest of the PR, are those spaces in the strings intended?

sunfishcode (commented):
It seems TCGETS2 etc. were added in Linux 2.6.20, which is older than the oldest supported Linux version in Rust. As such, would it make sense to just switch `tcgetattr` on Linux to just do `tcgetattr2` and make the public `Termios` type just be `struct termios2`?

There are some merge conflicts now, but they're hopefully just removing boilerplate that's no longer needed after #543.

Also, I agree it'd be cleaner if Rustix didn't directly export libc/linux-raw-sys types in its public API, but it already does this in several areas, so I don't think we need to hold up this PR for that.

src/backend/libc/termios/syscalls.rs
SUPERCILEX: Hot damn, #543 is great!

> make the public Termios type just be struct termios2

Would that be a breaking change? And what about the CFGs? It seems like termios2 isn't a thing outside those architectures, but OG termios is.

sunfishcode: I meant to say make `Termios` be `termios2` on Linux etc., and `termios` otherwise. It looks like `termios2` is a superset of `termios`. I guess one potentially annoying thing about that approach is if users are deconstructing the struct, it obliges them to name the other fields, but maybe they can just do `..`?

SUPERCILEX: Gotya, that seems reasonable :+1:

SUPERCILEX: Hmmm, what do we do about libc and its methods like `cfgetospeed`. Do we re-implement them? That doesn't seem portable so maybe libc just doesn't get termios2? Or maybe we just keep 'em separate?

sunfishcode: Oh, good point. Yeah, I'd like to avoid linux-raw gaining APIs that we can't implement on libc. And on libc we should stick with libc rather than reimplementing these. So I guess we should keep 'em separate.

SUPERCILEX: Done.

PR #543, opened 2023-02-22
Make termios use single source from backend

ecnelises:
Previous implementation produces duplicated definitions and cfg attributes. Use constants defined in termios backend now.


sunfishcode:
Thanks, this looks much tidier!

Issue #526, opened 2023-01-25
Intermittent "feature cannot be used on stable release channel"


jyn514:
I temporarily got this error trying to build rustix as a dependency:
```
error[E0554]: `#![feature]` may not be used on the stable release channel
--> /Users/jyn/.local/lib/cargo/registry/src/github.com-1ecc6299db9ec823/rustix-0.36.4/src/lib.rs:99:26
   |
99 | #![cfg_attr(rustc_attrs, feature(rustc_attrs))]
   |                          ^^^^^^^^^^^^^^^^^^^^
```

I have unfortunately not been able to reproduce this error; I updated the version of rustix, which I originally thought fixed the issue, but when I downgraded to 0.36.4 I couldn't reproduce it again.

I think the build script needs to have `rerun-if-env-changed=RUSTC` or something like that? If it helps, my default toolchain is `nightly` but I have a `rust-toolchain` file that says to use 1.65.

jyn514:
cc https://github.com/rust-lang/cargo/issues/10367

intgr:
FWIW running `rm -rf target` to delete the cache fixed this issue for me.

p-kraszewski:
I have this very error popping in my various projects (95% of the time related to `rustix`). A simple `cargo clean` is a fix, but it makes me fully recompile big projects, which is slightly annoying.

Sometimes just adding a new dependency to my projects triggers it.

sunfishcode:
I submitted https://github.com/bytecodealliance/rustix/pull/544 in hopes that it may help. However, I don't yet have a reliable way to reproduce the problem locally, so I can't say for sure whether it does.

jyn514:
> I have this very error popping in my various projects (95% of the time related to `rustix`). A simple `cargo clean` is a fix, but it forces me to fully recompile big projects, which is slightly annoying.

@p-kraszewski as a workaround, you can use `cargo clean -p rustix` to only recompile rustix instead of all your dependencies

sunfishcode:
#544, which may fix this, is now released in version 0.37.

sunfishcode:
Does anyone who encountered this issue continue to see it with 0.37, or 0.36.11?

p-kraszewski:
Not encountered within at least the last 2 weeks. Thank you.

mschneider:
just encountered it with rustix 0.37.7

sunfishcode:
[...]

PR #544, opened 2023-02-24
Add more `cargo:rerun-if-env-changed`'s to build.rs

sunfishcode:
Add more `cargo:rerun-if-env-changed`'s to build.rs for rustix's features, configuration flags, and the toolchain variables rustix uses for autoconfiguration, to hopefully ensure it gets rerun when anything changes.

See also #526.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -107,6 +107,44 @@ fn main() {
     }
 
     println!("cargo:rerun-if-env-changed=CARGO_CFG_RUSTIX_USE_EXPERIMENTAL_ASM");
+    println!("cargo:rerun-if-env-changed=CARGO_CFG_RUSTIX_USE_LIBC");
+
+    // Rerun this script if any of our features or configuration flags change,
+    // or if the toolchain we used for feature detection changes.
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_USE_LIBC");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_USE_LIBC_AUXV");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_LIBC");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_RUSTC_DEP_OF_STD");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_DEFAULT");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_STD");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_CC");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_ITOA");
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -107,6 +107,44 @@ fn main() {
...
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_MM");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_RAND");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_RUNTIME");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_ALL_APIS");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_OS_PIPE");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_FS_ERR");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_ALL_IMPLS");
+    println!("cargo:rerun-if-env-changed=CARGO_CFG_MIRI");
+    println!("cargo:rerun-if-env-changed=CARGO_ENCODED_RUSTFLAGS");
+    println!("cargo:rerun-if-env-changed=RUSTC");
```
jyn514, line 144: I would *expect* just this to fix the issue, but having more `rerun` directives won't hurt anything.

jyn514 (approved):

build.rs
sunfishcode: That might be true. I've now scaled back the list a little to exclude all the cargo features, because those seem less likely to be the problem here, and are more tedious to maintain. Hopefully this works!

build.rs
```diff
@@ -107,6 +107,18 @@ fn main() {
...
+
+    // Rerun this script if any of our features or configuration flags change,
+    // or if the toolchain we used for feature detection changes.
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_USE_LIBC");
+    println!("cargo:rerun-if-env-changed=CARGO_FEATURE_RUSTC_DEP_OF_STD");
+    println!("cargo:rerun-if-env-changed=CARGO_CFG_MIRI");
+    println!("cargo:rerun-if-env-changed=CARGO_ENCODED_RUSTFLAGS");
+    println!("cargo:rerun-if-env-changed=RUSTC");
+    println!("cargo:rerun-if-env-changed=TARGET");
+    println!("cargo:rerun-if-env-changed=CARGO_RUSTC_WRAPPER");
```
vlad20012, line 120: It seems like `CARGO_RUSTC_WRAPPER` is a wrong name. Could you please change it to the proper `RUSTC_WRAPPER` ?

For reference: https://github.com/dtolnay/anyhow/issues/248

sunfishcode: Good catch! I've now submitted https://github.com/bytecodealliance/rustix/pull/563 to fix that.

PR #545, opened 2023-02-24
Emulate flock() on Solaris via fcntl().

psumbera:
This should resolve problem with Rust build:
https://github.com/rust-lang/rust/issues/103630

### DIFF
--- a/src/backend/libc/fs/syscalls.rs
+++ b/src/backend/libc/fs/syscalls.rs
@@ -86,7 +86,7 @@ use crate::fs::Advice;
     target_os = "solaris",
 )))]
 use crate::fs::FallocateFlags;
-#[cfg(not(any(target_os = "solaris", target_os = "wasi")))]
+#[cfg(not(target_os = "wasi"))]
 use crate::fs::FlockOperation;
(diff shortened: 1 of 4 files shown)


sunfishcode:
Unfortunately, `fcntl`-style locks have different semantics than `flock`-style locks. `fcntl`-style locks have the old process-associated behavior, which means that they don't work if multiple threads use a lock to synchronize access between themselves. Rustix aims to be transparent with respect to significant system call semantics, so it shouldn't silently provide `fcntl` locking semantics to users expecting `flock` semantics.

Is the fd-lock crate the best place to implement this emulation? It might be. I wonder if it would make sense on Solaris to have fd-lock keep a static set of which files it has locked, so that a thread could check whether any other thread has a particular file locked before trying to acquire an `fcntl`-style lock. In any case, I think the [fd-lock repo](https://github.com/yoshuawuyts/fd-lock) is the best place to discuss it.

sunfishcode:
Unfortunately I think this isn't the right approach for rustix; please file an issue in the fd-lock repo to discuss options there.

sunfishcode:
I submitted a PR adding an `fcntl_lock` function to rustix in https://github.com/bytecodealliance/rustix/pull/555, which an fd-lock PR could be built on.

PR #546, opened 2023-02-28
Some io_uring additions

jordanisaacs:
A couple io_uring additions and an adjustment to `io_uring_register`.

`io_uring_register` can return positive values:

> On success, [io_uring_register(2)](https://man7.org/linux/man-pages//man2/io_uring_register.2.html) returns either 0 or a positive
> value, depending on the opcode used.  On error, a negative error
> value is returned. The caller should not rely on the [errno](https://man7.org/linux/man-pages//man3/errno.3.html)
> variable.

`sqe.ioprio` takes in flags for certain operations so added a union for those ([accept](https://github.com/torvalds/linux/blob/master/include/uapi/linux/io_uring.h#L293) and [send/recv](https://github.com/torvalds/linux/blob/ae3419fbac845b4d3f3a9fae4cc80c68d82cdf6e/include/uapi/linux/io_uring.h#L330))

Adds `io_uring_recvmsg_out` which outs recvmsg flags (see [here](https://man7.org/linux/man-pages/man3/io_uring_recvmsg_out.3.html))

> __u32 flags;      /* Flags result as would have been populated by recvmsg(2) */

### DIFF
--- a/src/io_uring.rs
+++ b/src/io_uring.rs
@@ -56,7 +56,7 @@ pub unsafe fn io_uring_register<Fd: AsFd>(
     opcode: IoringRegisterOp,
     arg: *const c_void,
     nr_args: u32,
-) -> io::Result<()> {
+) -> io::Result<u32> {
     backend::io_uring::syscalls::io_uring_register(fd.as_fd(), opcode, arg, nr_args)
 }
 
@@ -467,6 +467,12 @@ bitflags::bitflags! {
 
         /// `IORING_CQE_F_MORE`
         const MORE = sys::IORING_CQE_F_MORE as _;
+
+        /// `IORING_CQE_F_SOCK_NONEMPTY`
+        const SOCK_NONEMPTY = sys::IORING_CQE_F_SOCK_NONEMPTY as _;
+
+        /// `IORING_CQE_F_NOTIF`
+        const NOTIF = sys::IORING_CQE_F_NOTIF as _;
     }
 }
 
@@ -558,6 +564,9 @@ bitflags::bitflags! {
 
         /// `IORING_FEAT_SUBMIT_STABLE`
         const SUBMIT_STABLE = sys::IORING_FEAT_SUBMIT_STABLE;
+
+        /// `IORING_FEAT_LINKED_FILE`
+        const LINKED_FILE = sys::IORING_FEAT_LINKED_FILE;
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/io_uring.rs
```diff
@@ -1047,6 +1128,18 @@ pub struct io_uring_buf {
     pub resv: u16,
 }
 
+impl Default for ioprio_or_flags_union {
+    #[inline]
+    fn default() -> Self {
+        let mut s = ::core::mem::MaybeUninit::<Self>::uninit();
```
SUPERCILEX, line 1134: `::zeroed()` would be much clearer instead of the ptr write.

sunfishcode: I believe the original idea for `write_bytes` in other places that do this is that `zeroed()` doesn't guarantee to write to any padding bytes, while `write_bytes` does. However, looking at it now, this doesn't seem like it should matter. This `default()` function still relies on Rust to move the value into the return value, which possibly doesn't guarantee to preserve padding bytes, in which case we could just as well use `zeroed()` instead.

SUPERCILEX: It seems like the UCG WG hasn't fully flushed out union semantics yet, but Ralf is leaning towards making it bag-o-bytes. Either way, zeroed can't leave padding bytes that are potentially used by one of the variants in an initialized state since it's a union and you don't know which variant will be used. If there are uninit padding bytes, they'd have to be unused by all variants (which seems like what currently happens: https://github.com/rust-lang/unsafe-code-guidelines/issues/354#issuecomment-1191826685)

---

I wonder if it's just better to return `Self { flags: 0 }`. You could argue that if the union is ever expanded, then it's remaining bytes will be uninitialized (I think) which is bad, but then again it's a union and you have no idea what variant it contains anyway. Also Linux can't change it's API so it seems pretty safe to just say flags: 0 and call it good.

jordanisaacs: I am more a fan of using `::mem::zeroed()` as functionally it does the same thing as `Self { ioprio: 0 }` (note: changed the field name). But it won't break if the union is ever extended. As what sunfishcode said makes sense, changed this to be how all unions do the defaults.

I'm not sure if the padding bytes not being zeroed is an issue anyway. Only the unions that are fields of `io_uring_sqe` implement default. As those are going to be aligned correctly, I don't see why the kernel would be touching those uninit padding bytes. As `zeroed()` will zero the largest variant of the union.

PR #547, opened 2023-03-01
io_uring: async cancel flags and msgring types

jordanisaacs:
Missed these additions in my previous pr. Also fixes the `open_flags` type.

The `sqe` union field names will keep getting worse as more operations are added to the fields. Additionally, the rustix sqe struct is continuing to diverge from the kernel header struct, The header sqe struct does not seem to get updated for all the different variations the fields are being used for.

My proposed solution (done in this PR) is that the field names should stay the same as the kernel struct . Only use the `_or_` convention if the union is part of the header. Extra field types should just be added without updating the field name (eg see `ioprio` and `len`).

### DIFF
--- a/src/io_uring.rs
+++ b/src/io_uring.rs
@@ -14,9 +14,11 @@
 //!
 //! # References
 //!  - [Linux]
+//!  - [io_uring header]
 //!
 //! [Linux]: https://man.archlinux.org/man/io_uring.7.en
 //! [io_uring]: https://en.wikipedia.org/wiki/Io_uring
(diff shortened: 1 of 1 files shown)


jordanisaacs:
Sorry for the extra commit & force pushes but this is all good now. I have a working port of `io-uring` library with this branch, [rustix-uring](https://github.com/jordanisaacs/rustix-uring) which I would love to publish onto cargo once this is merged and a new rustix release is published. I explained my proposed solution to field naming in the main PR comment.

sunfishcode:
Looks good. I think your approach to the naming here makes sense.

sunfishcode:
This PR is now released in 0.37.

PR #552, opened 2023-03-03
Switch away from deprecated CI images

LingMan:
- Switch from `macos-10.15` to `macos-11`
- Switch from `ubuntu-18.04` to `ubuntu-20.04`

Should fix the failing CI the last two days. Looks like another brownout before the final removal of `macos-10.15`.


sunfishcode:
Thanks!

PR #555, opened 2023-03-03
Basic fcntl-style locking.

sunfishcode:
Add an `fcntl_lock` function implementing fcntl-style process-associated locking. Currently this only supports locking the entire file.


src/backend/linux_raw/fs/syscalls.rs
```diff
@@ -1022,6 +1022,52 @@ pub(crate) fn fcntl_add_seals(fd: BorrowedFd<'_>, seals: SealFlags) -> io::Resul
...
+            l_type: l_type as _,
+            l_whence: 0,
+            l_start: 0,
+            l_len: 0,
+            ..core::mem::zeroed()
+        };
+        #[cfg(target_pointer_width = "32")]
+        {
+            ret(syscall_readonly!(
+                __NR_fcntl64,
```
cgwalters, line 1053: How about `if cfg!(target_pointer_width = "32") { __NR_fcntl64 } else { __NR_fcntl }` here and dedup the blocks?

src/backend/linux_raw/fs/syscalls.rs
```diff
@@ -1022,6 +1022,52 @@ pub(crate) fn fcntl_add_seals(fd: BorrowedFd<'_>, seals: SealFlags) -> io::Resul
...
+        FlockOperation::NonBlockingUnlock => (F_SETLK, F_UNLCK),
+    };
+
+    unsafe {
+        let lock = flock {
+            l_type: l_type as _,
+            l_whence: 0,
+            l_start: 0,
+            l_len: 0,
+            ..core::mem::zeroed()
```
cgwalters, line 1048: Feels like we could rely on `core::mem::zeroed()` also handling the zero-init for the other members?

cgwalters (approved):
Somewhat superficial LGTM

src/backend/linux_raw/fs/syscalls.rs
sunfishcode: Unfortunately `syscall_readonly!` is a macro that does special things with the first parameter and needs it to be a `__NR_*` literal.

src/backend/linux_raw/fs/syscalls.rs
sunfishcode: Setting `l_len` to zero is a special case in the documented behavior of `fcntl`, so I like keeping it explicit. I've now added more comments about this though.

PR #563, opened 2023-03-18
Check `RUSTC_WRAPPER` instead of `CARGO_RUSTC_WRAPPER`.

sunfishcode:
Similar to dtolnay/anyhow#248, and as suggested [here], check `RUSTC_WRAPPER` instead of `CARGO_RUSTC_WRAPPER`.

[here]: https://github.com/bytecodealliance/rustix/pull/544#discussion_r1140467731


aldanor:
@sunfishcode Would it be worth pushing a new crate version with this?

sunfishcode:
I've now released 0.37.2 and 0.36.10 with this fix.

PR #565, opened 2023-03-20
Only use RUSTC_WRAPPER when non empty

markdingram:
We have a build that sets RUSTC_WRAPPER to either sccache or empty string -> after latest release it is now failing on the empty string case.

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads suggests this should be acceptable:

> Setting this to the empty string overwrites the config and resets cargo to not use a wrapper.

Thanks!


sunfishcode:
Thanks!

sunfishcode:
This is now released in 0.37.3 and 0.36.11.

PR #566, opened 2023-03-27
README.md: Add a note about LFS and y2038

joshtriplett:
I went looking for `fstatvfs64`, didn't find it, and had to trawl
through the source code to realize that `rustix` does the obvious
awesome thing: only expose LFS-safe and y2038-safe APIs.


sunfishcode:
Thanks!

PR #567, opened 2023-03-27
Add dup2_std{in,out,err} safe helpers

joshtriplett:
Doing a dup2 over one of the stdio fds is safe, but currently requires a
three-line unsafe dance: unsafely get an OwnedFd for the target fd,
`dup2` over it, then `std::mem::forget` the OwnedFd so it doesn't get
closed.

Encapsulate that in a trio of safe helpers.

### DIFF
--- a/src/io/dup.rs
+++ b/src/io/dup.rs
@@ -44,6 +44,9 @@ pub fn dup<Fd: AsFd>(fd: Fd) -> io::Result<OwnedFd> {
 /// set `O_CLOEXEC`, use [`dup3`] with [`DupFlags::CLOEXEC`] on platforms which
 /// support it, or [`fcntl_dupfd_cloexec`]
 ///
+/// For `dup2` to stdin, stdout, and stderr, see [`crate::io::stdio::dup2_stdin`],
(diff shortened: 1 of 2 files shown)


src/io/dup.rs
```diff
@@ -78,3 +78,36 @@ pub fn dup2<Fd: AsFd>(fd: Fd, new: &mut OwnedFd) -> io::Result<()> {
 pub fn dup3<Fd: AsFd>(fd: Fd, new: &mut OwnedFd, flags: DupFlags) -> io::Result<()> {
     backend::io::syscalls::dup3(fd.as_fd(), new, flags)
 }
+
+/// Utility function to safely `dup2` over stdin (fd 0).
+#[cfg(not(any(windows, target_os = "wasi")))]
+#[inline]
+#[allow(unsafe_code)]
```
sunfishcode, line 85: Could we move these functions to src/io/stdio.rs? I think "stdio-using things" is the higher-level grouping here, than "dup-related things". And, that file already has a `#![allow(unsafe_code)]` so we wouldn't need these per-function attributes.

sunfishcode:
This is now released in 0.37.4.

PR #568, opened 2023-03-28
Fix build failure on AIX

bzEq:
On AIX, there is also `d_namlen` field.


sunfishcode:
Thanks!

sunfishcode:
This is now released in 0.37.4.

Issue #575, opened 2023-03-28
rustix keeps rebuilding in my dependency tree.


rukai:
My application has rustix in its dependency tree due to clap:
```
├── clap v4.1.11
│   ├── bitflags v2.0.2
│   ├── clap_derive v4.1.9 (proc-macro)
│   │   ├── heck v0.4.1
│   │   ├── proc-macro-error v1.0.4
│   │   │   ├── proc-macro-error-attr v1.0.4 (proc-macro)
│   │   │   │   ├── proc-macro2 v1.0.52 (*)
│   │   │   │   └── quote v1.0.26 (*)
│   │   │   │   [build-dependencies]
│   │   │   │   └── version_check v0.9.4
│   │   │   ├── proc-macro2 v1.0.52 (*)
│   │   │   ├── quote v1.0.26 (*)
│   │   │   └── syn v1.0.109 (*)
│   │   │   [build-dependencies]
│   │   │   └── version_check v0.9.4
│   │   ├── proc-macro2 v1.0.52 (*)
│   │   ├── quote v1.0.26 (*)
│   │   └── syn v1.0.109 (*)
│   ├── clap_lex v0.3.3
│   │   └── os_str_bytes v6.5.0
│   ├── is-terminal v0.4.5
│   │   ├── io-lifetimes v1.0.9
│   │   │   └── libc v0.2.140
│   │   └── rustix v0.36.11
│   │       ├── bitflags v1.3.2
│   │       ├── io-lifetimes v1.0.9 (*)
│   │       ├── libc v0.2.140
│   │       └── linux-raw-sys v0.1.4
│   ├── once_cell v1.17.1
│   ├── strsim v0.10.0
│   └── termcolor v1.2.0
```

I have no idea why but sometimes rustix will just randomly trigger a rebuild.
```
   Compiling rustix v0.36.11
    Checking is-terminal v0.4.5
    Checking clap v4.1.11
    Checking inferno v0.11.15
    Checking shotover-proxy v0.1.9 (/home/rukai2/Projects/Crates/shotover/shotover-proxy/shotover-proxy)
    Checking test-helpers v0.1.0 (/home/rukai2/Projects/Crates/shotover/shotover-proxy/test-helpers)
    Finished dev [unoptimized + debuginfo] target(s) in 6.51s
```

I've never seen this for any other dependency before.

If I observe anything else that might explain whats going on I'll add it to this issue.

sunfishcode:
[...]

Issue #562, opened 2023-03-10
Rustix recompiles every time switching between Intellij and CLI


howardjohn:
This is maybe not even a rustix bug, but may help someone else seeing the same issue.

Rustix's build.rs has `cargo:rerun-if-env-changed=RUSTC`.

When I run `cargo check` from my terminal, I see `RUSTC=rustc`. When I run it from intellij, I get `RUSTC=$HOME/.cargo/bin/rustc`. This means each time the build is invalidated.

For now I am just doing `export RUSTC=$HOME/.cargo/bin/rustc` to work around this

sunfishcode:
Rustix needs to autodetect features from the rustc, in particular to know whether it can used `OwnedFd` from std or whether it needs to use its own copy. If it's possible that `cargo check` in your terminal could use a different version of rustc than what intellij uses, then rustix does need to rerun its build script.

Another consideration here is that https://github.com/intellij-rust/intellij-rust/issues/10186 is a report where rustix may not be running its build script in a situation where it should. It's not yet clear what the problem is.

Also relevant in this space is that my earlier fix had a bug, which I'm now fixing in https://github.com/bytecodealliance/rustix/pull/563.

sunfishcode:
In Rust 1.68, `cargo build --verbose` now includes a `Dirty` message the indicates why it triggered a rebuild, which would be useful to see when this happens.

Nutomic:
Im have the same problem described in this issue, this is the output for rustix from `cargo check --verbose`:
```
[...]

PR #576, opened 2023-03-29
Remove some rerun-if-env-changed lines from build.rs.

sunfishcode:
Now that we appear to have a known-working fix for #526, we can investigate relaxing the fix. If #526 reappears, we now have a known-working state we can quickly revert to.

Reports in #562 and #575 are that having rerun-if-env-changed=RUSTC and rerun-if-env-changed=RUSTC_WRAPPER are triggering spurious rebuilds in some situations, so remove these. In theory, cargo should already be aware of these environment variables. In practice, there were reports of build failures which had the appearance of using an older version of rustc with build.rs output from a newer version of rustc, and it wasn't clear what was happening. But, it's possible that #563 fixes the issue properly. So let's try removing these rerun-if-env-changed lines, and see if the problem resurfaces.

And, rerun-if-env-changed=TARGET is explicitly documented as being unnecessary [here].

[here]: https://doc.rust-lang.org/cargo/reference/build-scripts.html

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -147,11 +147,6 @@ fn main() {
     println!("cargo:rerun-if-env-changed=CARGO_FEATURE_USE_LIBC");
     println!("cargo:rerun-if-env-changed=CARGO_FEATURE_RUSTC_DEP_OF_STD");
     println!("cargo:rerun-if-env-changed=CARGO_CFG_MIRI");
-    println!("cargo:rerun-if-env-changed=CARGO_ENCODED_RUSTFLAGS");
-    println!("cargo:rerun-if-env-changed=RUSTC");
-    println!("cargo:rerun-if-env-changed=TARGET");
-    println!("cargo:rerun-if-env-changed=RUSTC_WRAPPER");
-    println!("cargo:rerun-if-env-changed=PROFILE");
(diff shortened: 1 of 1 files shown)


OvermindDL1:
I can reliably test both states, when it is too strict and when it is not strict enough.  If you give me a branch or checkout to test then I can change my deps to that and test it out?

sunfishcode:
@OvermindDL1 I'd much appreciate it. Would you be able to test the "sunfishcode/build-scripts" branch, which is what's in this PR? https://github.com/bytecodealliance/rustix/tree/sunfishcode/build-scripts

OvermindDL1:
Also went through my normal workflow (cargo-watch test in terminal on one screen, compiling and running in intellij on the other), it's much faster now!

OvermindDL1:
And a quick test with rust-analyzer via emacs and I can't reproduce it there either.

sunfishcode:
This is really great, thanks! I can't think of anything else that urgently needs testing here. It seems very likely that #563 fixed the root issue here. So I'll plan to merge this PR soon and do a point release, and then keep an eye out to see if any problems surface.

PR #577, opened 2023-03-29
Add more epoll flags

notgull:


sunfishcode:
Looks good! The FreeBSD builder failure looks like something already fixed on main.

PR #581, opened 2023-03-29
Add more procctls for FreeBSD

valpackett:
In particular I'm planning on playing around with the reaper API (which actually does inescapable process subtree tracking and allows signaling all descendants of a child no matter how many double-triple-forking has happened in between). Also adding a couple more random things here.


sunfishcode:
With the caveat that I'm not familiar with these APIs at all, these look good.

PR #582, opened 2023-03-29
Add kqueue signal filter

valpackett:
Should be supported everywhere…

Usage example:

```rust
fn main() -> std::io::Result<()> {
    use rustix::io::kqueue::*;
    use rustix::process::Signal;
    let kq = kqueue()?;
    let mut out = Vec::with_capacity(10);
    unsafe {
        kevent(
            &kq,
            &[Event::new(
                EventFilter::Signal {
                    signal: Signal::Ttin,
                    times: 0,
                },
                EventFlags::ADD,
                0,
            )],
            &mut out,
            None,
        )
    }?;
    loop {
        while let Some(e) = out.pop() {
            match e.filter() {
                EventFilter::Signal { signal, times } => eprintln!("{:?} x {}", signal, times),
                EventFilter::Unknown => eprintln!("unk???"),
                _ => eprintln!("wat"),
            }
        }
        unsafe { kevent(&kq, &[], &mut out, None) }?;
    }
}
```

While here, I've also documented the output vector capacity thing.

### DIFF
--- a/examples/kq.rs
+++ b/examples/kq.rs
@@ -1,0 +1,68 @@
+#[cfg(bsd)]
+fn main() -> std::io::Result<()> {
+    use rustix::io::kqueue::*;
+    #[cfg(feature = "fs")]
+    use rustix::{fd::AsRawFd, fs};
+
+    let kq = kqueue()?;
+    let mut out = Vec::with_capacity(10);
+
+    #[cfg(feature = "fs")]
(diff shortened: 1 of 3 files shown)


sunfishcode:
Would it make sense to put that example in `examples/`?

valpackett:
I kinda assumed you wouldn't want more random junk in `examples/`… Now extended it to a fuller kqueue example with three different events and added.

sunfishcode:
It's true, we don't want random junk in there, but my vague sense is that kqueue is an interesting enough API that it's worth having an example to show how to get started with it.

PR #583, opened 2023-03-30
Implement From between RawMode and Mode

joshtriplett:
This allows, for instance, `0o700.into()` as a `Mode` argument.


sunfishcode:
Makes sense.

PR #584, opened 2023-03-30
Add setresuid_thread and setresgid_thread

joshtriplett:
These functions support setting the real, effective, and saved UID/GID
in one call.

### DIFF
--- a/src/thread/id.rs
+++ b/src/thread/id.rs
@@ -40,6 +40,29 @@ pub fn set_thread_uid(uid: Uid) -> io::Result<()> {
     backend::thread::syscalls::setuid_thread(uid)
 }
 
+/// `setresuid(ruid, euid, suid)`
+///
+/// # Warning
+///
+/// This is not the setresxid you are looking for... POSIX requires xids to be
+/// process granular, but on Linux they are per-thread. Thus, this call only
+/// changes the xid for the current *thread*, not the entire process even
+/// though that is in violation of the POSIX standard.
+///
+/// For details on this distinction, see the C library vs. kernel differences
+/// in the [man page][linux_notes] and the notes in [`set_thread_uid`]. This
+/// call implements the kernel behavior.
+///
+/// # References
+///  - [Linux]
+///
+/// [Linux]: https://man7.org/linux/man-pages/man2/setresuid.2.html
+/// [linux_notes]: https://man7.org/linux/man-pages/man2/setresuid.2.html#NOTES
+#[inline]
+pub fn set_thread_res_uid(ruid: Uid, euid: Uid, suid: Uid) -> io::Result<()> {
+    backend::thread::syscalls::setresuid_thread(ruid, euid, suid)
+}
+
 /// `setgid(gid)`
 ///
 /// # Warning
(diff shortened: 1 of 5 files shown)


src/thread/id.rs
```diff
@@ -40,6 +40,30 @@ pub fn set_thread_uid(uid: Uid) -> io::Result<()> {
...
+/// For details on this distinction, see the C library vs. kernel differences
+/// in the [setuid man page][linux_notes]. This call implements the kernel behavior.
+///
+/// # References
+///  - [POSIX]
+///  - [Linux]
+///
+/// [POSIX]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/setuid.html
+/// [Linux]: https://man7.org/linux/man-pages/man2/setuid.2.html
+/// [linux_notes]: https://man7.org/linux/man-pages/man2/setuid.2.html#NOTES
```
sunfishcode, line 61: The Linux man page links should point to https://man7.org/linux/man-pages/man2/setresuid.2.html

src/thread/id.rs
```diff
@@ -63,3 +87,27 @@ pub fn set_thread_uid(uid: Uid) -> io::Result<()> {
...
+/// For details on this distinction, see the C library vs. kernel differences
+/// in the [setgid man page][linux_notes]. This call implements the kernel behavior.
+///
+/// # References
+///  - [POSIX]
+///  - [Linux]
+///
+/// [POSIX]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/setgid.html
+/// [Linux]: https://man7.org/linux/man-pages/man2/setgid.2.html
+/// [linux_notes]: https://man7.org/linux/man-pages/man2/setgid.2.html#NOTES
```
sunfishcode, line 109: As above, these links should point to https://man7.org/linux/man-pages/man2/setresgid.2.html

sunfishcode (commented):
On arm and x86, `__NR_setresuid` only supports 16-bit UIDs, so we'll need to use `__NR_setresuid32` instead. [This note](https://man7.org/linux/man-pages/man2/setresuid.2.html#NOTES) has more info.

And it looks like sparc only has `setresuid32` and not `setresuid`, so we'll need to add a `cfg` to the `__NR_setresuid` case for sparc.

And similar for `setresgid`.

src/thread/id.rs
```diff
@@ -40,6 +40,30 @@ pub fn set_thread_uid(uid: Uid) -> io::Result<()> {
...
+/// though that is in violation of the POSIX standard.
+///
+/// For details on this distinction, see the C library vs. kernel differences
+/// in the [setuid man page][linux_notes]. This call implements the kernel behavior.
+///
+/// # References
+///  - [POSIX]
+///  - [Linux]
+///
+/// [POSIX]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/setuid.html
```
sunfishcode, line 59: It appears `setresuid` and `setresgid` are not in POSIX at all, so we should remove the POSIX references.

joshtriplett: The POSIX reference was to explain the whole-process requirement, which is why I changed the reference to say "the setuid manpage". Do you want me to remove that? Or perhaps just cross-reference the note under setuid?

joshtriplett:
> On arm and x86, `__NR_setresuid` only supports 16-bit UIDs, so we'll need to use `__NR_setresuid32` instead. [This note](https://man7.org/linux/man-pages/man2/setresuid.2.html#NOTES) has more info.
> 
> And it looks like sparc only has `setresuid32` and not `setresuid`, so we'll need to add a `cfg` to the `__NR_setresuid` case for sparc.
> 
> And similar for `setresgid`.

Right! I remember seeing the 32-bit syscalls, and then forgot when I got to the patch. Will update.

joshtriplett:
To what degree does rustix need to be compatible with systems that only support the 16-bit syscalls? Would it be reasonable to just call the 32-bit ones only?

src/thread/id.rs
sunfishcode: Just cross-referencing the note under `setuid` sounds good to me.

src/thread/id.rs
joshtriplett: Done.

src/thread/id.rs
joshtriplett: Done.

joshtriplett:
Done. The current version should address everything, assuming CI passes.

sunfishcode:
Thanks!

Issue #578, opened 2023-03-29
0.38 release planning


sunfishcode:
As of this writing, Rustix 0.37 was released about a month ago, and many crates started adopting it a few days ago. We hopefully won't need another semver bump for quite a while.

This issue is for collecting changes that we should make in 0.38, when the time comes. So far, I'm aware of:

 - [x] `chmodat_with` introduced in #541 should be renamed to just `chmodat`.
 - [x] The epoll `EventVec` iterator should return `Event`s rather than tuples of their parts. If we publicly expose the `Event` struct, we can also add a `EventVec::from_raw_parts`, which is needed by mustang.
 - [x] Update to `bitflags` 2.x
   - [x] ~~Use the new - [`BitFlags` trait](https://docs.rs/bitflags/latest/bitflags/trait.BitFlags.html) to simplify conv.rs~~
 - [x] Harmonizing the types between libc and linux_raw as discussed in https://github.com/bytecodealliance/rustix/issues/513
   - Some of this has been done; there's more to do, but we don't need to block 0.38 on it.
 - [x] #693 
 - [x] Finer-grained API modules
   - [x] Consider splitting some things out of `rustix::io`, such as the various polling APIs into a `rustix::poll` or so, and maybe moving the procfs API to `rustix::procfs`.
   - [x] Split "system" and "membarrier" APIs out of `process` and into a new `system` module. Maybe also `fs::sync`?
 - [x]  Rename `getsockopt_nosigpipe` and `setsockopt_nosigpipe` to `get_socket_nosigpipe` and `set_socket_nosigpipe`, for consistency with the rest of the getsockopt/setsockopt function names.
 - [x] Rename `sendmsg_noaddr` to `sendmsg`
 - [x]  - [`AcceptFlags`](https://docs.rs/rustix/latest/rustix/net/struct.AcceptFlags.html) is now just an alias for - [`SocketFlags`](https://docs.rs/rustix/latest/rustix/net/struct.SocketFlags.html), so we should remove it.
 - [x] Move `poll`, `epoll`, `kqueue`, and `port` out of `rustix::io`, and put them behind a feature flag
 - [x] https://github.com/bytecodealliance/rustix/pull/615
 - [x] https://github.com/bytecodealliance/rustix/issues/604
 - [x] https://github.com/bytecodealliance/rustix/pull/502
 - [x] https://github.com/bytecodealliance/rustix/issues/695

sunfishcode:
[...]

PR #590, opened 2023-04-02
Expand waitid support, add waitid tests

valpackett:
- add a bunch of stuff to `WaitidStatus` 
- detect successful return with no process more carefully according to documentation
  - Linux manpage:
    > If WNOHANG was specified in options and there were no children in a waitable state, then waitid() returns 0 immediately and the state of the siginfo_t structure pointed to by infop is unspecified. To distinguish this case from that where a child was in a waitable state, zero out the si_pid field before the call and check for a nonzero value in this field after the call returns.
  - BSD manpages don't really mention the unspecified thing e.g.:
    > If WNOHANG was specified and waitid() didn't find a matching process, si_signo and si_pid will be set to zero.
- add tests
- would it be ok to underive `Copy` on `WaitidStatus`? Technically an API break that should go to #578? But `WaitidStatus` was currently completely unusable with nothing on it…


valpackett:
wow, `siginfo_t` is such a royal mess… at least in the `libc` crate there are accessors, just some missing on some platforms, but for the raw linux backend… I guess we'll have to make an extension trait with accessors in `backend::linux_raw::c`?

valpackett:
Oofff okay, now it builds everywhere and isn't actually *too* bad, tho some `cfg()`'s are quite silly.

src/backend/libc/process/syscalls.rs
```diff
@@ -426,7 +426,7 @@ pub(crate) fn waitid(id: WaitId<'_>, options: WaitidOptions) -> io::Result<Optio
 #[cfg(not(any(target_os = "wasi", target_os = "redox", target_os = "openbsd")))]
 #[inline]
 fn _waitid_all(options: WaitidOptions) -> io::Result<Option<WaitidStatus>> {
-    let mut status = MaybeUninit::<c::siginfo_t>::uninit();
+    let mut status = MaybeUninit::<c::siginfo_t>::zeroed();
```
sunfishcode, line 429: Is the reason for explicit zeroing here the comment in the Linux man page about zeroing out the `si_pid` field? If so, could you quote that documentation here, so that it's recorded in the code why this uses zeroed memory?

src/process/wait.rs
```diff
@@ -99,9 +112,145 @@ impl WaitStatus {
...
+        }
+    }
+
+    /// Returns a reference to the raw platform-specific `siginfo_t` struct.
+    #[inline]
+    pub const fn as_raw(&self) -> &backend::c::siginfo_t {
+        &self.0
+    }
+
+    #[cfg(linux_raw)]
```
sunfishcode, line 213: Rustix does have a few places where the public API for `linux_raw` differs from the public API for the libc backend on Linux, but we're working on eliminating them, as it makes code less portable. Is it complicated to add Linux support using the libc backend as well? Also, yeah, your suggestion to push these details into src/backend/* sounds good; perhaps in new src/backend/*/wait.rs files?

valpackett: It should already work with the libc backend (CI should test this, right?) — it's the final `not(any(…))` version. The libc crate does make an effort to offer the same interface here on all platforms, just lacking in a few places such as https://github.com/rust-lang/libc/pull/3185 and OpenBSD/NetBSD currently.

sunfishcode: Ah, I see now. I was confused by the `cfg`s :-).

valpackett:
Okay, now it looks decent I guess…

sunfishcode:
Cool, thanks!

pawelchcki:
Thanks for this change. I was just looking for the missing waitidstatus methods! :tada:

sunfishcode:
This is now released in rustix 0.37.8.

PR #605, opened 2023-04-08
kqueue: fix timers

valpackett:
They were passing the time as ident instead of data. Somehow this worked with non-oneshot timers, but was completely broken otherwise.

---

This is a breaking change… but the previous behavior was *completely* broken. We can't just keep the API and pass the same `ident=0` and "let users use `udata` for identifying timers", that would mean there's only one timer ever. And right now timers with the same delay would clash. Events are identified by `ident,type`, after all.

### DIFF
--- a/src/io/kqueue.rs
+++ b/src/io/kqueue.rs
@@ -3,11 +3,8 @@
 use crate::fd::{AsFd, AsRawFd, OwnedFd, RawFd};
 use crate::{backend, io};
 
-use backend::c::{self, kevent as kevent_t, uintptr_t};
+use backend::c::{self, intptr_t, kevent as kevent_t, uintptr_t};
 use backend::io::syscalls;
-
-#[cfg(any(apple, freebsdlike))]
-use backend::c::intptr_t;
 
 use alloc::vec::Vec;
 use core::ptr::slice_from_raw_parts_mut;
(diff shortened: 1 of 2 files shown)


src/io/kqueue.rs
```diff
@@ -198,7 +202,13 @@ pub enum EventFilter {
     },
 
     /// A timer filter.
-    Timer(Option<Duration>),
+    Timer {
+        /// The identifier for this event.
+        ident: intptr_t,
```
sunfishcode, line 207: Does this ever hold an actual pointer value? If so, rustix's convention is to use an actual raw pointer type rather than `intptr_t`, to better follow strict provenance.

sunfishcode (commented):
I'm not familiar with the `kqueue` API in detail, but at a quick glance this looks good to me.

@notgull Would it be ok with you if we made this change in 0.37 without a semver bump?

src/io/kqueue.rs
valpackett: No, it's an arbitrary ID number; I've just copied `intptr_t` from another variant here. On FreeBSD for example kqueue idents are `__uintptr_t`, but I guess signed was chosen due to what other platforms have

sunfishcode:
Thanks!

sunfishcode:
This is now released in 0.37.12.

PR #608, opened 2023-04-18
Implement `test_kill_process{,_group}`

oxalica:
They pass zero as signal to test the ability to kill() without actually sending signals. This behavior is part of POSIX.
https://pubs.opengroup.org/onlinepubs/9699919799/functions/kill.html

Rationale: `0` is not a valid `Signal` so should not be added as a new variant. Changing existing API of `kill_process*` to accept `Option<Signal>` is breaking and complicates the most usual usage of `kill()`. Thus I added two additional APIs instead.

Note sure about the naming, `test_` prefix reminds me about "unit tests". But IMO its better than `try_` (actually performs actions sometimes?) or `check_` (returns `bool`? assertion?).

### DIFF
--- a/src/backend/libc/process/syscalls.rs
+++ b/src/backend/libc/process/syscalls.rs
@@ -548,6 +548,18 @@ pub(crate) fn kill_current_process_group(sig: Signal) -> io::Result<()> {
     unsafe { ret(c::kill(0, sig as i32)) }
 }
 
+#[cfg(not(target_os = "wasi"))]
+#[inline]
+pub(crate) fn test_kill_process(pid: Pid) -> io::Result<()> {
+    unsafe { ret(c::kill(pid.as_raw_nonzero().get(), 0)) }
+}
+
+#[cfg(not(target_os = "wasi"))]
+#[inline]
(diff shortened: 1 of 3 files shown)


sunfishcode:
Leaving the existing `kill_process` functions as-is and adding new `test_*` functions sounds reasonable to me, under the umbrella of "demultiplexing" and the theory that use cases should practically always know which form they want at compile time.

My one comment here is: could you add comments to the documentation for the regular `kill_process` and `kill_process_group` which point users to `test_kill_process` and `test_kill_process_group` if they want to send a 0 signal?

oxalica:
> could you add comments to the documentation for the regular `kill_process` and `kill_process_group` which point users to `test_kill_process` and `test_kill_process_group` if they want to send a 0 signal?

The 0 pid behavior (`kill_current_process_group`) is not mentioned either. Are they really needed? With `#[doc(alias = "kill")]`, users can search keyword `kill` to get the list of all functions backed by `kill()` syscall.

sunfishcode:
Sounds reasonable!

PR #614, opened 2023-04-20
Enable available termios flags on macOS

joshtriplett:
These flags seem to exist on macOS, and this helps portable code run on
macOS.

I checked many of these to confirm they exist; let's see if this passes CI.

### DIFF
--- a/examples/stdio.rs
+++ b/examples/stdio.rs
@@ -66,39 +66,30 @@ fn show<Fd: AsFd>(fd: Fd) -> io::Result<()> {
             }
 
             print!(" - in flags:");
-            #[cfg(not(apple))]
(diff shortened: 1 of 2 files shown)


joshtriplett:
Working on converting the types over to Tcflag.

sunfishcode:
This looks great! I don't recall why rustix ended up with all these disabled on `cfg(apple)`.

In case it's of interest, I have a work-in-progress patch which modernizes rustix' termios API using `bitflags` and `enum`s instead of quite so many raw integer constants: https://github.com/bytecodealliance/rustix/pull/615. I'll port these changes to that PR.

sunfishcode:
This is now released in 0.37.13.

PR #622, opened 2023-04-21
In no-std mode, disable the std feature in libc.

sunfishcode:


Cargo.toml
```diff
@@ -121,7 +121,7 @@ default = ["std", "use-libc-auxv"]
 
 # This enables use of std. Disabling this enables `#![no_std], and requires
 # Rust 1.64 or newer.
-std = ["io-lifetimes"]
+std = ["io-lifetimes", "libc/std"]
```
taiki-e, line 124: Note that this will enable a dependency on libc whenever the std feature is enabled. To work around it, you need [weak dependency features (`?` syntax) that require Rust 1.60](https://doc.rust-lang.org/stable/cargo/reference/features.html#dependency-features).

sunfishcode: Oh, good point! This PR isn't urgently needed by anything, so I think I'll just make a note about this and close the PR for now.

PR #630, opened 2023-04-25
Update the bitflags dependency to 1.3.2.

### LINKED ISSUE #629: Incorrect bitflags dependency version
34841fc25bf1a0cdcb043295e0bf515f4dd3a77a introduced a call to bitflags' `difference()` function, which was only introduced in bitflags 1.3.1 (maybe 1.3.0, but that was yanked). rustix only specifies 1.2.1 as dependency though.

This caused build failures after the 0.37.14 release. I'm actually a bit surprised `cargo update` doesn't also bump the bitflags version to 1.3.2, but the dependency is wrong either way.

sunfishcode:
Fixes #629.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ rust-version = "1.48"
 cc = { version = "1.0.68", optional = true }
 
 [dependencies]
-bitflags = "1.2.1"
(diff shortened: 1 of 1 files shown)


sunfishcode:
This is released in 0.37.15.

PR #636, opened 2023-04-28
Use PR_GET_AUXV if available; fall back to /proc/self/auxv

joshtriplett:
This avoids relying on /proc being mounted, and on file operations; it
also reduces the total number of syscalls.

For a first pass, I've kept the fallback to /proc/self/auxv on any error. In
the future, it might make sense to have some combination of feature flags that
*only* supports this path, but that would add more complexity to handle.

### DIFF
--- a/src/backend/linux_raw/arch/mod.rs
+++ b/src/backend/linux_raw/arch/mod.rs
@@ -133,6 +133,85 @@ macro_rules! syscall {
     };
 }
 
+// Macro to invoke a syscall that always uses direct assembly, rather than the vdso. Useful when
+// still finding the vdso.
+#[allow(unused_macros)]
+macro_rules! syscall_always_asm {
+    ($nr:ident) => {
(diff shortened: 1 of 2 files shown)


sunfishcode:
Thanks for working on this, and for adding this feature to the kernel to support it! :smile: The PR looks good, and I agree it makes sense to consider flags for making this the only path in the future. I do think this should wait until Linux 6.4 is released with `PR_GET_AUXV`, so that we don't depend on unreleased features.

joshtriplett:
@sunfishcode This has now been merged upstream, and released in 6.4-rc1 and 6.4-rc2. Would that suffice for updating and merging this?

joshtriplett:
@sunfishcode No, that's theoretically possible. We can hold off until 6.4 then.

joshtriplett:
Linux 6.4 just shipped, and I've rebased this to apply to current rustix.

sunfishcode:
Thanks!

PR #637, opened 2023-04-28
Provide safe wrapper for EXT4_IOC_RESIZE_FS

joshtriplett:
I'm assuming that you don't want a production rustix to depend on a prerelease
linux-raw-sys. Let me know how you'd like to handle this.

### DIFF
--- a/src/backend/libc/io/syscalls.rs
+++ b/src/backend/libc/io/syscalls.rs
@@ -373,6 +373,18 @@ pub(crate) fn ioctl_ficlone(fd: BorrowedFd<'_>, src_fd: BorrowedFd<'_>) -> io::R
     }
 }
 
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[inline]
+pub(crate) fn ext4_ioc_resize_fs(fd: BorrowedFd<'_>, blocks: u64) -> io::Result<()> {
+    unsafe {
+        ret(c::ioctl(
+            borrowed_fd(fd),
+            linux_raw_sys::ioctl::EXT4_IOC_RESIZE_FS as _,
+            &blocks,
+        ))
+    }
+}
+
 #[cfg(not(any(target_os = "redox", target_os = "wasi")))]
 #[cfg(all(feature = "fs", feature = "net"))]
 pub(crate) fn is_read_write(fd: BorrowedFd<'_>) -> io::Result<(bool, bool)> {
--- a/src/backend/libc/io/syscalls.rs
+++ b/src/backend/libc/io/syscalls.rs
@@ -376,13 +376,14 @@ pub(crate) fn ioctl_ficlone(fd: BorrowedFd<'_>, src_fd: BorrowedFd<'_>) -> io::R
 #[cfg(any(target_os = "android", target_os = "linux"))]
 #[inline]
 pub(crate) fn ext4_ioc_resize_fs(fd: BorrowedFd<'_>, blocks: u64) -> io::Result<()> {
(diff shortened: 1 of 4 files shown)


sunfishcode:
I posted https://github.com/sunfishcode/linux-raw-sys/pull/57 to add the definitions of the ext4 ioctl codes to linux-raw-sys, as a temporary workaround. This provides the needed definitions while avoiding adding any of the unreleased bits from the Linux git tree.

sunfishcode:
linux-raw-sys 0.3.6 now includes `EXT4_IOC_RESIZE_FS`.

joshtriplett:
Updated.

joshtriplett:
Rebased atop changes in main that fix CI.

sunfishcode:
Or rather, someone will need to rewrite the generator. I don't know when I'll have time for this. The current situation is:
 - bindgen doesn't get the values of any ioctl codes because they're defined with somewhat complex macros and bindgen only understands simple constants
 - linux-raw-sys uses a generator which works by generating C programs that print the value of the ioctl codes to a header file which bindgen can be run on, and running them under qemu
 - qemu-sparc64 is segfaulting

Alternatively, one could send a PR to the libc crate adding `EXT4_IOC_RESIZE_FS` there, and then use that instead of linux_raw_sys in the libc backend.

sunfishcode:
Or, it'd also be ok to just hard-code a constant in rustix on sparc for now.

src/backend/libc/io/syscalls.rs
```diff
@@ -373,6 +373,18 @@ pub(crate) fn ioctl_ficlone(fd: BorrowedFd<'_>, src_fd: BorrowedFd<'_>) -> io::R
...
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[inline]
+pub(crate) fn ext4_ioc_resize_fs(fd: BorrowedFd<'_>, blocks: u64) -> io::Result<()> {
+    unsafe {
+        ret(c::ioctl(
+            borrowed_fd(fd),
+            linux_raw_sys::ioctl::EXT4_IOC_RESIZE_FS as _,
+            &blocks,
+        ))
+    }
```
sunfishcode, line 385: // TODO: Fix linux-raw-sys to define ioctl codes for sparc.
    #[cfg(any(target_arch = "sparc", target_arch = "sparc64"))]
    const EXT4_IOC_RESIZE_FS: u32 = 0x8008_6610;

    #[cfg(not(any(target_arch = "sparc", target_arch = "sparc64")))]
    use linux_raw_sys::ioctl::EXT4_IOC_RESIZE_FS;

    unsafe { ret(c::ioctl(borrowed_fd(fd), EXT4_IOC_RESIZE_FS as _, &blocks)) }

sunfishcode:
I took the liberty of updating this to hard-code a constant for sparc.

sunfishcode:
This is now released in rustix 0.37.16

joshtriplett:
And works great, thank you!

Karrq:
Hey I believe [this](https://github.com/bytecodealliance/rustix/blob/main/src/backend/libc/io/syscalls.rs#L384) broke building rustix under miri...
From what I can tell, the `ioctl` feature  isn't enabled for `linux_raw_sys`  under [miri](https://github.com/bytecodealliance/rustix/blob/main/Cargo.toml#L59)

sunfishcode:
@Karrq Thanks for the report! That issue is fixed in #645, which is now released in rustix 0.37.18.

PR #643, opened 2023-04-29
Re-enable FreeBSD CI

taiki-e:
Reverts #600 and fixes FreeBSD CI failure.

`curl` is already installed in the FreeBSD image. So removing the installation of `curl` fixes the problem.


sunfishcode:
Thanks!

PR #645, opened 2023-04-29
Add `ioctl` cargo feature for `linux-raw-sys` ...

zachs18:
... when using `libc` backend.

May fix issue described in https://github.com/bytecodealliance/rustix/pull/637#issuecomment-1528848589 and [CI errors for `nix`](https://github.com/nix-rust/nix/pull/2029/checks?check_run_id=13122856368) under some Linux and Android targets, when using `rustix` `>=0.37.16`. (Tested locally, it does fix `nix`'s test build errors on `mips-unknown-linux-gnu`)


sunfishcode:
Yes, this looks like the correct fix, thanks!

sunfishcode:
This is now released in 0.37.18.

PR #647, opened 2023-04-30
Fix no-std builds for MacOS and BSDs

niluxv:
Now to my knowledge all targets except windows compile fine for no-std again.

### DIFF
--- a/src/net/send_recv/msg.rs
+++ b/src/net/send_recv/msg.rs
@@ -168,14 +168,12 @@ impl<'buf, 'slice, 'fd> SendAncillaryBuffer<'buf, 'slice, 'fd> {
         self.length = new_length;
 
         // Get the last header in the buffer.
-        let mut last_header = leap!(messages::Messages::new(buffer).last());
+        let last_header = leap!(messages::Messages::new(buffer).last());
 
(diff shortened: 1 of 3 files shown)


niluxv:
That CI failure seems unrelated to my changes, but I don't really understand the error

```
 error: variable does not need to be mutable
   --> src/net/send_recv/msg.rs:171:13
    |
171 |         let mut last_header = leap!(messages::Messages::new(buffer).last());
    |             ----^^^^^^^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `-D unused-mut` implied by `-D warnings`
```

It says unused mut but as far as I can see the variable needs to be mutable:

```rust
        // Get the last header in the buffer.
        let mut last_header = leap!(messages::Messages::new(buffer).last());

        // Set the header fields.
        unsafe {
            last_header.cmsg_len = c::CMSG_LEN(source_len) as _;
            last_header.cmsg_level = cmsg_level;
            last_header.cmsg_type = cmsg_type;
        }
```

sunfishcode:
Thanks!

sunfishcode:
This is now released in rustix 0.37.19.

PR #651, opened 2023-05-03
Update bitflags dependency to 2.2.1.

qwandor:
This requires some minor code changes, mostly around deriving traits explicitly as they are no longer derived by default by the bitflags macro.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        build: [stable, nightly, 1.48]
+        build: [stable, nightly, 1.56]
         include:
           - build: stable
             os: ubuntu-latest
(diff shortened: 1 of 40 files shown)


qwandor:
This is needed to unblock https://github.com/clap-rs/clap/pull/4861, which in turn is needed to avoid having two different version of bitflags in Android.

qwandor:
It looks like this will mean updating the MSRV to 1.56, to allow the 2021 edition.

sunfishcode:
Rustix's current MSRV policy is to support Debian stable, and a new Debian stable release is expected [in about a month from now](https://lwn.net/ml/debian-devel-announce/E1ps5uu-00FaGp-U5@respighi.debian.org/). So one option here is to wait for that before updating here.

Is there a need to move more quickly here?

oxalica:
The Debian stable release is out. Could we make this forward to 1.63?
https://packages.debian.org/bookworm/rustc

oxalica:
> The Debian stable release is out. Could we make this forward to 1.63? https://packages.debian.org/bookworm/rustc

Oh, there's already a separated one #687

sunfishcode:
Yes, I apologize, I forget about this PR and ended up making a separate patch to update to bitflags 2. That patch is now merged, so we can now close this.

PR #657, opened 2023-05-04
Reconsider guarantees of Pid and refactor

### LINKED ISSUE #604: `Pid::from_raw{,_nonzero}` should not be `unsafe`
As the docs say,
>  the process it refers to could disappear at any time and be replaced by another, unrelated, process.

It provides no guarantee about the validness of the pid, and is impossible to do so. It's just a wrapper for an non-zero integer. A runtime-invalid `Pid` fed into `pidfd_open` or `kill_process` results in `ESRCH`, without breaking either IO safety or memory safety.

oxalica:
Fixes #604.

## Summary

```rust
mod backend {
    mod linux_raw {
        pub type RawPid = i32;
        pub type RawNonZeroPid = core::num::NonZeroI32;
    }
}

pub mod process {
    impl Pid {
        pub const fn from_raw(raw: RawPid) -> Option<Self>;
        pub const unsafe fn from_raw_unchecked(raw: RawPid) -> Self;

        // Removed.
        // pub const unsafe fn from_raw_nonzero(raw: RawNonZeroPid) -> Self;
    }
}
```

## Things done & Rationales

1.  `Pid` and `RawPid` use `NonZeroI32` and `i32` now. `Pid` are guaranteed to be in range `1..=i32::MAX`.

    Notice that pid range should be enforced for all APIs, and negative values are exploited by Linux for different functionalities (eg, kill(2) a process group instead of a process). We should validate that range to prevent accidental misfunction, no matter for `i32` or `u32`.

    So here we consistently use a wrapped `i32` for both linux_raw and libc backend. Two constructor `Pid::from_raw{,_unchecked}` are provided to either do or bypass the check. These APIs also simplify code a lot.

2.  `Pid::from_raw_nonzero` is removed.

    As mentioned in (1), the range guarantee is stricter now and only the nonzero property cannot help much.

    **This is a breaking change.**

3.  `Pid::from_raw` is safe now.

    We are not possible to make any guarantee for an arbitrary non-children pid at all, and an invalid pid causes nothing more than a logic error.

## Unresolved

With the changes made, `Pid` (range checked) and `RawPid` (no restriction) covers almost all usages. `RawNonZeroPid` is under-checked for syscalls since `i32::MAX as u32 + 1` is type-valid but will malfunction. Is `RawNonZeroPid` still useful after this PR? If no, should we make it private as a implementation detail?

### DIFF
--- a/src/process/id.rs
+++ b/src/process/id.rs
@@ -33,7 +33,7 @@ pub use crate::ugid::{Gid, Uid};
 #[derive(Copy, Clone, Eq, PartialEq, Debug, Hash)]
 pub struct Cpuid(RawCpuid);
 
-#[cfg(linux_kernel)]
+#[cfg(any(target_os = "android", target_os = "linux"))]
 impl Cpuid {
     /// Converts a `RawCpuid` into a `Cpuid`.
     ///
--- a/src/process/id.rs
+++ b/src/process/id.rs
@@ -21,9 +21,6 @@ pub use crate::ugid::RawGid;
 /// The raw integer value of a Unix process ID.
 pub use crate::pid::RawPid;
 
-/// The raw integer value of a Unix process ID.
-pub use crate::pid::RawNonZeroPid;
-
 pub use crate::pid::Pid;
 pub use crate::ugid::{Gid, Uid};
 
--- a/src/backend/libc/pid/syscalls.rs
+++ b/src/backend/libc/pid/syscalls.rs
@@ -1,7 +1,7 @@
 //! libc syscalls for PIDs
 
 use crate::backend::c;
-use crate::pid::{Pid, RawNonZeroPid};
+use crate::pid::Pid;
 
 #[cfg(not(target_os = "wasi"))]
 #[inline]
@@ -9,7 +9,6 @@ use crate::pid::{Pid, RawNonZeroPid};
 pub(crate) fn getpid() -> Pid {
     unsafe {
         let pid = c::getpid();
-        debug_assert_ne!(pid, 0);
-        Pid::from_raw_nonzero(RawNonZeroPid::new_unchecked(pid))
+        Pid::from_raw_unchecked(pid)
     }
 }
--- a/src/backend/libc/process/syscalls.rs
+++ b/src/backend/libc/process/syscalls.rs
@@ -26,7 +26,7 @@ use crate::process::Uid;
 #[cfg(linux_kernel)]
 use crate::process::{Cpuid, MembarrierCommand, MembarrierQuery};
 #[cfg(not(target_os = "wasi"))]
-use crate::process::{Gid, Pid, RawNonZeroPid, RawPid, Signal, WaitOptions, WaitStatus};
+use crate::process::{Gid, Pid, RawPid, Signal, WaitOptions, WaitStatus};
 #[cfg(not(any(target_os = "fuchsia", target_os = "redox", target_os = "wasi")))]
 use crate::process::{Resource, Rlimit};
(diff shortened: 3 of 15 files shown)


src/process/id.rs
```diff
@@ -106,44 +106,52 @@ impl Gid {
...
     ///
-    /// `raw` must be the value of a valid Unix process ID. It must not be
-    /// zero.
+    /// The caller must guarantee `raw` is strictly positive.
     #[inline]
-    pub const unsafe fn from_raw_nonzero(raw: RawNonZeroPid) -> Self {
-        Self(raw)
+    pub const unsafe fn from_raw_unchecked(raw: RawPid) -> Self {
+        // FIXME: `const_panic` is stabilized since Rust 1.57.
+        // debug_assert!(raw > 0);
```
oxalica, line 143: Ideally, we want to have this `debug_assert`. But it needs to bump MSRV.
It is not required to be correct, though.

sunfishcode: It's looking like we'll bump the MSRV at our next semver bump, so this should be ok. If you want, you can bump the MSRV in this PR to 1.63, so that you can re-enable this assert.

oxalica: > If you want, you can bump the MSRV in this PR to 1.63, so that you can re-enable this assert.

It'll collide with #651 I think?

sunfishcode: If you keep it as a separate commit, it should be pretty easy to drop when it's time to merge things.

1.63 is the version of Rust in the [next Debian stable release](https://packages.debian.org/bookworm/rustc), which is what rustix's MSRV policy is currently based on. It's expected to be released [next month](https://lists.debian.org/debian-devel-announce/2023/04/msg00007.html).

oxalica: > If you keep it as a separate commit, it should be pretty easy to drop when it's time to merge things.
> 
> 1.63 is the version of Rust in the [next Debian stable release](https://packages.debian.org/bookworm/rustc), which is what rustix's MSRV policy is currently based on. It's expected to be released [next month](https://lists.debian.org/debian-devel-announce/2023/04/msg00007.html).

I prefer waiting for another individual PR to do such a big MSRV bump, after the official Debian release. I'm afraid there are many unrelated code should also be cleaned up (both to make CI happy and avoid outdated comments).  Not a fan of mixing them here.

sunfishcode:
This looks good! I had introduced a bunch of merge conflicts with some earlier refactorings, so I took the liberty here to resolve the merge conflicts, and now that the MSRV is Rust 1.63, I uncommented the `debug_assert`.

@valpackett FYI, This changes the `pid` return value of `get_reaper_status` from `Pid` to `Option<Pid>`, since in CI testing it appears the value `-1` can be returned. The documentation does say "The rs_pid returns the pid of one reaper child if there are any descendants". Also, [the implementation does appear to use -1 to indicate three are no children](https://github.com/freebsd/freebsd-src/blob/b7198dcfc03967cba191a373d99df47ee52d6e2a/sys/kern/kern_procctl.c#L187).

oxalica:
@sunfishcode Thanks!
Seems there are [still some](https://github.com/bytecodealliance/rustix/pull/657/files#diff-10cdf164063d2c54539d1aa8db2aa24d650ec72be0075c2d52f14829c2760c63R16) `debug_assert!({p,t}id > 0);` right before `Pid::from_raw_unchecked` not removed. It is already contained by `from_raw_unchecked` now.

sunfishcode:
@oxalica Ah, I likely reintroduced those by accident when resolving the merge conflicts. Do you want to submit a PR removing them?

oxalica:
> @oxalica Ah, I likely reintroduced those by accident when resolving the merge conflicts. Do you want to submit a PR removing them?

#694

PR #659, opened 2023-05-04
Fix the `test_ioctl_ficlone` test to work on more filesystems.

sunfishcode:
Rename the typo `test_ioctl_fioclone` to `test_ioctl_ficlone`, and fix it to avoid depending on the ioctl always failing, as there are filesystems on which it doesn't always fail.


oxalica (approved):
The test now passes for me.

Issue #660, opened 2023-05-06
Bug in recvmsg msghdr decoding?


psychon:
Hi,

@notgull added support for `recvmsg` in #505 and then wanted to use that code in x11rb: https://github.com/psychon/x11rb/pull/815 However, there `rustix` panics in one of the existing examples.

I tried to port this into a self-contained example. This requires something (e.g. an X11 server) to listen on the abstract unix socket at `/tmp/.X11-unix/X0`.

<details><summary>Cargo.toml and main.rs</summary>

```toml
[package]
name = "foo"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
rustix = { version = "0.37.19", default-features = false, features = ["std", "fs", "net"] }
```
```rust
use rustix::fd::OwnedFd;
use rustix::net::{
    connect_unix, recvmsg, sendmsg_noaddr, socket_with, AddressFamily, Protocol,
    RecvAncillaryBuffer, RecvFlags, SendFlags, SocketAddrUnix, SocketFlags, SocketType,
};

fn send_x11_setup_request(socket: &OwnedFd) -> Result<(), rustix::io::Errno> {
    let data = [b'l', 0, 11, 0, 0, 0, 0, 0, 0, 0, 0, 0];
    let result = sendmsg_noaddr(
        socket,
        &[std::io::IoSlice::new(&data)],
        &mut Default::default(),
        SendFlags::empty(),
    )?;
    assert_eq!(data.len(), result);
    Ok(())
}

fn do_receive(socket: &OwnedFd) -> Result<(), rustix::io::Errno> {
    let mut buffer = [0; 100];
    let mut cmsg = vec![0u8; rustix::cmsg_space!(ScmRights(16))];
    let mut iov = [std::io::IoSliceMut::new(&mut buffer)];
    let mut cmsg_buffer = RecvAncillaryBuffer::new(&mut cmsg);
    recvmsg(socket, &mut iov, &mut cmsg_buffer, RecvFlags::CMSG_CLOEXEC)?;
    Ok(())
}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let socket = socket_with(
        AddressFamily::UNIX,
        SocketType::STREAM,
        SocketFlags::CLOEXEC,
        Protocol::default(),
    )?;
    connect_unix(
        &socket,
[...]

PR #661, opened 2023-05-06
Fix decoding of abstract unix sockets

psychon:
This is my attempt at fixing #660.

I am not sure what will happen on the libc backend, but I expect things to still be broken there. Enabling the `use-libc` feature makes things not compile for me:
```
error[E0432]: unresolved import `libc_errno`
 --> src/backend/libc/io/errno.rs:7:5
  |
7 | use libc_errno::errno;
  |     ^^^^^^^^^^ use of undeclared crate or module `libc_errno`
```

### DIFF
--- a/tests/net/unix.rs
+++ b/tests/net/unix.rs
@@ -144,7 +144,7 @@ fn test_unix() {
 #[cfg(not(any(target_os = "redox", target_os = "wasi")))]
 fn do_test_unix_msg(addr: SocketAddrUnix) {
     use rustix::io::{IoSlice, IoSliceMut};
-    use rustix::net::{recvmsg, sendmsg_noaddr, RecvFlags, SendFlags};
+    use rustix::net::{recvmsg, sendmsg_noaddr, RecvFlags, SendFlags, SocketAddrAny};
 
     let server = {
         let connection_socket = socket(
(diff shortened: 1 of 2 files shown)


tests/net/unix.rs
```diff
@@ -208,21 +190,10 @@ fn test_unix_msg() {
...
     };
 
     let client = move || {
-        {
-            let (lock, cvar) = &*ready;
-            let mut started = lock.lock().unwrap();
-            while !*started {
-                started = cvar.wait(started).unwrap();
-            }
-        }
```
sunfishcode, line 223: I'm curious why the condition variable code is removed here. This is meant to ensure that the client waits for the server to be ready; is this no longer needed?

sunfishcode (commented):
Thanks for the fix! Overall this looks good.

tests/net/unix.rs
psychon: See the commit message for that commit: https://github.com/bytecodealliance/rustix/pull/661/commits/b5e97708f9f6f0ae9a57965ea3cb4779827fcd22

> test_unix_msg: Remove thread synchronisation
>
> This refactors the test_unix_msg() test a bit. No change in behaviour is
> intended.
>
> Previously, two threads were started in parallel. The server thread used
> a mutex and a condition variable to indicate that it set up its
> listening socket. The client thread waited for this signal before it
> attempted to connect.
> 
> This commit makes the code instead bind the server socket before
> starting the worker threads. That way, no synchronisation is necessary.

The actual reason why I did this was that at first I passed in the already-connected socket to the new helper function. That would have been a lot more complicated with this synchronisation mechanism. Only later did I notice that I only need to pass in the socket address and let the helper set up the socket itself.

Still, binding the socket before the "worker threads" are started seems like a simplification to me that gets rid of some unnecessary code. AFAIR there should not be any difference in which thread creates the socket and whether the socket is moved to another thread.

sunfishcode:
Thanks!

PR #674, opened 2023-05-19
Remove some rerun-if-env-changed lines from build.rs. (#576)

sunfishcode:
Now that we appear to have a known-working fix for #526, we can investigate relaxing the fix. If #526 reappears, we now have a known-working state we can quickly revert to.

Reports in #562 and #575 are that having rerun-if-env-changed=RUSTC and rerun-if-env-changed=RUSTC_WRAPPER are triggering spurious rebuilds in some situations, so remove these. In theory, cargo should already be aware of these environment variables. In practice, there were reports of build failures which had the appearance of using an older version of rustc with build.rs output from a newer version of rustc, and it wasn't clear what was happening. But, it's possible that #563 fixes the issue properly. So let's try removing these rerun-if-env-changed lines, and see if the problem resurfaces.

And, rerun-if-env-changed=TARGET is explicitly documented as being unnecessary [here].

[here]: https://doc.rust-lang.org/cargo/reference/build-scripts.html

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -232,13 +232,13 @@ jobs:
       QEMU_BUILD_VERSION: 7.0.0
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


sunfishcode:
This is a backport of #576 to the 0.36 branch

sunfishcode:
And #552 and an MSRV fix from #638.

PR #676, opened 2023-05-22
Fix panic on dropping `RecvAncillaryBuffer` after failed `recvmsg`

ids1024:
Calling drain on `RecvAncillaryBuffer`, including in its `Drop` implementation was failing, with an "attempted to subtract with overflow" error in `cvt_msg`.

If `recvmsg` returns `-1`, `msg_controllen` will not be updated by the call. So it had a non-zero value as passed into the function, despite there not being any control messages to parse.

### DIFF
--- a/src/backend/libc/net/msghdr.rs
+++ b/src/backend/libc/net/msghdr.rs
@@ -7,7 +7,7 @@ use super::super::c;
 use super::super::conv::{msg_control_len, msg_iov_len};
 use super::super::net::write_sockaddr::{encode_sockaddr_v4, encode_sockaddr_v6};
 
-use crate::io::{IoSlice, IoSliceMut};
+use crate::io::{self, IoSlice, IoSliceMut};
 use crate::net::{RecvAncillaryBuffer, SendAncillaryBuffer, SocketAddrV4, SocketAddrV6};
 use crate::utils::as_ptr;
 
@@ -19,8 +19,8 @@ pub(crate) fn with_recv_msghdr<R>(
     name: &mut MaybeUninit<c::sockaddr_storage>,
     iov: &mut [IoSliceMut<'_>],
     control: &mut RecvAncillaryBuffer<'_>,
-    f: impl FnOnce(&mut c::msghdr) -> R,
-) -> R {
+    f: impl FnOnce(&mut c::msghdr) -> io::Result<R>,
+) -> io::Result<R> {
     let namelen = size_of::<c::sockaddr_storage>() as c::socklen_t;
     let mut msghdr = {
         let mut h: c::msghdr = unsafe { zeroed() };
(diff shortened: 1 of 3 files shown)


src/backend/libc/net/msghdr.rs
```diff
@@ -36,8 +36,10 @@ pub(crate) fn with_recv_msghdr<R>(
     let res = f(&mut msghdr);
 
     // Reset the control length.
-    unsafe {
-        control.set_control_len(msghdr.msg_controllen.try_into().unwrap_or(usize::MAX));
+    if res.is_ok() {
+        unsafe {
+            control.set_control_len(msghdr.msg_controllen.try_into().unwrap_or(usize::MAX));
+        }
```
notgull, line 42: I'm not sure if the contents of the control buffer are defined on error. I think it would be safest to set the buffer's length to zero on error.

src/backend/linux_raw/net/msghdr.rs
```diff
@@ -49,8 +49,10 @@ pub(crate) fn with_recv_msghdr<R>(
     let res = f(&mut msghdr);
 
     // Reset the control length.
-    unsafe {
-        control.set_control_len(msghdr.msg_controllen.try_into().unwrap_or(usize::MAX));
+    if res.is_ok() {
+        unsafe {
+            control.set_control_len(msghdr.msg_controllen.try_into().unwrap_or(usize::MAX));
+        }
     }
```
notgull, line 56: Ditto here.

notgull (commented):
notgull code after spending five minutes in production:

robot fallx299

src/backend/libc/net/msghdr.rs
ids1024: `set_control_len` just sets the `length` (and `read`) field of the `RecvAncillaryBuffer`, which isn't part of the buffer that libc will be touching so it won't be changed by the `recvmsg`  call if this `set_control_len` line isn't run.

And for a newly allocated `RecvAncillaryBuffer`, it will already be `0`. So there's no need to set it, in that case.

If the same `RecvAncillaryBuffer` is re-used in multiple calls, I'm not sure how that works currently. Should check that doesn't leak file descriptors or anything.

notgull: We should probably be clearing out all of the inner data before we do anything with it.

ids1024: Added a commit that clears the `RecvAncillaryBuffer` at the start of the recv operation.

sunfishcode:
@notgull Do you have any further comments here?

sunfishcode:
@notgull Is this ready to merge?

notgull (approved):
Looks good to me!

sunfishcode:
Thanks!

PR #681, opened 2023-06-03
Add pidfd_getfd syscall

moshevds:
Adding the `pidfd_getfd` syscall may require some discussion, but after doing some research, I think the approach used in this PR makes some sense. So, I sent this as a PR directly instead of creating an issue first.

The 3 main things that I think require some thought, are:
 - Is this even a desired feature to add to rustix?
 - How does pidfd_getfd measure up in the I/O safety concept?
 - Are the correct types used? (Especially for `targetfd`)

In this PR, I marked the function unsafe because it _can_ be used as a kind of `dup` from a RawFd to OwnedFd, when an application uses it on itself. That would still be memory safe, but violates I/O safety as I understand from the RFC.
The argument against marking this unsafe would be that using this does not impact the safety of the calling application directly, and the safety of other processes is outside of the bounds of the concepts of memory and I/O safety anyway. (Even when the target process could end up being the calling process in practice?) I'm curious what the thoughts are about this, and what safety considerations I may have missed.

For the `targetfd` arg: I think there is no other code that uses RawFd/AsRawFd for values that are not potential file descriptors of the current process. At least to me, it feels like a bit of a stretch to use it for `targetfd`. On the other hand, using `c_int` directly seems worse. Any thoughts on this?

And the flags: The syscall currently does not take any flags, but might in the future. I added an empty enum to represent this. Is this the correct approach?

### DIFF
--- a/src/process/mod.rs
+++ b/src/process/mod.rs
@@ -14,6 +14,8 @@ mod kill;
 mod membarrier;
 #[cfg(target_os = "linux")]
 mod pidfd;
+#[cfg(target_os = "linux")]
+mod pidfd_getfd;
 #[cfg(linux_kernel)]
 mod prctl;
 #[cfg(not(any(target_os = "fuchsia", target_os = "wasi")))] // WASI doesn't have [gs]etpriority.
@@ -44,6 +46,8 @@ pub use kill::*;
 pub use membarrier::*;
 #[cfg(target_os = "linux")]
 pub use pidfd::*;
+#[cfg(target_os = "linux")]
+pub use pidfd_getfd::*;
 #[cfg(linux_kernel)]
 pub use prctl::*;
 #[cfg(not(any(target_os = "fuchsia", target_os = "wasi")))]
--- a/src/process/pidfd_getfd.rs
+++ b/src/process/pidfd_getfd.rs
@@ -1,0 +1,46 @@
+#![allow(unsafe_code)]
+use crate::fd::OwnedFd;
+use crate::{backend, io};
+use backend::fd::{AsFd, RawFd};
+
+/// Raw file descriptor in another process.
+///
+/// A distinct type alias is used here to inform the user that normal file descriptors from the
+/// calling process should not be used. The provided file descriptor is used by the kernel as the
+/// index into the file descriptor table of an entirely different process.
+pub type ForeignRawFd = RawFd;
+
+bitflags::bitflags! {
+    /// All flags are reserved for future use.
+    pub struct PidfdGetfdFlags: backend::c::c_uint {}
+}
+
+/// `syscall(SYS_pidfd_getfd, pidfd, flags)`—Obtain a duplicate of another
+/// process's file descriptor.
+///
+/// # References
+///  - [Linux]
+///
+/// # Warning
+///
+/// This function is generally safe for the calling process, but it can impact the target process
+/// in unexpected ways. If you want to ensure that Rust I/O safety assumptions continue to hold in
+/// the target process, then the target process must have communicated the file description number
+/// to the calling process from a value of a type that implements AsRawFd, and the target process
+/// must not drop that value until after the calling process has returned from `pidfd_getfd`.
+///
+/// When `pidfd_getfd` is used to debug the target, or the target is not a Rust aplication, or
+/// `pidfd_getfd` is used in any other way, then extra care should be taken to avoid unexpected
+/// behaviour or crashes.
+///
+/// For further details, see the references above.
+///
+/// [Linux]: https://man7.org/linux/man-pages/man2/pidfd_getfd.2.html
+#[inline]
+pub fn pidfd_getfd<Fd: AsFd>(
+    pidfd: Fd,
+    targetfd: ForeignRawFd,
+    flags: PidfdGetfdFlags,
+) -> io::Result<OwnedFd> {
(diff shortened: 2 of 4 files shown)


sunfishcode:
Thanks! I didn't know about `pidfd_getfd` before. It's surprising to me that this exists. But indeed there it is :wink: .

> The 3 main things that I think require some thought, are:
> 
>     * Is this even a desired feature to add to rustix?

Yes. Rustix in general tries not to second-guess OS features (unless they're *really* problematic like `vfork`).

>     * How does pidfd_getfd measure up in the I/O safety concept?

Great question. As you say, making this safe and using `RawFd` could make this an I/O safety bypass. But, my current thinking is, it looks like `pidfd_getfd` requires `PTRACE_MODE_ATTACH_REALCREDS` permissions over the target process, and that puts it in the category of `ptrace`-like functionality, or debugger-like functionality. So from a safety point of view, this is equivalent to attaching a debugger to the target. Rust programs can already attach debuggers to themselves and corrupt themselves, even using only safe Rust code. Consequently, it's ok for `pidfd_getfd` to be a safe function too.

This interpretation of classifing features as "debugger-like" vs. not isn't formalized anywhere, and it could evolve, but so far, it's been a rule that's been fairly straightforward to apply, and it has held up to a fair amount of informal reasoning. Rust's safety as we know it today isn't defending against malicious code; it's just defending against bugs. This suggests that we should document the hazards of `pidfd_getfd`, but we don't need to make it `unsafe`.

> For the `targetfd` arg: I think there is no other code that uses RawFd/AsRawFd for values that are not potential file descriptors of the current process. At least to me, it feels like a bit of a stretch to use it for `targetfd`. On the other hand, using `c_int` directly seems worse. Any thoughts on this?

I think it's ok to use `RawFd` for this, because it is just an integer type and can hold all the needed values. But, to better communicate it in the API, what if we defined a new type alias, like `ForeignRawFd` or some name like that, which is an alias for `RawFd`, and used it here?

> And the flags: The syscall currently does not take any flags, but might in the future. I added an empty enum to represent this. Is this the correct approach?

Yes, this looks good.

src/process/mod.rs
```diff
@@ -14,6 +14,8 @@ mod kill;
 mod membarrier;
 #[cfg(target_os = "linux")]
 mod pidfd;
+#[cfg(linux_raw)]
+mod pidfd_getfd;
```
sunfishcode, line 18: Rustix aims to avoid having functions available on `linux_raw` and not `libc`. `pidfd_getfd` can be supported in the `libc` backend using `libc::syscall`.

moshevds:
Thank you for the in-depth response. I have changed the PR as suggested. I decided to added the ForeignRawFd alias, I think it makes sense. I also changed the argument to use the alias directly, without being generic on `AsRawFd` (or a new `AsForeignRawFd` trait). I think removing the generic is better because the previous version enabled things like
```rust
pidfd_getfdpid(fd, File::open("foo.txt")?, flags)
```
which seems very wrong to me, and I cannot think of other use-cases where the generic actually makes sense for `pidfd_getfd`.

Maybe the alias should go to another place entirely? In that case I'm unsure whether the alias is really worth it, so if you/others think just using RawFd (without the generic) is better, then I can see that point as well.

src/process/pidfd_getfd.rs
```diff
@@ -0,0 +1,42 @@
...
+///
+/// # Safety
+///
+/// This function is generally safe for the calling process, but it can impact the target process
+/// in unexpected ways. If you want to ensure that rust I/O safety assumptions continue to hold in
+/// the target process, then the target process must have communicated the file description number
+/// to the calling process from a value of a type that implements AsRawFd, and the target process
+/// must not drop that value until after the calling process has returned from `pidfd_getfd`.
+///
+/// When `pidfd_getfd` is used to debug the target, or the target is not a rust aplication, or
```
sunfishcode, line 28: The word "Rust" should be capitalized within this comment.

src/process/pidfd_getfd.rs
```diff
@@ -0,0 +1,42 @@
+#![allow(unsafe_code)]
+use crate::fd::OwnedFd;
+use crate::{backend, io};
+use backend::fd::{AsFd, RawFd};
+
+/// Raw file descriptor in another process.
```
sunfishcode, line 6: It'd be good to add a note here that users should take extra care to avoid using host file descriptors here, since they use the same underlying type.

src/process/pidfd_getfd.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub struct PidfdGetfdFlags: backend::c::c_uint {}
+}
+
+/// `syscall(SYS_pidfd_getfd, pidfd, flags)`—Obtain a duplicate of another
+/// process's file descriptor.
+///
+/// # References
+///  - [Linux]
+///
+/// # Safety
```
sunfishcode, line 20: Strictly speaking, this is no longer a `Safety` comment, because the function is not `unsafe`. Could you change this to say `# Warning`, which rustix uses in a few other places for similar things?

sunfishcode (commented):
Overall looks good, just a few comments about comments:

moshevds:
I added the suggestions, and also fixed a typo ("resuerved") that I noticed

sunfishcode:
Looks good!

PR #686, opened 2023-06-07
Reorganise std polyfill

niluxv:
Rustix contains some std polyfill code for the `no_std` build. I tried reorganising it a bit, and I feel like the following structure makes sense:
* Add a new crate local top level module `polyfill`
* On `no_std` build use `polyfill/nostd/mod.rs`, otherwise `polyfill/std/mod.rs`
* Relocate all polyfill code to `polyfill/nostd/`
* Use the same paths as the standard library uses
* `polyfill/std/mod.rs` replicates the same structure, but by importing the required items from `std`
* The rest of rustix can now use `crate::polyfill::path::to::item` where it would otherwise use `std::path::to::item` regardless of whether we have `std` or not

How do you feel? If you like the idea I'll add some more documentation and check whether there is more code to migrate.


niluxv:
CI failures look unrelated.

sunfishcode:
I like it! 

I've now fixed the CI failures on the 0.38-planning branch. Unfortunately I rewrote history on that branch, so you probably won't be able to do a simple rebase, but if you `git reset --hard 0.38-planning` and then cherry-pick your original patch on top of that, that would probably just work.

I'm a little uncomfortable with how the name `polyfill` makes it look like we're always using a polyfill, when in fact we're only doing so in no_std builds. And the name `nostd` will trip me up because the feature name is `no_std` and I'll forget sometimes which contexts need the underscore and which don't. What would you think of renaming `polyfill` to `std` and `nostd` to `polyfill`? Is that too cute? Alternatively, what would you think of renaming `pollyfill` to `maybe_polyfill` and `nostd` to `no_std`?

niluxv:
> Unfortunately I rewrote history on that branch, so you probably won't be able to do a simple rebase

Fortunately there is `git rebase -i` :)

Regarding naming, that is of course open to bikeshed. Sometimes naming is  the hardest part of programming. I do feel `maybe_polyfill` is too long though. I also considered changing `polyfill` to `std`, but then it sounds like it would contain the entirety of `std`... I would like to mention that even the `std` implementation does a tiny bit of polyfilling: it exports `std::os::unix/wasi::io` as if it were `std::os::fd` (which we cannot use due to MSRV requirements).

I agree with the `nostd` -> `no_std`. Please let me know what naming you decide on.

It seems there is one piece of polyfill I hadn't covered yet: `io/io_slice.rs` in the backends (it is duplicated across backends). The only difference between the backends AFAICT is [a cast](https://github.com/bytecodealliance/rustix/blob/61e653b1df7de0d74766de92d70dd3cffff990b1/src/backend/linux_raw/io/io_slice.rs#LL80C35-L80C53). Still, having backend aware code in polyfill is quite ugly.

sunfishcode:
I think I like `maybe_polyfill` and `no_std`. `maybe_polyfill` is longer than what one might want to type a lot, but it shouldn't appear many places in the code. It isn't a public module, and inside rustix, most places in the code reference the re-exported names from `rustix::net` or `rustix::fd`, so `maybe_polyfill` only ever needs to appear in a handful of places.

niluxv:
Ok, renamed, added some documentation, rebased and clarified the licensing status of the `std` derived code (it is dual licensed Apache 2.0 or MIT since that is the license `std` uses, in contrast to the rest of `rustix`, which is triple licensed).

PR #690, opened 2023-06-11
Move all std polyfill code to a single module/directory

niluxv:
Same as #686, but now against `main` (and rebased again).


sunfishcode (commented):
Thanks for this patch, and thanks for rebasing it (again!).

PR #692, opened 2023-06-15
Support no-std on windows

### LINKED ISSUE #682: broken no_std support on windows
Hi, no_std completely broken on rustix (especially "net" feature), compile with `default-features=false` and `#![no_std]` in _main.rs_ causes many errors like 

```
src\backend\libc\io_lifetimes.rs:21:28
   |
21 |     fn as_raw_fd(&self) -> RawFd;
   |                            ^^^^^ not found in this scope
   |
help: consider importing this type alias
   |
6  + use crate::io::fd::RawFd;
```

not found function:
```
src\io\fd\owned.rs:95:29
   |
95 |         let fd = crate::io::fcntl_dupfd_cloexec(self, 0)?;
   |                             ^^^^^^^^^^^^^^^^^^^ not found in `crate::io`

   ```

also, there's a std import that uncovered under `#[cfg(feature = "std")]`:

```
src\backend\libc\io\poll_fd.rs:9:5
  |
9 | use std::fmt;
  |     ^^^ use of undeclared crate or module `std`
  ```
  
  
  Cargo.toml import line:
  ```
  rustix = {version = "0.37.19", features=["net","io-lifetimes","use-libc"], default-features=false}
  ```

niluxv:
Fixes #682.

The new `maybe_polyfill` module made this very easy to do in my opinion.

### DIFF
--- a/src/backend/libc/event/poll_fd.rs
+++ b/src/backend/libc/event/poll_fd.rs
@@ -6,7 +6,7 @@ use core::marker::PhantomData;
 #[cfg(windows)]
 use {
     crate::backend::fd::{AsSocket, RawFd},
-    std::fmt,
+    core::fmt,
 };
 
 bitflags! {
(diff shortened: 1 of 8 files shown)


sunfishcode:
Cool, and I agree, the `maybe_polyfill` makes this really nice.

PR #693, opened 2023-06-16
Improve strict-provenance compatibility in the `epoll` API.

sunfishcode:
Change the user-data field of epoll's `Event` from a bare `u64` to a `union` which can be either a `u64` or a `*mut c_void` to allowe users to store pointers in it that preserve strict provenance.

Also, rename `epoll::epoll_add` and similar to just `epoll::add` and similar, for tidiness.


sunfishcode:
There's a FIXME here; once #615 lands, I'll replace the FIXME comment with tests that `Event`'s layout matches `epoll_event`.

PR #694, opened 2023-06-17
Clean up redundant Pid check

oxalica:
`Pid::from_raw_unchecked` already contains a positivity `debug_assert`.


sunfishcode:
Cool, thanks for catching that, and for the PR!

PR #721, opened 2023-07-03
Remove naked_functions feature usage for x86

nbdd0121:
Given that this function has no generics, it could easily use `global_asm!` instead.

This would remove the outline asm and associated pre-built binaries for x86.


nbdd0121:
It's weird that the test of `assert_eq!` is failing with two sides being equal... Seems that UB is triggered somewhere but it is unlikely to be in the code that I touched.

nbdd0121:
Not sure what happened? CI failed, this shows as merged, but not in the main branch?

sunfishcode:
Oops, I started debugging the test failure and accidentally pushed this to main, which I didn't realize auto-closed this, and then reset main to fix the mistake. Would you mind re-opening this, or opening a new PR with this patch? I think the patch is a good idea, using `global_asm` now that our MSRV is bumped to 1.63 is much nicer than shipping around a prebuilt archive.

nbdd0121:
There is no way to reopen a merged PR on GitHub. Will re-open a new PR.

PR #722, opened 2023-07-04
Remove naked_functions feature usage for x86

nbdd0121:
Given that this function has no generics, it could easily use `global_asm!` instead.


nbdd0121:
I haven't been able to reproduce the issue locally. Probably need to steal the version out from CI to figure out what happened.

src/backend/linux_raw/vdso_wrappers.rs
```diff
@@ -310,21 +310,31 @@ unsafe fn _rustix_clock_gettime_via_syscall(
...
-    asm!("int $$0x80", "ret", options(noreturn))
-}
-
-// The outline version of the `rustix_int_0x80` above.
-#[cfg(all(not(asm), target_arch = "x86"))]
+#[cfg(target_arch = "x86")]
 extern "C" {
+    /// A symbol pointing to an `int 0x80` instruction. This “function” is only
+    /// called from assembly, and only with the x86 syscall calling convention.
+    /// so its signature here is not its true signature.
```
sunfishcode, line 317: Could you add a comment here mentioning that this could use the `naked_functions` feature should it be stabilized?

sunfishcode (commented):
Thanks! Using global_asm is much nicer than shipping a prebuilt archive.

The CI failure turned out to be an existing bug, which is now fixed in #723.

PR #750, opened 2023-07-18
Detect MIPS R6

chenx97:
This PR aims to add support for recently [proposed](https://github.com/rust-lang/compiler-team/issues/632) architectures, which should be much easier to do now thanks to the removal of outline assembly.

MIPS R6 shares syscall conventions with MIPS, so the inline assembly files may only differ in their comments.


sunfishcode:
linux-raw-sys 0.4.5 is now released with https://github.com/sunfishcode/linux-raw-sys/pull/73 in it, adding MIPS R6 support. Is this rustix PR ready for review now?

chenx97:
It should be ready now.

PR #751, opened 2023-07-18
Linux on loongarch64 has no open syscall so use openat

xiangzhai:
Hi,

Linux on loongarch64 has no `open` syscall neither, so use `openat` instead.

Passed cargo test.

Thanks,
Leslie Zhai


sunfishcode:
Thanks!

Issue #758, opened 2023-07-29
Typographical error and missing flags


ImBearChild:
`CLONE_NEWUTS` is missing from `rustix::thread::UnshareFlags` . And `rustix::thread::UnshareFlags::NWCGROUP` seems to be a typographical error of N***E***WCGROUP.  `rustix::fs::MountFlags` also have some items missing, compared to `nix::mount::MsFlags`.  I have found some of those flags in `rustix::fs::MountPropagationFlags`. That's a bit of counterintuitive.

Can I open a pull request to fix them ?

ImBearChild:
For example, I want to port following c code to rust using rustix:

```c
  if (mount (NULL, "/", NULL, MS_SILENT | MS_SLAVE | MS_REC, NULL) < 0)
    die_with_error ("Failed to make / slave");
```

I can't use change_mount(), beacuse there's no `MountPropagationFlags::SILENT`.

```rust
        change_mount(
            "/",
            MountPropagationFlags::SILENT | MountPropagationFlags::SLAVE | MountPropagationFlags::REC,
        ).unwrap();
```

Also, I can't use mount(), because there's no `MountFlags::SLAVE`.

```rust
        mount(
            "", "/", "",
            MountFlags::SILENT | MountFlags::SLAVE | MountFlags::REC, 
            "",
        ).unwrap();
```

Maybe `MountPropagationFlags::SILENT` should be added to `MountPropagationFlags`, because it "Suppress the display of certain (printk()) warning messages in the kernel log." , which should be valid for both `mount()` and `change_mount()`

sunfishcode:
These fixes are now published in rustix 0.38.6.

PR #759, opened 2023-07-29
Fix missing UTS flags and typo error for unshare

ImBearChild:
Fix missing UTS flags and typo error in #758.


sunfishcode:
Thanks!

PR #760, opened 2023-07-29
Add MS_SILENT for MountPropagationFlags

ImBearChild:
Fix MS_SILENT for #758


sunfishcode:
Could you also add a definition to src/backend/libc/fs/types.rs? It can also use the linux_raw_sys value.

ImBearChild:
> Could you also add a definition to src/backend/libc/fs/types.rs? It can also use the linux_raw_sys value.

ok, I have pushed a now commit to this branch.

sunfishcode:
Thanks!

PR #763, opened 2023-07-31
Factor out a new top-level `mount` module.

sunfishcode:
Create `rustix::mount`, and move the mount-related and fsopen-related functions into it, adding deprecated aliases in `rustix::fs` for compatibility.

This also renames several functions which are all specialized forms of the `mount` system call, so I renamed them to have `mount` at the beginning. It also fixes the collision with the real `move_mount` system call.

 - `remount` to `mount_remount`
 - `bind_mount` to `mount_bind`
 - `recursive_bind_mount` to `mount_recursive_bind`
 - `change_mount` to `mount_change`
 - `move_mount` to `mount_move`

And, it adds documentation links to a [new repository] I set up to host formatted versions of the draft manual pages that were posted.

[new repository]: https://github.com/sunfishcode/linux-mount-api-documentation

@yujincheng08 @SUPERCILEX


sunfishcode:
Thanks! The documentation calls it a "recursive bind mount" so the answer to "what kind of mount is it?" is "a recursive bind one", so I went with `mount_recursive_bind` for now, though I also made a note to revisit this.

PR #769, opened 2023-08-03
Bindings of eventfd for the ESP-IDF framework

ivmarkov:
The `eventfd` sys call [is available on the ESP IDF framework](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/vfs.html#event-fds), although with some restrictions (no support for `EFD_CLOEXEC`, `EFD_NONBLOCK` and `EFD_SEMAPHORE` and non-blocking mode enabled by default).

Given that the ESP IDF framework is already supported in `rustix` we would like to extend the support to include this syscall as well.

Background: support for `eventfd` is key for [upstreaming ESP IDF support in the `smol/polling` crate](https://github.com/esp-rs-compat/polling/blob/espidf/src/poll.rs#L105).


sunfishcode:
Looks good, thanks!

PR #772, opened 2023-08-06
Add data() method to kqueue events

notgull:
This allows us to get the raw data field out, which is useful for reading receipt events.

This lets us fully remove libc from polling, see this line: https://github.com/smol-rs/polling/blob/6d13def8abd73da96d1e73090f490ae0ad36d6bd/src/kqueue.rs#L132

### DIFF
--- a/src/event/kqueue.rs
+++ b/src/event/kqueue.rs
@@ -100,8 +100,8 @@ impl Event {
     }
 
     /// Get the raw data for this event.
-    pub fn data(&self) -> isize {
-        // On openbsd, data is an i64 and not an isize
+    pub fn data(&self) -> i64 {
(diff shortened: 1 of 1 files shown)


src/event/kqueue.rs
```diff
@@ -99,6 +99,12 @@ impl Event {
         self.inner.udata as _
     }
 
+    /// Get the raw data for this event.
+    pub fn data(&self) -> isize {
+        // On openbsd, data is an i64 and not an isize
```
sunfishcode, line 104: It looks like netbsd and newer versions of freebsd have this too. If this is sometimes an i64, should we cast to i64 here instead of isize?

PR #777, opened 2023-08-09
Fix AIX build with all-apis feature

ecnelises:
- Fix networking API support
- Make stat family function signature consistent


src/backend/libc/c.rs
```diff
@@ -133,7 +130,7 @@ pub(super) use libc::{
         target_arch = "mips64r6"
     )
 )))]
-#[cfg(any(linux_like, target_os = "aix"))]
+#[cfg(linux_like)]
 pub(super) use libc::{lstat64 as lstat, stat64 as stat};
```
sunfishcode, line 134: According to Rust's libc bindings, [AIX does have blksize64_t](https://github.com/rust-lang/libc/blob/main/src/unix/aix/mod.rs#L27) and all the other `*64*` things. At least at some point, a powerpc-ibm-aix target existed, where `c_long` was presumably 32-bit and `blksize_t` was 32-bit. It's rustix's intention to always use 64-bit file offsets, so unless AIX can guarantee that `c_long` and `blksize_t` and all the rest will always be 64-bit, I'd prefer to continue using the `*64*` things.

src/backend/libc/net/send_recv.rs
```diff
@@ -22,7 +23,10 @@ bitflags! {
         /// `MSG_DONTROUTE`
         const DONTROUTE = bitcast!(c::MSG_DONTROUTE);
         /// `MSG_DONTWAIT`
-        #[cfg(not(windows))]
+        #[cfg(not(any(
+            windows,
+            target_os = "aix",
+        )))]
         const DONTWAIT = bitcast!(c::MSG_DONTWAIT);
```
sunfishcode, line 30: On AIX, would it make sense to define `DONTWAIT` with the value of `MSG_NONBLOCK`?

sunfishcode:
Thanks!

PR #781, opened 2023-08-09
Construct non-zero types safely at compile-time

### LINKED ISSUE #776: Non-zero constants could be constructed using safe code
Code like the following:

https://github.com/bytecodealliance/rustix/blob/a19658282a26b3ac6ea581b5a69f570774a37dd5/src/pid.rs#L20-L23

Could be replaced by safe code using const panic:

```rust
match NonZeroI32::new(1) {
    Some(nz) => nz,
    None => panic!("unreachable: 1 is non-zero"), // valid in a const context
}
```

joshlf:
Closes #776

### DIFF
--- a/src/net/types.rs
+++ b/src/net/types.rs
@@ -1,5 +1,3 @@
-#![allow(unsafe_code)]
-
 use crate::backend::c;
 use bitflags::bitflags;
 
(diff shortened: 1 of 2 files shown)


sunfishcode:
Cool, thanks!

PR #783, opened 2023-08-10
Remove some unsafe code

joshlf:


src/maybe_polyfill/no_std/net/ip_addr.rs
```diff
@@ -1275,8 +1272,7 @@ impl Ipv6Addr {
     #[inline]
     pub const fn segments(&self) -> [u16; 8] {
         // All elements in `self.octets` must be big endian.
-        // SAFETY: `[u8; 16]` is always safe to transmute to `[u16; 8]`.
-        let [a, b, c, d, e, f, g, h] = unsafe { transmute::<_, [u16; 8]>(self.octets) };
```
sunfishcode, line 1279: Code in the src/maybe_polyfill/no_std directory is derived from Rust standard library code, and modifying it makes it more work to keep in sync with upstream. These two unsafe blocks are identical to unsafe blocks in the standard library, and they're easy to verify, so it seems most practical to leave them as-is here.

src/bitcast.rs
```diff
@@ -1,32 +1,8 @@
...
-            #[allow(unsafe_code, unused_unsafe, clippy::useless_transmute)]
-            unsafe {
-                ::core::mem::transmute($x)
-            }
-        } else {
-            // Do the conversion.
-            $x as _
-        }
-    }};
-}
```
sunfishcode, line 23: It's unclear how this change improves theoretical safety, since this `bitcast` macro does not execute unsafe code, while zerocopy::transmute [does execute unsafe code](https://docs.rs/zerocopy/latest/src/zerocopy/lib.rs.html#1409).

sunfishcode (commented):
Thanks for the PR! Would you mind posting a comment giving some background for this change?

In particular, is this change worth adding a dependency for?

src/bitcast.rs
joshlf: Ah I misread the macro; you're right that zerocopy doesn't add anything unsafe-wise; removed.

joshlf:
With those two removals, there's very little left; happy to just close this if you'd prefer!

I'd also be happy to add zerocopy trait impls to linux-raw-sys; from my skim of the code, it looks like that'd unlock a lot more than what I was able to do here.

sunfishcode:
Rustix doesn't use very many `transmute`s or `zeroed`s, and in looking into it just now, I found some opportunities to factor out and remove some, which I've now added to #785. So with what I know right now, my guess is that there aren't enough remaining opportunities to make `zerocopy` worthwhile here.

joshlf:
Yeah, that sounds right; I'll close this!

PR #785, opened 2023-08-10
Move some `transmute`s out of the main code and into `#[test]`s.

sunfishcode:
In a few places, `transmute` was being used to assert that two types have the same size. Remove these `transmute`s and add corresponding `assert_eq_size` tests instead.


sunfishcode:
I've now also add `default_array` and `default_union` utilities for zero-initializing arrays and unions, and factored out the code for zero-initializing `msghdr`s.

Issue #683, opened 2023-06-04
Support arbitrary ioctls


notgull:
For several drivers, they expose a unique, sometimes proprietary `ioctl` interface. Since it is unfeasible for `rustix` to expose implementations for all of these `ioctl`s, there should be some API that allows users to call `ioctl` with arbitrary arguments.

PR #788, opened 2023-08-12
Set up a unified system for handling ioctl calls

notgull:
A pain point in rustix is not being able to do custom ioctls; see https://github.com/bytecodealliance/rustix/issues/683
for rationale.

This commit adds an `Ioctl` trait which can be used to define arbitrary
ioctls. It is unsafe to implement, but once implemented can be used
safely. It can then be passed into an `ioctl` method with a file
descriptor in order to call an ioctl.

This commit also reimplements some of the functions used in this crate
in terms of this new system.

Closes https://github.com/bytecodealliance/rustix/issues/683

### DIFF
--- a/src/fs/ioctl.rs
+++ b/src/fs/ioctl.rs
@@ -2,8 +2,10 @@
 
 #[cfg(linux_kernel)]
 use {
-    crate::fd::AsFd,
-    crate::{backend, io},
+    crate::fd::{AsFd, AsRawFd, BorrowedFd},
+    crate::{backend, io, ioctl},
+    backend::c,
+    core::mem::MaybeUninit,
 };
 
 /// `ioctl(fd, BLKSSZGET)`—Returns the logical block size of a block device.
@@ -15,7 +17,7 @@ use {
 #[inline]
 #[doc(alias = "BLKSSZGET")]
 pub fn ioctl_blksszget<Fd: AsFd>(fd: Fd) -> io::Result<u32> {
-    backend::fs::syscalls::ioctl_blksszget(fd.as_fd())
+    ioctl::ioctl(fd, Blksszget(MaybeUninit::uninit()))
 }
 
 /// `ioctl(fd, BLKPBSZGET)`—Returns the physical block size of a block device.
@@ -23,7 +25,7 @@ pub fn ioctl_blksszget<Fd: AsFd>(fd: Fd) -> io::Result<u32> {
 #[inline]
 #[doc(alias = "BLKPBSZGET")]
 pub fn ioctl_blkpbszget<Fd: AsFd>(fd: Fd) -> io::Result<u32> {
-    backend::fs::syscalls::ioctl_blkpbszget(fd.as_fd())
+    ioctl::ioctl(fd, Blkpbszget(MaybeUninit::uninit()))
 }
 
 /// `ioctl(fd, FICLONE, src_fd)`—Share data between open files.
@@ -38,7 +40,7 @@ pub fn ioctl_blkpbszget<Fd: AsFd>(fd: Fd) -> io::Result<u32> {
 #[inline]
 #[doc(alias = "FICLONE")]
 pub fn ioctl_ficlone<Fd: AsFd, SrcFd: AsFd>(fd: Fd, src_fd: SrcFd) -> io::Result<()> {
-    backend::fs::syscalls::ioctl_ficlone(fd.as_fd(), src_fd.as_fd())
+    ioctl::ioctl(fd, Ficlone(src_fd.as_fd()))
 }
 
 /// `ioctl(fd, EXT4_IOC_RESIZE_FS, blocks)`—Resize ext4 filesystem on fd.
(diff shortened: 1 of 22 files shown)


src/io/ioctl.rs
```diff
@@ -66,5 +69,71 @@ pub fn ioctl_fionbio<Fd: AsFd>(fd: Fd, value: bool) -> io::Result<()> {
...
+        (&mut self.0 as *mut c::c_int).cast()
+    }
+
+    unsafe fn output_from_ptr(
+        _: ioctl::IoctlOutput,
+        _: *mut c::c_void,
+    ) -> io::Result<Self::Output> {
+        Ok(())
+    }
+}
```
sunfishcode, line 117: A lot of ioctls are simple "setters" like this, and they'll all return `()`, all be mutating, they'll all do `(&mut self.0 as *mut T)` in their `as_ptr`, and `Ok(())` in their `output_from_ptr`.

Similarly, a lot of ioctls are "getters" which will also have a lot in common with each other.

"getters" and "setters" are common enough that I wonder if it's possible to define some more concise way to define them so that we don't have to repeat all this code so much.

sunfishcode (commented):
Overall I like this idea. One quick thought:

tests/fs/ioctl.rs
```diff
@@ -19,6 +19,7 @@ fn test_ioctl_ficlone() {
     // that don't support this.
     match rustix::fs::ioctl_ficlone(&dest, &src) {
         Ok(()) | Err(io::Errno::OPNOTSUPP) => (),
+        Err(e) if e == io::Errno::from_raw_os_error(0x12) => (),
```
notgull, line 22: Unrelated to this PR; I keep getting `EXDEV` for this test and I don't know why.

sunfishcode: That looks like a bug in the test. `src` is "Cargo.toml" and `dest` is a `tempfile()` so if you have a tmpfs on /tmp or so, that does sound like it could get an `EXDEV`. We should change that code to create two files in `dir` and then call `ioctl_ficlone` on those files.

src/io/ioctl.rs
```diff
@@ -42,7 +45,12 @@ pub fn ioctl_fioclex<Fd: AsFd>(fd: Fd) -> io::Result<()> {
 #[inline]
 #[doc(alias = "FIONBIO")]
 pub fn ioctl_fionbio<Fd: AsFd>(fd: Fd, value: bool) -> io::Result<()> {
-    backend::io::syscalls::ioctl_fionbio(fd.as_fd(), value)
+    // SAFETY: FIONBIO is a pointer setter opcode.
+    #[allow(unsafe_code)]
```
sunfishcode, line 49: Could you move these `#[allow(unsafe_code)]`s to a `#![allow(unsafe_code)]` at the top of the file?

The goal isn't to add extra ceremony to every `unsafe` block, but to try to encourage factoring out `unsafe` blocks from files that don't otherwise have any `unsafe` in them.

src/io/ioctl.rs
```diff
@@ -42,7 +45,12 @@ pub fn ioctl_fioclex<Fd: AsFd>(fd: Fd) -> io::Result<()> {
...
+    let ctl = unsafe {
+        ioctl::PtrSetter::<ioctl::BadOpcode<{ c::FIONBIO }>, c::c_int>::new(c::c_int::from(value))
```
sunfishcode, line 51: Instead of `PtrSetter`, could it be called just `Setter`? From the perspective of code defining an ioctl like this, the fact that it uses a pointer internally isn't what's most important.

Why is this using a "bad" opcode? `FIONBIO` seems like a perfectly good opcode to me ;-).

Instead of naming `c::c_int` twice, could `c::c_int::from(value)` be `value.into()`?

src/ioctl/patterns.rs
```diff
@@ -0,0 +1,198 @@
...
+    fn as_ptr(&mut self) -> *mut c::c_void {
+        self.output.as_mut_ptr().cast()
+    }
+
+    unsafe fn output_from_ptr(_: IoctlOutput, ptr: *mut c::c_void) -> Result<Self::Output> {
+        Ok(ptr.cast::<Output>().read())
+    }
+}
+
+/// Implements the pattern for `ioctl`s where a pointer argument is given to the `ioctl`.
```
sunfishcode, line 102: It's worth mentioning here that "setter" opcodes must be readonly.

src/ioctl/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+#[cfg(linux_kernel)]
+use linux as platform;
+
+#[cfg(any(apple, bsd))]
+use bsd as platform;
+
+/// Perform an `ioctl` call.
+///
+/// `ioctl` was originally intended to act as a way of modifying the behavior of files, but has
+/// since been adopted as a general purpose system call for making calls into the kernel. In
```
sunfishcode, line 41: My sense is that this context about `ioctl` originally being intended for modifying files isn't needed here. These days, `ioctl` just is a general-purpose system call.

src/fs/ioctl.rs
```diff
@@ -15,15 +19,21 @@ use {
 #[inline]
 #[doc(alias = "BLKSSZGET")]
 pub fn ioctl_blksszget<Fd: AsFd>(fd: Fd) -> io::Result<u32> {
-    backend::fs::syscalls::ioctl_blksszget(fd.as_fd())
+    // SAFETY: BLZSSZGET is a getter opcode that gets a u32.
+    #[allow(unsafe_code)]
+    let ctl = unsafe { ioctl::Getter::<ioctl::BadOpcode<{ c::BLKSSZGET }>, c::c_uint>::new() };
+    ioctl::ioctl(fd, ctl)
```
sunfishcode, line 25: This approach of factoring out the `unsafe` from the actual `ioctl` call is really nice! However I have one potential concern: could there be a problem if a kernel module gets unloaded and another gets loaded while the program is running?

If `unsafe` is on the `ioctl` call, we're effectively protected against that, because if you have an open `fd` that ought to prevent the module from being unloaded. But if we separate the `unsafe` from the `ioctl` call, then in theory there could be a point where all fds referring to a module are dropped, the module could be unloaded, another could be loaded, new fds could be created, and the same `ioctl::Getter` could be used without `unsafe`, but having a different meaning in the new module.

I'm not super familiar with either ioctls or kernel modules, so I may be missing something though.

sunfishcode (commented):
Cool! Here are some comments from an initial readthrough:

src/fs/ioctl.rs
sunfishcode: Ah, and a potential broader version of this concern is that with driver-specific ioctls, the safety of calling a particular ioctl depends on the dynamic type of the fd, because as I understand it, different drivers could interpret the same integer opcode as different operations.

src/io/ioctl.rs
notgull: > Why is this using a "bad" opcode? FIONBIO seems like a perfectly good opcode to me ;-).

"Bad" opcode refers to the "old" strategy of handling opcodes, where it's just a number instead of a composition of the IOCTL code, subtype, direction, etc etc.

src/fs/ioctl.rs
notgull: I've made the `ioctl` call `unsafe` for now.

src/io/ioctl.rs
sunfishcode: Could we call them "old" opcodes instead? And add a comment explaining that "old" refers to how the opcode is encoded as just a number?

src/ioctl/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+    /// Get a pointer to the data to be passed to the `ioctl` command.
+    ///
+    /// See trait-level documentation for more information.
+    fn as_ptr(&mut self) -> *mut c::c_void;
+
+    /// Cast the output data to the correct type.
+    ///
+    /// # Safety
+    ///
+    /// The `ptr` must be the resulting value after a successful `ioctl` call.
```
sunfishcode, line 165: Could you add a comment here saying that `out` is the return value of the `ioctl` call if it doesn't fail, and `ptr` is the pointer that was passed to the `ioctl` call?

And perhaps also rename it to something like `extract_output` or so, because it isn't always using the ptr.

sunfishcode:
Thanks!

PR #798, opened 2023-08-28
Update freebsd ci

morr0ne:
This pr updates the cirrus ci config to the latest version of freebsd 12 and freebsd 13. It also adds freebsd 14 as a testing target. The freebsd image has to be pinned manually since freebsd 14 has not been release yet. A stable release is very near so I think it makes sense to also start testing on that version.


sunfishcode:
Sounds good, thanks!

PR #800, opened 2023-08-28
Put features that use global allocation behind cargo feature.

sunfishcode:
Put features that use global allocation, such as `ttyname`, that return owned strings, or `getgroups` that returns owned `Vec`s, behind a "global-allocator" cargo feature, which is enabled by default and with "std", but can be disabled with `default-features = false`.


notgull:
In other crates, this feature is generally named `alloc`. I would prefer that name in order to fit with the convention

sunfishcode:
I had named it that to avoid conflicting with the `alloc` dependency used for rustc-dep-of-std mode, however it looks like I can use the new `dep:` syntax in Cargo.toml to fix that instead.

sunfishcode:
Looks like the `dep:` feature works. Thanks for prompting me to look into that!

PR #804, opened 2023-08-30
Add a `runtime::entry` function.

sunfishcode:
This returns the program entrypoint, which is needed in origin. It's in the undocumented `runtime` module.


src/backend/linux_raw/param/auxv.rs
```diff
@@ -121,14 +123,31 @@ pub(in super::super) fn sysinfo_ehdr() -> *const Elf_Ehdr {
...
+pub(crate) fn entry() -> usize {
+    let mut entry = ENTRY.load(Relaxed);
+
+    if entry == 0 {
+        init_auxv();
+        entry = ENTRY.load(Relaxed);
+    }
+
+    entry
+}
```
notgull, line 137: Shouldn't this be a pointer for provenance reasons? I'm not familiar with Linux program startup.1

PR #824, opened 2023-09-12
Updated deps

Dirreke:
Bump linux-raw-sys from 0.4.7 to 0.5.2
Bump itoa from 1.0.1 to 1.0.9
Bump tempfile from 3.5.0 to 3.8.0
Bump once_cell from 1.5.2 to 1.18.0
Bump libc_errno from 0.3.1 to 0.3.3
Bump ctor from 0.2.0 to 0.2.4

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,17 +17,17 @@ rust-version = "1.63"
 
 [dependencies]
 bitflags = { version = "2.4.0", default-features = false }
-itoa = { version = "1.0.1", default-features = false, optional = true }
+itoa = { version = "1.0.9", default-features = false, optional = true }
 
 # Special dependencies used in rustc-dep-of-std mode.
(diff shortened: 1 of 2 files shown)


sunfishcode:
linux-raw-sys 0.5 updates to Linux 6.4 and I haven't yet done the work to determine whether it has semver implications for rustix or other users. I've been continuing to maintain linux-raw-sys 0.4, so it's not urgent to update to 0.5.

The other changes here are all semver-compatible updates, and to my knowledge, rustix doesn't depend on these newer versions. I've learned as a mtaintainer that it's valuable to give downstream users the flexibility to chose which versions they use.

Dirreke:
Actually, what I really want is the csky support. Then I will make a new PR to add csky to linux-raw-sys.

PR #831, opened 2023-09-18
fs: Add `StatxFlags::MOUNT_ROOT`

cgwalters:
I wanted to use this in some ostree code to detect a mount point root efficiently.

### DIFF
--- a/src/fs/statx.rs
+++ b/src/fs/statx.rs
@@ -21,6 +21,35 @@ use compat::statx as _statx;
 /// # References
 ///  - [Linux]
 ///
+/// # Examples
+///
+/// ```
+/// # use std::path::Path;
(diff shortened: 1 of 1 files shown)


sunfishcode:
`STATX_ATTR_MOUNT_ROOT` is an `stx_attributues_mask` flag, while the `StatxFlags` type is only for the `flags` argument to the `statx` call.

Rustix currently just returns the `stx_attributes_mask` mask as the raw `u64` value, so in the code where you want to use this, you can test it for `libc::STAX_ATTR_MOUNT_ROOT` manually.

Another option would be to change rustix to always use a custom `statx` struct so that it can replace `stx_attributes_mask` with a bitflags type with all the `STATX_ATTR_*` flags.

PR #832, opened 2023-09-18
Don't truncate the last character in ttyname

asomers:
CStr::len(), as used in backend::libc::termios::syscalls::ttyname, does not include the trailing NUL.  But CString::from_vec_with_nul_unchecked requires the NUL to be present.  Without it, it drops off the last character from the string.  This led to it returning paths like "/dev/pts/" instead of "/dev/pts/8".

Tested on FreeBSD and Linux x86_64


sunfishcode:
Wow, thanks for catching that!

PR #834, opened 2023-09-19
Support MIPS R6 targets in ioctl/linux.rs

chenx97:
To fix a regression introduced by this new piece of code on these tier-3 targets.


sunfishcode:
Thanks!

sunfishcode:
This is now released in rustix 0.38.14.

PR #847, opened 2023-09-22
Add `MapFlags::hugetlb_with_size_log2` method

newpavlov:
Currently it's not possible to create `MapFlags` for huge page sizes outside of 2 MiB and 1 GiB without going all the way down to `from_bits`.

### DIFF
--- a/src/mm/mmap.rs
+++ b/src/mm/mmap.rs
@@ -16,6 +16,31 @@ pub use backend::mm::types::MlockFlags;
 pub use backend::mm::types::MremapFlags;
 pub use backend::mm::types::{MapFlags, MprotectFlags, ProtFlags};
 
+impl MapFlags {
+    /// Create `MAP_HUGETLB` with provided size of huge page.
+    ///
+    /// Under the hood it computes `MAP_HUGETLB | (huge_page_size_log2 << MAP_HUGE_SHIFT)`.
+    /// `huge_page_size_log2` denotes logarithm of huge page size to use and should be
+    /// between 16 and 63 (inclusively).
+    ///
+    /// ```
+    /// use rustix::mm::MapFlags;
+    ///
+    /// let f = MapFlags::hugetlb_with_size_log2(30).unwrap();
+    /// assert_eq!(f, MapFlags::HUGETLB | MapFlags::HUGE_1GB);
+    /// ```
+    #[cfg(linux_kernel)]
+    pub const fn hugetlb_with_size_log2(huge_page_size_log2: u32) -> Option<Self> {
+        use linux_raw_sys::general::{MAP_HUGETLB, MAP_HUGE_SHIFT};
+        if 16 <= huge_page_size_log2 && huge_page_size_log2 <= 63 {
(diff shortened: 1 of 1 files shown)


src/backend/linux_raw/mm/types.rs
```diff
@@ -111,6 +111,30 @@ bitflags! {
...
+    /// `huge_page_size_log2` denotes logarithm of huge page size to use and should be
+    /// between 16 and 63 (inclusively).
+    ///
+    /// ```
+    /// use rustix::mm::MapFlags;
+    ///
+    /// let f = MapFlags::hugetlb_with_size(30).unwrap();
+    /// assert_eq!(f, MapFlags::HUGETLB | MapFlags::HUGE_1GB);
+    /// ```
+    pub const fn hugetlb_with_size(huge_page_size_log2: u32) -> Option<Self> {
```
sunfishcode, line 127: Could you name this something that indicates a log2 encoding, since "size" often means "size in bytes"? Perhaps with a "p2" in the name, following [`.p2align](https://ftp.gnu.org/old-gnu/Manuals/gas-2.9.1/html_node/as_111.html), so something like `huge_p2`?

sunfishcode:
Also, it's a goal of rustix to provide the same functionality on the libc backend as the linux_raw backend; could you move this code into src/mm/mmap.rs or so, so that it can be used by both?

src/backend/linux_raw/mm/types.rs
newpavlov: How about `hugetlb_with_log2_size` or `hugetlb_with_size_log2`, to make it consistent with the argument name? I don't think `p2` would be meaningful for most users.

newpavlov:
>could you move this code into src/mm/mmap.rs or so, so that it can be used by both?

Can do, but from where should I import `MAP_HUGETLB` and `MAP_HUGE_SHIFT`?

src/backend/linux_raw/mm/types.rs
sunfishcode: `hugetlb_with_size_log2` sounds reasonable to me. Or `huge_with_size_log2` since the specific constants like `MAP_HUGE_1GB` don't have the "tlb" in the name.

sunfishcode:
The libc backend can use `linux_raw_sys`, as long as it's under `#[cfg(linux_kernel)]` which this will need to be anyway, and as long as we're reasonable sure that popular libc implementations won't ever define these macros with a different name, which seems like a safe assumption here.

src/backend/linux_raw/mm/types.rs
newpavlov: I think `hugetlb_with_size_log2` is a better name since it includes the `MAP_HUGETLB` bit, while `MAP_HUGE_1GB` does not.

newpavlov:
>as we're reasonable sure that popular libc implementations won't ever define these macros with a different value, which seems like a safe assumption here

It's a matter of kernel ABI, not of `libc` API, no?

newpavlov:
Should I do it? I think the current code is fine.

sunfishcode:
I think it's ok as-is.

sunfishcode:
This is now released in rustix 0.38.15.

Issue #705, opened 2023-06-25
Missing shm_open?


Arisa-Snowbell:

Arisa-Snowbell:
Okay, It seems just creating file normal way in Rust works completely okay, I just have to specify /dev/shm prefix to the shared memory name.
So maybe there is no need to add it? Not sure if this is a completely correct way to do that but it seems to be what glibc does.

valpackett:
> Not sure if this is a completely correct way to do that but it seems to be what glibc does.

It's a Linux-specific way; on other systems shm fd's aren't necessarily implemented as files on the filesystem.

Issue #797, opened 2023-08-27
Add an API for shared memory APIs


notgull:
`softbuffer` currently uses `shmget`, `shmctl`, and other shared memory APIs. It's the only thing left preventing us from switching from `nix` to `rustix`.

sunfishcode:
At a glance, the specific items needed seem to be:
 - `shmat`
 - `shmdt`
 - `shmget`
 - `shmctl`
 - `ShmAtFlags` bitflags for `SHM_*` for `shmat`
 - `ShmGetFlags` bitflags for `SHM_*` and `IPC_*` for `shmctl`
 - `ShmIdDs` type to correspond to struct `shmid_ds`
 - `ShmCmd` enum for the `SHM_*` and `IPC_*` that can be used with `shmctl`

owenthewizard:
Are the SysV APIs planned?

sunfishcode:
I'm not aware of anyone working on them at this time.

PR #848, opened 2023-09-22
Add `shm_open`/`shm_unlink` (POSIX shared memory)

ids1024:
Fixes https://github.com/bytecodealliance/rustix/issues/705.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,6 +149,9 @@ thread = ["linux-raw-sys/prctl"]
 # Enable `rustix::process::*`.
 process = ["linux-raw-sys/prctl"]
 
+# Enable `rustix::shm::*`.
+shm = ["fs"]
+
 # Enable `rustix::time::*`.
 time = []
 
@@ -198,6 +201,7 @@ all-apis = [
     "pty",
     "rand",
     "runtime",
+    "shm",
     "system",
     "stdio",
     "termios",
--- a/src/backend/libc/mod.rs
+++ b/src/backend/libc/mod.rs
@@ -184,6 +184,14 @@ pub(crate) mod pid;
 #[cfg(any(feature = "process", feature = "thread"))]
 #[cfg(linux_kernel)]
 pub(crate) mod prctl;
+#[cfg(not(any(
+    windows,
+    target_os = "android",
+    target_os = "espidf",
+    target_os = "wasi"
+)))]
+#[cfg(feature = "shm")]
+pub(crate) mod shm;
 #[cfg(any(feature = "fs", feature = "thread", feature = "process"))]
 #[cfg(not(any(windows, target_os = "wasi")))]
 pub(crate) mod ugid;
--- a/src/backend/libc/shm/mod.rs
+++ b/src/backend/libc/shm/mod.rs
@@ -1,0 +1,2 @@
+pub(crate) mod syscalls;
+pub(crate) mod types;
--- a/src/backend/libc/shm/syscalls.rs
+++ b/src/backend/libc/shm/syscalls.rs
@@ -1,0 +1,25 @@
+use crate::ffi::CStr;
+
+use crate::backend::c;
+use crate::backend::conv::{c_str, ret, ret_owned_fd};
+use crate::fd::OwnedFd;
+use crate::fs::Mode;
+use crate::io;
+use crate::shm::ShmOFlags;
+
+pub(crate) fn shm_open(name: &CStr, oflags: ShmOFlags, mode: Mode) -> io::Result<OwnedFd> {
+    // On this platforms, `mode_t` is `u16` and can't be passed directly to a
+    // variadic function.
+    #[cfg(apple)]
+    let mode: c::c_uint = mode.bits().into();
+
+    // Otherwise, cast to `mode_t` as that's what `open` is documented to take.
+    #[cfg(not(apple))]
+    let mode: c::mode_t = mode.bits() as _;
+
+    unsafe { ret_owned_fd(c::shm_open(c_str(name), bitflags_bits!(oflags), mode)) }
+}
+
+pub(crate) fn shm_unlink(name: &CStr) -> io::Result<()> {
+    unsafe { ret(c::shm_unlink(c_str(name))) }
+}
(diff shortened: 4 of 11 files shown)


src/backend/linux_raw/fs/syscalls.rs
```diff
@@ -1458,6 +1458,45 @@ pub(crate) fn copy_file_range(
...
+pub(crate) fn shm_open(name: &CStr, oflags: OFlags, mode: Mode) -> io::Result<OwnedFd> {
+    let path = get_shm_name(name)?;
+    // In Rust 1.69, can use `from_bytes_until_nul`
+    let path = &path[..=path.iter().position(|x| *x == 0).unwrap()];
+    open(CStr::from_bytes_with_nul(&path).unwrap(), oflags, mode)
+}
+
+pub(crate) fn shm_unlink(name: &CStr) -> io::Result<()> {
+    let path = get_shm_name(name)?;
+    let path = &path[..=path.iter().position(|x| *x == 0).unwrap()];
```
sunfishcode, line 1496: We could avoid these `position` calls by having `get_shm_name` also return the length of the name, which it already knows.

sunfishcode (commented):
Thanks for the patch! Overall this looks good.

My main concern here is that I'm not sure `shm_open` and `shm_unlink` belong in `rustix::fs`. There is no clear rule, but my instinct is that the shared-memory namespace is distinct from the filesystem namespace, even if there is aliasing on Linux, that it justifies putting these in a new `rustix::shm` module. What would you think about that?

ids1024:
I put it next to `memfd_create`. But it could make sense to have an `shm` module containing both of these things, and System V shared memory (https://github.com/bytecodealliance/rustix/issues/797). `memfd_create` certainly isn't a particularly "file system" related API.

src/backend/linux_raw/fs/syscalls.rs
ids1024: Good point. I originally wrote it with `from_bytes_until_nul`, but then CI complained since that's too new of a standard library addition.

Either `from_bytes_with_nul` or `from_bytes_until_nul` still need to traverse the string. `method.from_bytes_with_nul_unchecked` would avoid that.

ids1024:
Of course, moving `memfd_create` would be a breaking API change.

sunfishcode:
Here's a rough sketch for how we might group all the things in this space that we may eventually add:

 - `shm` - POSIX shared memory
 - `mq` - POSIX message queues
 - `sysvipc::sem`, `sysvipc::shm`, and `sysvipc::msg`, for System V semaphores, shared memory, and message queues
 - `fs` - `memfd_create`

The unifying themes for the `shm` module are the namespace, and the fact that on some platforms, `shm_open` is not backed by plain files (even if it is plain files on Linux). `memfd_create` doesn't fit into either of those two themes. The unifying themes for the `fs` namespace are the filesystem namespace, and also operations on file objects, like `fstat`, `lseek`, `ftruncate`, and so on. `memfd_create` doesn't interact with the filesystem namespace, but it does produce a completely file-like file object, so it goes in `fs`.

Does that make sense?

ids1024:
`shm_open` and `shm_close` seem more "fs" like than `memfd_create`, given they mirror `open` and `close`, while also returning file descriptors. Even if they don't necessarily exist in the filesystem.

But you're right that `memfd_create` is defined to return a "completely file-like object" (except for sealing support). I'd expect `shm_open` to as well, and on Linux it does, but the POSIX spec doesn't actually seem to say exactly what operations are supported, and macOS does in fact seem to  be quite restrictive about what calls work with shm fds.

sunfishcode:
Ultimately, what we're talking about here is just the top-level API organization. You can always call `fs` functions on `shm` file descriptors, regardless of how we organize things. Just like you can always call many `fs` functions like `fstat` on socket fds from the `net` module. It's not a strict separation, just a general grouping.

Interestingly, POSIX did add a "future direction" note that `shm_open`'s objects may have full file semantics in the future. If that ever happens, and macOS implements it, maybe that'd tip the scales and we should move `shm_open` into `fs`. But we don't need to predict that future today; we have tools like deprecation and semver for evolving APIs. So my sense is, until then, `shm` makes sense.

ids1024:
Moving it out of `fs` is a bit complicated with how things currently work, since it uses `open` and `unlink` from the `fs` module, as well as `Mode` and `OFlags`.

Though looking at the [POSIX spec](https://pubs.opengroup.org/onlinepubs/9699919799/functions/shm_open.html) and `shm_open(3)`, both note flags must contain "exactly one of" `O_RDONLY` and `O_RDWR`, and optionally `O_CREAT`, `O_EXCL`, and `O_TRUNC`. But no other flags. So possibly a custom flags type may be appropriate, instead of using the same one as `open`.

But `Mode` would probably be unchanged, so it makes sense to use the same type.

sunfishcode:
> Moving it out of `fs` is a bit complicated with how things currently work, since it uses `open` and `unlink` from the `fs` module, as well as `Mode` and `OFlags`.

It'd be ok to have shm depend on fs, with `shm = ["fs"]` in Cargo.toml, and then the `shm` module can use functions and types in the `fs` module.

> Though looking at the [POSIX spec](https://pubs.opengroup.org/onlinepubs/9699919799/functions/shm_open.html) and `shm_open(3)`, both note flags must contain "exactly one of" `O_RDONLY` and `O_RDWR`, and optionally `O_CREAT`, `O_EXCL`, and `O_TRUNC`. But no other flags. So possibly a custom flags type may be appropriate, instead of using the same one as `open`.

A custom flags type for this makes sense to me.

Also, looking at the `shm_open` docs, I just noticed "The FD_CLOEXEC file descriptor flag associated with the new file descriptor is set.", so we should have the code add `OFlags::CLOEXEC` when calling `open` in the linux_raw backend.

> But `Mode` would probably be unchanged, so it makes sense to use the same type.

Makes sense. We can declare a type alias in the `shm` module for convenience.

ids1024:
Ah, I should have thought about `CLOEXEC`.

This now has a passes that and has a separate flags type.

sunfishcode (commented):
Looks good, thanks!

sunfishcode:
This is now released in #848.

PR #850, opened 2023-09-23
Handle static PIE in `startup_tls_info`.

sunfishcode:
In static PIE mode, we don't get a `PT_PHDR` segment, but we do get a `PT_DYNAMIC` segment, and we can compare that against the `_DYNAMIC` symbol that the linker creates and dynamic linker relocates to learn our virtual address offset.


sunfishcode:
I think a better approach here is to move `startup_tls_into` out of rustix and into origin, which I've started in sunfishcode/origin#58, but for now, it makes sense to fix rustix's version to handle the static PIE case correctly.

PR #856, opened 2023-09-28
build.rs: fix mips32r6 detection for libc feature switch

chenx97:
and simplify the logic since inline assembly support for all variants of MIPS is experimental only.


sunfishcode:
Thanks!

sunfishcode:
This is now released in rustix 0.38.15.

PR #861, opened 2023-09-30
Fix Linux custom ioctl opcode computation

kevinmehall:
When attempting to use the recently added [custom ioctl APIs](https://github.com/bytecodealliance/rustix/pull/788) I was getting `Inappropriate ioctl for device` errors and found that the ioctl opcodes being passed to the kernel were equal to just the `NUM` part with the other components ignored. This seems to be because the shift and the mask were performed in the wrong order when combining the components.

I also added:
  * a test comparing a few computed opcodes to known opcodes from `linux_raw_sys` as a sanity check
  * documentation referring to the C macros where you would find the values to use with these functions / types.

By the way, I noticed that this library is itself using only `BadOpcode` with constants from `linux_raw_sys` rather than the `ReadOpcode` / `WriteOpcode` types. Is there any reason to prefer `ReadOpcode` / `WriteOpcode` over using an opcode found in `linux_raw_sys`?

The name `Bad` made it seem like something I shouldn't be using, but taking the value by name from `linux_raw_sys` seems less error prone than looking up these components in a C header file and hard-coding them.

cc @notgull

### DIFF
--- a/src/ioctl/linux.rs
+++ b/src/ioctl/linux.rs
@@ -10,7 +10,7 @@ pub(super) const fn compose_opcode(
     num: RawOpcode,
     size: RawOpcode,
 ) -> RawOpcode {
-    macro_rules! shift_and_mask {
+    macro_rules! mask_and_shift {
         ($val:expr, $shift:expr, $mask:expr) => {{
             ($val & $mask) << $shift
         }};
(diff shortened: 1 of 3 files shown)


notgull (approved):
Whoopsie daisy

src/ioctl/linux.rs
```diff
@@ -12,7 +12,7 @@ pub(super) const fn compose_opcode(
 ) -> RawOpcode {
     macro_rules! shift_and_mask {
         ($val:expr, $shift:expr, $mask:expr) => {{
-            ($val << $shift) & $mask
```
sunfishcode, line 15: Should we rename the macro to `mask_and_shift`?

kevinmehall: Good idea, renamed.

sunfishcode:
Thanks!

sunfishcode:
This is now released in rustix 0.38.17.

PR #889, opened 2023-10-21
Make `cmsg_space!` usable in const contexts.

sunfishcode:
Make `cmsg_space!` usable in const contexts, so that it can be used as a buffer size argument, and add a version of tests/net/unix.rs that uses stack-allocated buffers instead of `Vec`s.

This exposes an alignment sublety, that buffers must be aligned to the needed alignment of `cmsghdr`; handle this by auto-aligning the provided buffer to the needed boundary.


notgull (approved):

sunfishcode:

PR #892, opened 2023-10-23
Fix `renameat_with` with flags on Android.

sunfishcode:
When `renameat_with` is passed non-zero flags, fall back to the `syscall!` macro on ABIs such as Android where libc has a `renameat` but not a `renameat2`.

Fixes Stebalien/tempfile#259.


sunfishcode:

PR #895, opened 2023-10-25
Add connect_unspec

badeend:
Main use case: to change a UDP socket's association.

### DIFF
--- a/src/backend/libc/net/write_sockaddr.rs
+++ b/src/backend/libc/net/write_sockaddr.rs
@@ -101,3 +101,38 @@ unsafe fn write_sockaddr_unix(unix: &SocketAddrUnix, storage: *mut SocketAddrSto
     core::ptr::write(storage.cast(), unix.unix);
     unix.len()
 }
+
+/// Even though an AF_UNSPEC socket address contains only the family, they're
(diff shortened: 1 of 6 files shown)


src/backend/libc/net/write_sockaddr.rs
```diff
@@ -101,3 +101,38 @@ unsafe fn write_sockaddr_unix(unix: &SocketAddrUnix, storage: *mut SocketAddrSto
     core::ptr::write(storage.cast(), unix.unix);
     unix.len()
 }
+
+/// Even though an AF_UNSPEC socket address contains only the family, they're
+/// still required to be at least as big as the socket's regular address structure.
+/// We (ab)use `sockaddr_in6` to push the size up to satisfy both IPv4 and IPv6 sockets.
```
sunfishcode, line 107: Could you say more about what requires the encoded socket to be as big as the socket's regular address structure? If the socket is not IPv4 or IPv6, what would happen here? Would it make sense to use `c::sockaddr_storage` here instead?

badeend:
Replaced by https://github.com/bytecodealliance/rustix/pull/902

PR #896, opened 2023-10-25
Fix p{read,write}v{,v2}'s encoding of the offset argument on Linux.

sunfishcode:
Unlike with `p{read,write}`, Linux's `p{read,write}v` syscall's offset argument is not passed in an endian-specific order. And, the expectation is for syscall wrappers to always pass both the high and low halves of the offset as separate arguments, even though on 64-bit architectures the low half is passed throgh as a 64-bit value containing the full offset and the kernel doesn't mask it.

And `p{read,write}v2` follow the behavior of `p{read,write}`.


sunfishcode:

PR #901, opened 2023-10-25
Fix getsockopt zeroing memory

alexcrichton:
This commit fixes an accidental regression from #858 where a previous call to `core::mem::zeroed` was replaced with `MaybeUninit::uninit` instead of `MaybeUninit::zeroed`.


sunfishcode:
Thanks for tracking this down and fixing this!

sunfishcode:

PR #902, opened 2023-10-26
Add disconnect_v4 & disconnect_v6

badeend:
Main use case: to change a UDP socket's association.

### DIFF
--- a/src/backend/libc/net/syscalls.rs
+++ b/src/backend/libc/net/syscalls.rs
@@ -6,7 +6,7 @@ use crate::backend::c;
 use crate::backend::conv::{borrowed_fd, ret, ret_owned_fd, ret_send_recv, send_recv_len};
 use crate::fd::{BorrowedFd, OwnedFd};
 use crate::io;
-use crate::net::{SocketAddrAny, SocketAddrV4, SocketAddrV6};
+use crate::net::{Ipv4Addr, Ipv6Addr, SocketAddrAny, SocketAddrV4, SocketAddrV6};
 use crate::utils::as_ptr;
 use core::mem::{size_of, MaybeUninit};
 #[cfg(not(any(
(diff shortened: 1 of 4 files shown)


badeend:
Took me a few tries, but it looks like I've got it all working now

src/net/socket.rs
```diff
@@ -453,6 +453,80 @@ pub fn connect_unix<Fd: AsFd>(sockfd: Fd, addr: &SocketAddrUnix) -> io::Result<(
...
+/// [NetBSD]: https://man.netbsd.org/connect.2
+/// [OpenBSD]: https://man.openbsd.org/connect.2
+/// [DragonFly BSD]: https://man.dragonflybsd.org/?command=connect&section=2
+/// [illumos]: https://illumos.org/man/3SOCKET/connect
+/// [glibc]: https://www.gnu.org/software/libc/manual/html_node/Connecting.html
+#[inline]
+#[doc(alias = "connect")]
+pub fn disconnect_v4<Fd: AsFd>(sockfd: Fd) -> io::Result<()> {
+    backend::net::syscalls::disconnect_v4(sockfd.as_fd())
+}
```
sunfishcode, line 491: Could you say more about why having separate `_v4` and `_v6` interfaces is needed here? It seems like the implementation could use a zeroed `sockaddr_storage` with a `ss_family` set to `AF_UNSPEC`, which should work for any type of socket.

src/backend/libc/net/syscalls.rs
```diff
@@ -217,21 +217,57 @@ pub(crate) fn bind_unix(sockfd: BorrowedFd<'_>, addr: &SocketAddrUnix) -> io::Re
...
+
+#[cfg(not(any(target_os = "redox", target_os = "wasi")))]
+pub(crate) fn disconnect_v4(sockfd: BorrowedFd<'_>) -> io::Result<()> {
+    #[allow(unused_mut)]
+    let mut sockaddr = encode_sockaddr_v4(&SocketAddrV4::new(Ipv4Addr::UNSPECIFIED, 0));
+
+    #[cfg(linux_kernel)]
+    {
+        sockaddr.sin_family = c::AF_UNSPEC as _;
+    }
```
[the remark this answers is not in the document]
badeend: I'm puzzled...

That's exactly what I had in #895 , but the FreeBSD tests kept failing on it. However, now that I've removed the conditional in this PR and always assign AF_UNSPEC, FreeBSD is working just fine with it too.. :neutral_face:
I must have overlooked something somewhere.

src/net/socket.rs
badeend: (See comment above)
Because I thought FreeBSD didn't support AF_UNSPEC. However, now I see it _should_ work, so I'll revisit my previous PR

badeend:
The remarks have been resolved in #903

PR #908, opened 2023-10-28
Implement the Buffer trait

### LINKED ISSUE #81: Reading into uninitialized buffers
Currently, [`read`](https://docs.rs/rsix/latest/rsix/io/fn.read.html) takes a `&mut [u8]`. That's fine for many use cases, but it does require the buffer to be initialized before calling `read`, so it's currently not possible to read directly into uninitialized memory.

For Mustang, the C `read` API takes a raw pointer which may be pointing to an uninitialized buffer. Currently mustang does `slice::from_raw_parts_mut`, however this has undefined behavior if the slice is uninitialized.

notgull:
This is for discussion purposes before I go through the codebase and replace the
rest of the code.

Closes #81 

@sunfishcode @SUPERCILEX @ids1024 You all seem to work a lot with `rustix`, any thoughts on this API before I commit to it? It doesn't really support vectored reading but neither does libstd.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -5,6 +5,25 @@
 use core::mem::MaybeUninit;
 use core::slice;
 
+use crate::io::Result;
+
+/// Wrap around a buffer and do something with the pointer and length.
+///
+/// # Safety
+///
+/// The callback `f` must actually write the return value's worth of instances of `T` to the
+/// buffer.
+#[inline]
+pub(crate) unsafe fn with_buffer<
+    T: AnyBitPattern,
+    Buf: Buffer<T>,
+    F: FnOnce(*mut T, usize) -> Result<usize>
+>(mut buffer: Buf, f: F) -> Result<Buf::Result> {
+    let (ptr, cap) = buffer.as_buffer();
+    let len = f(ptr, cap)?;
+    Ok(buffer.finish(len))
+}
+
 /// A buffer that the system can write into.
 pub unsafe trait Buffer<T: AnyBitPattern>: Sized {
     /// The result of the process operation.
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1,0 +1,94 @@
+//! Code for handling buffers to write into.
+
+#![allow(unsafe_code)]
+
+use core::mem::MaybeUninit;
+use core::slice;
+
+/// A buffer that the system can write into.
+pub unsafe trait Buffer<T: AnyBitPattern>: Sized {
+    /// The result of the process operation.
+    type Result;
+
+    /// Convert this buffer into a pointer to a buffer and its capacity.
+    fn as_buffer(&mut self) -> (*mut T, usize);
+
+    /// Convert a finished buffer pointer into its result.
+    unsafe fn finish(self, len: usize) -> Self::Result;
+}
+
+unsafe impl<T: AnyBitPattern> Buffer<T> for &mut [T] {
+    type Result = usize;
+
+    #[inline]
+    fn as_buffer(&mut self) -> (*mut T, usize) {
+        (self.as_mut_ptr(), self.len())
+    }
+
+    #[inline]
+    unsafe fn finish(self, len: usize) -> Self::Result {
+        len
+    }
+}
+
+unsafe impl<'a, T: AnyBitPattern> Buffer<T> for &'a mut [MaybeUninit<T>] {
+    type Result = (&'a mut [T], &'a mut [MaybeUninit<T>]);
+
+    #[inline]
+    fn as_buffer(&mut self) -> (*mut T, usize) {
+        (self.as_mut_ptr().cast(), self.len())
+    }
+
+    #[inline]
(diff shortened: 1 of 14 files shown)


SUPERCILEX:
Long term, the method signature should probably be `fn read<Fd: AsFd>(fd: Fd, buf: BorrowedCursor<'_>) -> Result<usize>`. Note that I haven't thought about the lifetime, so might be wrong. Also, returning a `usize` is technically unnecessary as that information is encoded in the length of the cursor, but having to store a `prev_len` variable can be annoying. Then again, maybe consistency with the stdlib ([`fn read_buf(&mut self, buf: BorrowedCursor<'_>) -> Result<()>`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_buf)) is more important.

Sample implementation:
https://github.com/rust-lang/rust/blob/7cc36de72d9a5fd6881946c673ff47586214ad1e/library/std/src/sys/unix/fd.rs#L145-L159

---

I kind of think rustix should just make a straight copy of the `Borrowed{Buf,Cursor}` structs and then swap over to the stdlib implementation once they're stable.

Not sure how I feel about a different (though reasonable) route to the stdlib.

src/buffer.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    #[inline]
+    unsafe fn finish(mut self, len: usize) -> Self::Result {
+        self.set_len(len);
+        self
+    }
+}
+
+/// Types made up of plain-old-data.
```
ids1024, line 75: So this is supposed to be equivalent to requirements of the [`AnyBitPattern`](https://docs.rs/bytemuck/1.14.0/bytemuck/trait.AnyBitPattern.html) type in `bytemuck`, but avoids adding that as a dependency?

Bytemuck distinguishes that from its `Pod` trait which has slightly different requirements.

notgull: A secondary goal for this trait is to act as a buffer for `epoll` and `kqueue` events, which can have uninitialized bytes (IIRC). Which is why I'm denoting it like this.

ids1024:
Matching the unstable `Read::read_buf` does seem like the logical thing to do.That was also my first thought here. It's annoying that still isn't stable.

src/buffer.rs
```diff
@@ -0,0 +1,113 @@
...
+/// Implements [`Buffer`] around the `Vec` type.
+///
+/// This implementation fills the buffer with data and sets the length.
+#[cfg(feature = "alloc")]
+unsafe impl<T: AnyBitPattern> Buffer<T> for alloc::vec::Vec<T> {
+    type Result = alloc::vec::Vec<T>;
+
+    #[inline]
+    fn as_buffer(&mut self) -> (*mut T, usize) {
+        (self.as_mut_ptr(), self.len())
```
sunfishcode, line 84: I think you could get away with using `self.capacity()` here rather than just `self.len()`, which would make this approach more attractive because one could pass in a `Vec::with_capacity(n)` and read into it without initializing it first, or take an existing `Vec`, `clear()` it, and then reuse the existing memory directly.

Another option would be to use `.`[`spare_capacity_mut`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.spare_capacity_mut)`().as_buffer()`, and say the behavior here is to append elements to the `Vec` rather than rewriting the whole `Vec`. That seems nice for `read_to_end` kinds of use cases. On the other hand, I wonder if that's too subtle. Given a `v: Vec<T>`, passing in `v` would append, while `&mut v` would start from the beginning. And passing `&v` to write would still start from the beginning.

sunfishcode (commented):
At first glance, I... want to think about this more :smile: .

On one hand, the `Buffer` trait is a neat idea.

On the other hand, it feels like the API is getting more complex and reaching in the direction of end-user ergonomics, when functions like `read` are still low-level primitives.

sunfishcode:
After thinking about this more, I think I'd like to see an API here where we add new functions that take uninitialized buffers, rather than modifying the existing functions, for two reasons.

I recently learned that replacing an argument of type `T` with a trait argument where `T` is one of multiple types that implement a trait is not completely semver-compatible. With plain `T`, users can do `.into()` at the callsite and it infers the destination type. With a trait, there is no specific destination that can be inferred, so users get a compile error. I don't know how likely it is that any user is doing  `.into()` to convert into a `&mut [u8]`, but it's theoretically possible.

And, std has not yet decided how it wants to deal with uninitialized buffers, but when it does, I assume we'll want to change rustix to use the same approach. That means we may need to deprecate something in the future, and if we do, it'll be easier to do that with separate functions.

So, what would you think about adding new functions, named something like `read_buf` or `read_raw`?

SUPERCILEX:
My main concern with accepting a raw MaybeUninit slice (and why I suggested copying the std Borrowed structs) is that every caller will need to use unsafe to assume_init the bytes that were written. That unsafe should be part of the library and not be the caller's responsibility.

I also agree that separate functions are a good idea.

notgull:
Closing this so I can rethink this

PR #909, opened 2023-10-29
Add support for NO_SOCKET_INHERIT

### LINKED ISSUE #888: Add the ability to set HANDLE_FLAG_INHERIT for Windows sockets
This is basically the Windows equivalent of CLOEXEC. Two ways of going about this:

- Add a `INHERIT` flag to `SocketFlags` on Windows.
- Make the value for `CLOEXEC` on Windows be `HANDLE_FLAG_INHERIT`.

The second option might make more sense but also might be a footgun.

notgull:
Closes #888 by making `CLOEXEC` in `SocketFlags` use the no-inherit socket property in Windows. `accept_with` seems complicated so I haven't implemented it yet.

### DIFF
--- a/src/net/types.rs
+++ b/src/net/types.rs
@@ -1436,7 +1436,7 @@ bitflags! {
         const NONBLOCK = bitcast!(c::SOCK_NONBLOCK);
 
         /// `SOCK_CLOEXEC`
-        #[cfg(not(any(apple, windows, target_os = "aix", target_os = "haiku")))]
+        #[cfg(not(any(apple, target_os = "aix", target_os = "haiku")))]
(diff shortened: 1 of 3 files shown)


src/net/types.rs
```diff
@@ -1436,7 +1436,7 @@ bitflags! {
         const NONBLOCK = bitcast!(c::SOCK_NONBLOCK);
 
         /// `SOCK_CLOEXEC`
```
sunfishcode, line 1438: It would be good to add a documentation comment here mentioning that on Windows, `CLOEXEC` is implemented by way of `WSA_FLAG_NO_HANDLE_INHERIT`.

notgull:
Too much of a headache for me at the moment

PR #920, opened 2023-11-07
Add MCL_ONFAULT constant to libc backend

newpavlov:
The constant was added in libc v0.2.150.


sunfishcode:
This can use the `bitcast!` macro to cast the value to avoid `i32`/`u32` differences, as is now done for the other `MCL_*` constants.

newpavlov:
Ooops... Missed that, will fix it right away.

newpavlov:
Should I add `#[cfg(linux_kernel)]` on `ONFAULT` only for the `libc` backend?

sunfishcode:
It's not necessary to test for `linux_kernel` in the linux_raw backend, as linux_raw only supports Linux-kernel platforms.

sunfishcode:
Thanks!

sunfishcode:

PR #925, opened 2023-11-08
Add ioctl `Updater` type

ids1024:
Many ioctls are defined with `_IOWR` and take input and return output through the same struct. We can handle this by taking an `&mut` pointer.

I'm not sure about the name. `Updater` seems better than `GetterSetter`, but maybe something else would be clear.

https://github.com/Smithay/drm-rs/pull/180 demonstrates a use of this. DRM has a lot of ioctls that work this way. I've also noticed this is needed for v4l2.

(Currently that drm-rs branch is defining macros similar to what Nix defines. The ioctl API in Rustix seems potentially a bit verbose when dealing with dozens of ioctls....)

### DIFF
--- a/src/ioctl/patterns.rs
+++ b/src/ioctl/patterns.rs
@@ -153,6 +153,49 @@ unsafe impl<Opcode: CompileTimeOpcode, Input> Ioctl for Setter<Opcode, Input> {
     }
 }
 
+/// Implements an “updater” pattern for `ioctl`s.
+///
+/// The ioctl takes a reference to a struct that it reads its input from,
+/// then writes output to the same struct.
+pub struct Updater<'a, Opcode, Value> {
+    /// Reference to input/output data.
+    value: &'a mut Value,
+
+    /// The opcode.
+    _opcode: PhantomData<Opcode>,
+}
+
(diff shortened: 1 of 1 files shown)


ids1024:
I guess any `_IOWR` should use this pattern? Not sure if the return value of `ioctl` would be used for any. (It would be redundant when data could be passed in the struct, but ioctls aren't known for consistency.) And `_IOR` likewise should always use a `Getter` pattern? While `_IOW` could be `Setter`, but occasionally ioctls take an integer directly instead of accepting a pointer.

It would be nice if the `CompileTimeOpcode` types encoded which of these "patterns" an ioctl uses, and then a method or associated function could perform the ioctl in a type-safe way without specifying again what type the ioctl uses.

ids1024:
https://docs.rs/nix/0.27.1/nix/sys/ioctl/index.html#working-with-arrays

Apparently some ioctls work with arrays. And things like `SPI_IOC_MESSAGE` are defined as macros that take the length, so the ioctl number depends on how many values as passed...

sunfishcode:
This PR looks reasonable to me. @notgull do you have any concerns here?

notgull (approved):
Looks good to me!

I didn't know IOCTL numbers could change based on the input data. This kind of changes the model.

sunfishcode:

PR #926, opened 2023-11-09
Fix ioctl numbers on BSDs

ids1024:
`<sys/ioccom.h>` defines `IOC_OUT` as `0x40000000UL`. But this is "out" from the perspective of the kernel, not the application. So this corresponds to `READ` rather than `WRITE`.

Tested on FreeBSD with https://github.com/Smithay/drm-rs/pull/180. This also seems to be correct for NetBSD and OpenBSD.

### DIFF
--- a/src/ioctl/bsd.rs
+++ b/src/ioctl/bsd.rs
@@ -18,7 +18,10 @@ pub(super) const fn compose_opcode(
     dir | num | (group << 8) | ((size & IOCPARAM_MASK) << 16)
 }
 
+// `IOC_VOID`
 pub const NONE: RawOpcode = 0x2000_0000;
-pub const WRITE: RawOpcode = 0x4000_0000;
-pub const READ: RawOpcode = 0x8000_0000;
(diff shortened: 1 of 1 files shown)


sunfishcode:
@notgull I don't know how these work; do these look right to you?

notgull:
Yeah those look right to me. I don't know how IOCTL numbers work on BSD. I copied from `nix` and I must have switched something around

ids1024:
The C macro definitions in https://github.com/freebsd/freebsd-src/blob/master/sys/sys/ioccom.h aren't *too* incomprehensible. `compose_opcode` here looks like it matches the BSD definition of `_IOC`. So I think everything should be right on the BSDs after this change.

sunfishcode:
Sounds good. Could you add a comment mentioning that (a) those correspond to `IOC_IN`/`IOC_OUT` and (b) that the perspective is inverted?

sunfishcode:

PR #935, opened 2023-11-17
Update `windows-sys` 0.52

kennykerr:
This is the first update to the `windows-sys` crate in 8 months.  https://github.com/microsoft/windows-rs/releases/tag/0.52.0


sunfishcode:
Thanks for the PR! The CI failure is in a pretty subtle test; I can take on the task of debugging that.

sunfishcode:
The errno crate depending on windows-sys 0.48 appears to be the cause of the CI failure as well, so both issues here will be resolved when errno updates.

sunfishcode:
errno 0.3.8 is now released, updated to windows-sys 0.52. I've now added a patch here to update errno to 0.3.8.

sunfishcode:
This is now released in 0.38.26.

PR #942, opened 2023-11-24
fix(net): Fix the WinError 10014

ZengGengSen:
When sendto_v4 and sendto_v6 are called,
    the error WinError 10014 is raised, caused
    by passing the wrong sockaddr_len

    Note: WSAEFAULT(10014), Bad address.
        The system detected an invalid pointer
        address in attempting to use a pointer
        argument of a call. This error occurs
        if an application passes an invalid
        pointer value, or if the length of the
        buffer is too small. For instance, if
        the length of an argument, which is a
        sockaddr structure, is smaller than
        the sizeof(sockaddr).

### DIFF
--- a/src/backend/libc/net/syscalls.rs
+++ b/src/backend/libc/net/syscalls.rs
@@ -100,7 +100,7 @@ pub(crate) fn sendto_v4(
             send_recv_len(buf.len()),
             bitflags_bits!(flags),
             as_ptr(&encode_sockaddr_v4(addr)).cast::<c::sockaddr>(),
-            size_of::<SocketAddrV4>() as _,
+            size_of::<c::sockaddr_in> as _,
         ))
     }
 }
@@ -119,7 +119,7 @@ pub(crate) fn sendto_v6(
             send_recv_len(buf.len()),
             bitflags_bits!(flags),
             as_ptr(&encode_sockaddr_v6(addr)).cast::<c::sockaddr>(),
-            size_of::<SocketAddrV6>() as _,
+            size_of::<c::sockaddr_in6> as  c::socklen_t,
         ))
     }
 }


sunfishcode:
This change looks right to me. Do you have any guess as to why the `net_v4_sendto` and `net_v6_sendto` tests in tests/net/connect_bind_send.rs don't catch this?

src/backend/libc/net/syscalls.rs
```diff
@@ -100,7 +100,7 @@ pub(crate) fn sendto_v4(
             send_recv_len(buf.len()),
             bitflags_bits!(flags),
             as_ptr(&encode_sockaddr_v4(addr)).cast::<c::sockaddr>(),
-            size_of::<SocketAddrV4>() as _,
+            size_of::<c::sockaddr_in> as _,
```
sunfishcode, line 103: Ah, probably the reason the tests are failing is that this is missing the `()` to call the `size_of` function.

src/backend/libc/net/syscalls.rs
```diff
@@ -119,7 +119,7 @@ pub(crate) fn sendto_v6(
             send_recv_len(buf.len()),
             bitflags_bits!(flags),
             as_ptr(&encode_sockaddr_v6(addr)).cast::<c::sockaddr>(),
-            size_of::<SocketAddrV6>() as _,
+            size_of::<c::sockaddr_in6> as  c::socklen_t,
```
sunfishcode, line 122: And same here.

src/backend/libc/net/syscalls.rs
ZengGengSen: Haha, I was a little careless, OK, I fixed the error and re-pulled a new pull request

sunfishcode:
Closing in favor of #954.

PR #944, opened 2023-11-25
Add methods for reading into unitialized buffers

notgull:
An alternative to #908 that uses separate functions to enable reading into uninitialized bytes.

So far I've added uninitialized methods for `io` methods, `net` methods, and `rand`. There are other functions that take `&mut [u8]`, but those aren't as pressing for now.

cc #81

### DIFF
--- a/src/backend/libc/io/syscalls.rs
+++ b/src/backend/libc/io/syscalls.rs
@@ -27,7 +27,7 @@ use {
 };
 
 pub(crate) unsafe fn read(fd: BorrowedFd<'_>, buf: *mut u8, len: usize) -> io::Result<usize> {
-    unsafe { ret_usize(c::read(borrowed_fd(fd), buf.cast(), min(len, READ_LIMIT))) }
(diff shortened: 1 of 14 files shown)


sunfishcode:
The overall shape here looks good! If, in the future, the standard library adopts some convention for uninitialized buffers, we can follow it.

What would you think about making the doc comments for these new `foo_uninit` functions just say something like "Like `foo`, but read into a potentially-uninitialized buffer instead" or so, instead of copying all the documentation content? I know I've done a fair amount of doc-comment copy and pasting myself, but I've been appreciating how shorter comments like this serve to highlight the difference, rather than oblige the reader to carefully compare to full doc comments to find the differences.

notgull:
@sunfishcode When do you think that you'll be able to review this? I want to implement `rustix` support for some other crates, but `MaybeUninit` support is a must for these use cases.

sunfishcode:
Sounds good! I'll merge this, do some additional testing, and then cut a new release.

sunfishcode:
This is now released in rustix 0.38.27.

Issue #941, opened 2023-11-23
Support for syscalls required for AF_XDP


arctic-alpaca:
Hi,

I'm experimenting with [AF_XDP](https://www.kernel.org/doc/html/next/networking/af_xdp.html) and would love to use `rustix` instead of `libc`. For the BPF side of things, I use [`aya`](https://github.com/aya-rs/aya), so only the non-BPF related syscalls remain. Some syscalls are already supported like `mmap` and `socket` but others would need to be implemented or expanded/modified. I created a list of missing syscalls, constants and structs. Most of the syscalls, etc were added in 4.18 with some additions and modifications in later kernel versions.

### Syscalls:
- [ ] [setsockopt](https://github.com/torvalds/linux/blob/v6.6/net/xdp/xsk.c#L1258-L1351)
    - [ ] `setsockopt(fd, SOL_XDP, XDP_UMEM_REG, umem_reg, sizeof(umem_reg))`, `umem_reg` see [structs](#structs)
    - [ ] `setsockopt(fs, SOL_XDP, XDP_UMEM_FILL_RING, fill_size, sizeof(fill_size))`, `fill_size` = `u32` (https://github.com/torvalds/linux/blob/v6.6/net/xdp/xsk.c#L944)
    - [ ] `setsockopt(fs, SOL_XDP, XDP_UMEM_COMPLETION_RING, comp_size, sizeof(comp_size))`,  `comp_size` = `u32`
    - [ ] `setsockopt(fs, SOL_XDP, XDP_RX_RING, rx_size, sizeof(rx_size))`, `rx_size` = `u32`
    - [ ] `setsockopt(fs, SOL_XDP, XDP_TX_RING, tx_size, sizeof(tx_size))` `tx_size` = `u32`
- [ ] [getsockopt](https://github.com/torvalds/linux/blob/v6.6/net/xdp/xsk.c#L1373-L1502)
    - [ ] `getsockopt(fd, SOL_XDP, XDP_MMAP_OFFSETS, *xdp_mmap_offsets, &optlen)`, `xdp_mmap_offsets` see [structs](#structs)
    - [ ] `getsockopt(fd, SOL_XDP, XDP_STATISTICS, *xdp_statistics, &optlen)`, `xdp_statistics` see [structs](#structs)
    - [ ] `getsockopt(fd, SOL_XDP, XDP_OPTIONS, *xdp_options, &optlen)`, `xdp_options` see [structs](#structs)
[...]

PR #946, opened 2023-11-27
Add AF_XDP syscalls, constants and types

arctic-alpaca:
Experimental implementation of parts of https://github.com/bytecodealliance/rustix/issues/941, the `if_*` methods are not implemented.

This is still missing actual tests and I'm unsure about the use of `#[cfg(target_os = "linux")]`. It is also a very real possibility, that I missed some implementation for `SocketAddrXdp`.

In my limited tests, I could create and use an `AF_XDP` socket as expected.

Since this depends on additions to `linux-raw-sys`, it currently depends on my fork.


arctic-alpaca:
The `use-libc` test seems to fail because of the libc fork I'm using currently using. 

The fork is necessary because the libc crate currently doesn't expose the required structs/constants. I created https://github.com/rust-lang/libc/pull/3453 to add them.

sunfishcode (commented):
Thanks!

sunfishcode:

PR #951, opened 2023-11-28
Remove linux-raw-sys and libc from public API

notgull:
This prevents libc and linux-raw-sys from affecting the semver status of this crate.


sunfishcode:
Cool, this is great. This is an API break, so this will need to go in rustix 1.0.  There isn't yet any timeline for rustiix 1.0&mdash;feedback is welcome&mdash;and until we have one, would you mind resubmitting this PR to the [1.0-staging branch](https://github.com/bytecodealliance/rustix/tree/1.0-staging) that I've now created for this purpose?

One type that seems missing here is `Stat`, which is currently an alias for `libc::stat` on most platforms. We should ideally define our own `Stat` struct, and my rough idea was that we could use the assert macros in src/check_types.rs for checking that our layout matches the libc layout so that we can cast `*mut libc::stat` to `*mut Stat`, but I'm open to other approaches.

It would also be good to have asserts/tests somewhere checking that things like pid_t/uid_t/gid_t/etc. are not larger than the types we're using for them.

PR #954, opened 2023-11-30
fix(net): Fix the WinError 10014

ZengGengSen:
When sendto_v4 and sendto_v6 are called,
the error WinError 10014 is raised, caused
by passing the wrong sockaddr_len

Note: WSAEFAULT(10014), Bad address.
    The system detected an invalid pointer
    address in attempting to use a pointer
    argument of a call. This error occurs
    if an application passes an invalid
    pointer value, or if the length of the
    buffer is too small. For instance, if
    the length of an argument, which is a
    sockaddr structure, is smaller than
    the sizeof(sockaddr).


sunfishcode:
Thanks!

Would you be able to add a testcase that shows the failure here?

Or, do you have any guess as to why the net_v4_sendto and net_v6_sendto tests in tests/net/connect_bind_send.rs don't catch this?

sunfishcode:
I've now figured out why the existing tests don't fail. They were only testing on stream sockets, where the `sendto` destination address is ignored. I've now added tests using datagram sockets, and can confirm that they fail without this patch, and pass with it.

sunfishcode:
This is now released in rustix 0.38.26.

PR #956, opened 2023-12-01
Remove linux-raw-sys and libc from public API

notgull:
This prevents libc and linux-raw-sys from affecting the semver status of this crate.

Version of #951 targeting the `1.0-staging` branch


sunfishcode:
The CI has a few errors here, but I think it makes sense to merge this and iterate from here.

PR #970, opened 2023-12-09
Pass null as empty slices to getxattr on macOS

### LINKED ISSUE #957: Passing an empty slice to lgetxattr on macOS returns ERANGE no matter what
In https://github.com/Stebalien/xattr/pull/41, I discovered a strange bug. The `xattr` crate allocates a buffer for the attribute by first calling `Xgetxattr` with an empty buffer first to get the total buffer size, and then calling it with the full buffer after. However, on macOS there is a bug with `lgetxattr`, where when an empty slice is passed in (as opposed to a null pointer), it returns an `ERANGE` error that we can't handle.

This is probably a macOS bug that we might want to work around in this crate. The workaround for this is to pass a null pointer into `lgetxattr` when an empty slice is passed in, like below:

https://github.com/Stebalien/xattr/blob/32735c53b6e183be44dfbe7780704be4f0d32e03/src/sys/linux_macos.rs#L103-L125

notgull:
Closes #957

### DIFF
--- a/src/backend/libc/fs/syscalls.rs
+++ b/src/backend/libc/fs/syscalls.rs
@@ -2274,7 +2274,6 @@ pub(crate) fn lgetxattr(path: &CStr, name: &CStr, value: &mut [u8]) -> io::Resul
     #[cfg(apple)]
     {
         // Passing an empty to slice to getxattr leads to ERANGE on macOS. Pass null instead.
-        // Fixes https://github.com/bytecodealliance/rustix/issues/957
         let ptr = if value.is_empty() {
             core::ptr::null_mut()
         } else {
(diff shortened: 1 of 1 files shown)


src/backend/libc/fs/syscalls.rs
```diff
@@ -2272,15 +2272,25 @@ pub(crate) fn lgetxattr(path: &CStr, name: &CStr, value: &mut [u8]) -> io::Resul
...
-            path.as_ptr(),
-            name.as_ptr(),
-            value_ptr.cast::<c::c_void>(),
-            value.len(),
-            0,
-            c::XATTR_NOFOLLOW,
-        ))
+    {
+        // Passing an empty to slice to getxattr leads to ERANGE on macOS. Pass null instead.
+        // Fixes https://github.com/bytecodealliance/rustix/issues/957
```
sunfishcode, line 2277: I have a preference for self-sufficient comments, where feasible. In this case, I believe the first line of the comment is entirely sufficient to explain what this code is doing, so the second line with the URL isn't needed.

sunfishcode:
Thanks!

sunfishcode:

PR #972, opened 2023-12-10
Add `name_to_index()` & `index_to_name()`

arctic-alpaca:
This adds the `name_to_index()` and `index_to_name()` methods. They are basically wrappers around ioctl calls described [here](https://man7.org/linux/man-pages/man7/netdevice.7.html). The relevant code in glibc can be found [here for `name_to_index()`](https://github.com/bminor/glibc/blob/master/sysdeps/unix/sysv/linux/if_index.c#L33) and [here for `index_to_name()`](https://github.com/bminor/glibc/blob/master/sysdeps/unix/sysv/linux/if_index.c#L217).

The tests pass locally but I'm not sure whether `lo` and `/sys/class/net/lo/ifindex` are universal enough to consider them acceptable for tests.

### DIFF
--- a/src/backend/libc/net/netdevice.rs
+++ b/src/backend/libc/net/netdevice.rs
@@ -1,0 +1,52 @@
+#![allow(unsafe_code)]
+
+#[cfg(feature = "alloc")]
+use crate::alloc::string::String;
+use crate::backend::io::syscalls::ioctl;
+use crate::fd::AsFd;
+use crate::io;
+#[cfg(feature = "alloc")]
+use libc::SIOCGIFNAME;
+use libc::{__c_anonymous_ifr_ifru, c_char, ifreq, IFNAMSIZ, SIOCGIFINDEX};
+
+pub(crate) fn name_to_index(fd: impl AsFd, if_name: &str) -> io::Result<u32> {
+    let if_name_bytes = if_name.as_bytes();
+    if if_name_bytes.len() >= IFNAMSIZ as usize {
+        return Err(io::Errno::NODEV);
+    }
+
+    let mut ifreq = ifreq {
+        ifr_name: [0; 16],
+        ifr_ifru: __c_anonymous_ifr_ifru { ifru_ifindex: 0 },
+    };
+
+    let mut if_name_c_char_iter = if_name_bytes.iter().map(|byte| *byte as c_char);
+    ifreq.ifr_name[..if_name_bytes.len()].fill_with(|| if_name_c_char_iter.next().unwrap());
+
+    unsafe { ioctl(fd.as_fd(), SIOCGIFINDEX as _, &mut ifreq as *mut ifreq as _) }?;
+    let index = unsafe { ifreq.ifr_ifru.ifru_ifindex };
+    Ok(index as u32)
+}
+
+#[cfg(feature = "alloc")]
+pub(crate) fn index_to_name(fd: impl AsFd, index: u32) -> io::Result<String> {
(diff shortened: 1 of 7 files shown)


src/net/mod.rs
```diff
@@ -16,6 +16,8 @@ mod types;
 #[cfg(windows)]
 mod wsa;
 
+#[cfg(target_os = "linux")]
```
sunfishcode, line 19: Is this API known to be unavailable on Android? If not, would it make sense to change this from `target_os = "linux"` to `linux_kernel`, which includes "linux" and "android"?

src/net/netdevice.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[cfg(feature = "alloc")]
+use crate::alloc::string::String;
+use crate::fd::OwnedFd;
+use crate::io;
+use crate::io::Errno;
+use crate::net::{socket, AddressFamily, SocketType};
+
+/// Creates a socket used to communicate with the kernel in the ioctl calls.
+#[cfg(target_os = "linux")]
```
sunfishcode, line 16: Since this whole module is conditional on the target OS, it's not necessary to add `cfg`s like this to individual items.

src/backend/libc/net/netdevice.rs
```diff
@@ -0,0 +1,56 @@
...
+#[cfg(all(target_os = "linux", feature = "alloc"))]
+pub(crate) fn index_to_name(index: u32) -> io::Result<String> {
+    let mut ifreq = ifreq {
+        ifr_name: [0; 16],
+        ifr_ifru: __c_anonymous_ifr_ifru {
+            ifru_ifindex: index as _,
+        },
+    };
+
+    let fd = open_socket()?;
```
sunfishcode, line 43: It's unusual from rustix's perspective that this is creating a new socket, rather than having the user pass in a socket. It's an extra system call, which may fail, or which users may want more control over. Would it be ok to change the public API here to have an `fd` parameter, with documentation telling users how to create the `fd` if needed, rather than having this `open_socket` function call?

src/net/mod.rs
arctic-alpaca: I looked into the bionic source and it seems to be available. I changed the cfgs accordingly.

src/net/netdevice.rs
arctic-alpaca: Done.

src/backend/libc/net/netdevice.rs
arctic-alpaca: I changed the API accordingly and added docs what glibc and Android do and what the manual specifies.

arctic-alpaca:
Android support is blocked on https://github.com/rust-lang/libc/pull/3487 and its release in libc.

arctic-alpaca:
[Nightly added a new dead code warning](https://github.com/rust-lang/rust/pull/118297).

The Dragonfly CI failure seem to stem from [this change](https://github.com/rust-lang/libc/pull/3509) in libc. According to the docs linked by [`rustix::time::clock_settime`](https://docs.rs/rustix/latest/rustix/time/fn.clock_settime.html) and [`rustix::time::clock_gettime`](https://docs.rs/rustix/latest/rustix/time/fn.clock_gettime.html), Dragonfly does not support `CLOCK_BOOTTIME`. How should that be fixed?

arctic-alpaca:
Thanks! I removed the dead code commit since you also already handled that. With CI being green, this PR should hopefully be done now.

PR #525, opened 2023-01-23
Add ioctl_getflags and ioctl_setflags

ETKNeil:
Add the IOCTL get/setflags with the underlyng IFlags bitflag.

Resources: https://man7.org/linux/man-pages/man2/ioctl_iflags.2.html

To be supported, this need that libc merge this https://github.com/rust-lang/libc/pull/3089

PR #973, opened 2023-12-15
Add ioctl_getflags and ioctl_setflags

fsavy-tehtris:
This is an updated version of https://github.com/bytecodealliance/rustix/pull/525 because the original author will not revisit it.

I'm the author of the associated PR adding the missing constants in `libc` (https://github.com/rust-lang/libc/pull/3396) and I was waiting on the [libc 0.2.151 release](https://github.com/rust-lang/libc/releases/tag/0.2.151).

### DIFF
--- a/src/backend/libc/io/syscalls.rs
+++ b/src/backend/libc/io/syscalls.rs
@@ -4,8 +4,6 @@ use crate::backend::c;
 #[cfg(not(target_os = "wasi"))]
 use crate::backend::conv::ret_discarded_fd;
 use crate::backend::conv::{borrowed_fd, ret, ret_c_int, ret_owned_fd, ret_usize};
-#[cfg(linux_kernel)]
-use crate::backend::io::types::IFlags;
 use crate::fd::{AsFd, BorrowedFd, OwnedFd, RawFd};
 #[cfg(not(any(
     target_os = "aix",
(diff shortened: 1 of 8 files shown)


sunfishcode:
The CI failure here looks unrelated.

src/backend/libc/io/syscalls.rs
```diff
@@ -4,6 +4,8 @@ use crate::backend::c;
 #[cfg(not(target_os = "wasi"))]
 use crate::backend::conv::ret_discarded_fd;
 use crate::backend::conv::{borrowed_fd, ret, ret_c_int, ret_owned_fd, ret_usize};
+#[cfg(any(target_os = "android", target_os = "linux"))]
```
sunfishcode, line 7: Here and elsewhere, rustix has a macro providing a shorthand for `any(target_os = "android", target_os = "linux")`, which is `linux_kernel`.

src/io/ioctl.rs
```diff
@@ -75,3 +78,23 @@ pub fn ioctl_fionread<Fd: AsFd>(fd: Fd) -> io::Result<u64> {
...
+
+/// `ioctl(fd, FS_IOC_SETFLAGS)`—Modify the [inode flags] attributes
+///
+/// [inode flags]: https://man7.org/linux/man-pages/man2/ioctl_iflags.2.html
+#[cfg(any(target_os = "android", target_os = "linux"))]
+#[inline]
+#[doc(alias = "FS_IOC_GETFLAGS")]
+pub fn ioctl_setflags<Fd: AsFd>(fd: Fd, flags: IFlags) -> io::Result<()> {
+    backend::io::syscalls::ioctl_set_flags(fd.as_fd(), flags)
+}
```
sunfishcode, line 100: From my reading of the documentation, it seems these ioctls are specific to files and directories, rather than operating on any file descriptor. That would suggest they go in src/fs/ioctl.rs rather than src/io/ioctl.rs here. Do you think that would be reasonable?

sunfishcode, line 100: Also, since the time when the PR was updated, rustix has switched to a new `ioctl::Ioctl` mechanism; see the `ioctl::Ioctl` impls in src/io/ioctl.rs and src/fs/ioctl.rs for examples. Would you be able to adapt the implementation here to use that new mechanism?

sunfishcode (commented):
Thanks for taking this on!

src/io/ioctl.rs
fsavy-tehtris: Thanks for pointing this out, it makes the implementation much more concise.

I've tried to limit duplication of the `ctl` instantiation but the flags have different types in each backend (`u_long` in `libc`, `u32` in `linux_raw_sys`).

fsavy-tehtris (commented):
Thanks for the feedback! I've addressed your remarks.

src/io/ioctl.rs
fsavy-tehtris: Yes indeed that makes sense, I've moved them.

fsavy-tehtris:
We might want to remove the `IFlags` bitflags duplication too, but I'm not sure where to put them.

sunfishcode:
Could we put `IFlags` in src/fs/ioctl.rs now? It doesn't appear to use different types between the two backends now.

fsavy-tehtris:
I'm moving to another company so I won't be able to revisit this PR as I'll lose access to this account.

sunfishcode:
Thanks for the PR!

sunfishcode:

PR #979, opened 2024-01-05
On 32-bit x86, add more strategies for the syscall mechanism.

sunfishcode:
By default, rustix uses `libc` to read aux records, to get the sysinfo information needed to locate the vsyscall function. If that's disabled, add more fallback paths:

Add a way to use `PR_GET_AUXV` even when "alloc" is not available, by using a statically-sized buffer and being prepared to proceed to the next fallback if the buffer isn't big enough.

And, if `PR_GET_AUXV` and /proc/self/auxv both fail, fall back to using `int 0x80`.


sunfishcode:

PR #985, opened 2024-01-13
Add support for epoll to RedoxOS

notgull:
For now it seems that epoll is the preferred way to access Redox's
underlying polling interface. There are event schemes but until Redox's
syscalls are better integrated with Rustix it's probably not worth it
yet.

cc https://github.com/smol-rs/polling/issues/176


notgull:
macOS failure appears to be spurious

notgull:
@sunfishcode When you have time can you cut a release with this patch?

sunfishcode:
This is now released in rustix 0.38.31.

PR #996, opened 2024-01-17
Return PID from waitpid

### LINKED ISSUE #989: waitpid doesn't returned pid
Is there a reason for this? It should return a `(Pid, WaitStatus)` so you can spawn multiple processes in parallel and wait for them to finish in any order while performing specific actions on a process' completion.

SUPERCILEX:
Fixes #989

### DIFF
--- a/tests/process/wait.rs
+++ b/tests/process/wait.rs
@@ -18,9 +18,10 @@ fn test_waitpid_none() {
         .expect("failed to execute child");
     unsafe { kill(child.id() as _, SIGSTOP) };
 
-    let status = process::waitpid(None, process::WaitOptions::UNTRACED)
+    let (pid, status) = process::waitpid(None, process::WaitOptions::UNTRACED)
         .expect("failed to wait")
         .unwrap();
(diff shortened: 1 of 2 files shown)


tests/process/wait.rs
```diff
@@ -18,7 +18,7 @@ fn test_waitpid_none() {
         .expect("failed to execute child");
     unsafe { kill(child.id() as _, SIGSTOP) };
 
-    let status = process::waitpid(None, process::WaitOptions::UNTRACED)
+    let (_, status) = process::waitpid(None, process::WaitOptions::UNTRACED)
```
sunfishcode, line 21: Could you add an assert checking that the returned pid matches `child.id()`?

sunfishcode (commented):
Looks good, just one minor comment:

sunfishcode:
Also, those CI errors look unrelated, so don't worry about those; I'll fix those separately.

tests/process/wait.rs
SUPERCILEX: Done.

sunfishcode:
Thanks!

PR #999, opened 2024-01-17
Fix handling of negative timestamps in `Stat`.

sunfishcode:
Deprecate `Stat::st_mtime` etc., because they're unsigned and should be signed, and introduce a `StatExt` extension trait which adds `mtime()` etc. accessor functions which return the values in the appropriate signed type.

This `StatExt` trait can go away next time we have a semver breaking change, but for now this preserves compatibility.

### DIFF
--- a/src/backend/libc/fs/syscalls.rs
+++ b/src/backend/libc/fs/syscalls.rs
@@ -1778,6 +1778,7 @@ pub(crate) fn sendfile(
 
 /// Convert from a Linux `statx` value to rustix's `Stat`.
 #[cfg(all(linux_kernel, target_pointer_width = "32"))]
+#[allow(deprecated)] // for `st_[amc]time` u64->i64 transition
 fn statx_to_stat(x: crate::fs::Statx) -> io::Result<Stat> {
     Ok(Stat {
         st_dev: crate::fs::makedev(x.stx_dev_major, x.stx_dev_minor).into(),
@@ -1828,6 +1829,7 @@ fn statx_to_stat(x: crate::fs::Statx) -> io::Result<Stat> {
 
 /// Convert from a Linux `stat64` value to rustix's `Stat`.
 #[cfg(all(linux_kernel, target_pointer_width = "32"))]
+#[allow(deprecated)] // for `st_[amc]time` u64->i64 transition
 fn stat64_to_stat(s64: c::stat64) -> io::Result<Stat> {
     Ok(Stat {
         st_dev: s64.st_dev.try_into().map_err(|_| io::Errno::OVERFLOW)?,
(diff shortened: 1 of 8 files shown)


joshtriplett:
Why a `StatExt` extension trait rather than implementing these methods directly on the structure?

sunfishcode:
Depending on the target and configuration, `Stat` is sometimes just a type alias for `libc::stat` or `linux_raw_sys::general::stat`, and sometimes a custom type. When it's a type alias, we can't define methods on it.

This is one of the things in progress for the next major version of rustix, to move to encapsulate all the libc/linux-raw-sys types, as having them in the public API causes a variety of problems. But that requires a breaking change too, so for now, an extension trait seems the best we can do.

joshtriplett:
@sunfishcode Ah, I didn't realize it was ever directly exposing the underlying type. Fair enough.

This looks like a sensible interim fix, then.

sunfishcode:
This is now released in rustix 0.38.31.

Byron:
Thanks again for tackling this issue!

I was trying to use these improvements but had trouble doing so. The code [looks like this](https://github.com/Byron/gitoxide/blob/bf49cd4508433dba6aacf963f49662516b3a960c/gix-index/src/fs.rs#L55-L72) (*reproduced here for convenience*)

```Rust
            #[cfg(not(target_os = "aix"))]
            let seconds = self.0.st_mtime;
            #[cfg(target_os = "aix")]
            let seconds = self.0.st_mtim.tv_sec;

            // The fix I'd like to live without
            let seconds = seconds as i64;
            system_time_from_secs_nanos(seconds, nanoseconds.try_into().ok()?)
```

`self.0` is of type `rustix::fs::Stat`. However, the trait implementation of `rustix::fs::StatExt` doesn't seem to provide [an implementation for `aix`](https://github.com/bytecodealliance/rustix/pull/999/files#diff-9f7a9943fff7afdd02920889c8f84454a8968e2f8015fd1ae06742255f82fe85R165). 

The solution seems to be to add such an implementation, but maybe there was a reason to not implement it in the first place.

In any way, without `StatExt` being implemented for all platforms I am probably better off leaving the `gitoxide` code as is, but with guidance I should be able to contribute what might be missing.

This also gave me the idea to use `StatExt` to taper over other code I had to write, but would rather not:

```Rust
            #[cfg(not(any(target_os = "netbsd", target_os = "aix")))]
            let nanoseconds = self.0.st_mtime_nsec;
            #[cfg(target_os = "netbsd")]
            let nanoseconds = self.0.st_mtimensec;
            #[cfg(target_os = "aix")]
            let nanoseconds = self.0.st_mtim.tv_nsec;
```

Thanks for letting me know how to proceed here.

sthibaul:
In https://github.com/bytecodealliance/rustix/pull/1064/files I added the implementation

PR #1002, opened 2024-01-24
Add `mount_setattr`

### LINKED ISSUE #975: Add `mount_setattr`
The new mount api makes it possible to atomically create read-only binds mount together with `mount_setattr`.

https://www.man7.org/linux/man-pages/man2/mount_setattr.2.html

Relates to #744

yujincheng08:
Preliminary implementation of `mount_setattr`.

A few notes:
1. I leave the `size` non-exported. Should we export this?
2. I intentionally place `mount_setattr` in `misc.rs` because I am going to add [statmount](https://github.com/torvalds/linux/commit/46eae99ef73302f9fb3dddcd67c374b3dffe8fd6) and [listmount](https://github.com/torvalds/linux/commit/b4c2bea8ceaa50cd42a8f73667389d801a3ecf2d) syscall once 6.8 releases and `libc` adds the syscall number.
3. CI test requires rust-lang/libc#3558

Fix #975

### DIFF
--- a/src/backend/linux_raw/mount/types.rs
+++ b/src/backend/linux_raw/mount/types.rs
@@ -1,4 +1,5 @@
 use crate::backend::c;
+use crate::fd::BorrowedFd;
 use bitflags::bitflags;
 
 bitflags! {
@@ -147,9 +148,10 @@ pub(crate) enum FsConfigCmd {
 
 #[cfg(feature = "mount")]
 bitflags! {
-    /// `MOUNT_ATTR_*` constants for use with [`fsmount`].
+    /// `MOUNT_ATTR_*` constants for use with [`fsmount`, `mount_setattr`].
     ///
     /// [`fsmount`]: crate::mount::fsmount
+    /// [`mount_setattr`]: crate::mount::mount_setattr
     #[repr(transparent)]
     #[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
     pub struct MountAttrFlags: c::c_uint {
@@ -330,3 +332,13 @@ bitflags! {
 
 #[repr(transparent)]
 pub(crate) struct MountFlagsArg(pub(crate) c::c_uint);
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+#[allow(missing_docs)]
+pub struct MountAttr<'a> {
+    pub attr_set: MountAttrFlags,
+    pub attr_clr: MountAttrFlags,
+    pub propagation: MountPropagationFlags,
+    pub userns_fd: BorrowedFd<'a>,
+}
--- a/src/mount/misc.rs
+++ b/src/mount/misc.rs
@@ -1,0 +1,24 @@
+//! Miscellaneous mount APIs
+
+use crate::backend::mount::types::MountAttr;
+use crate::fd::BorrowedFd;
+use crate::fs::AtFlags;
+use crate::{backend, io, path};
+
+/// `mount_setattr(dir_fd, path, flags, mount_attr)`
+///
+/// # References
+///  - [Linux]
+///
+/// [Linux]: https://man7.org/linux/man-pages/man2/mount_setattr.2.html
+#[inline]
+pub fn mount_setattr<Path: path::Arg>(
+    dir_fd: BorrowedFd<'_>,
+    path: Path,
+    flags: AtFlags,
+    mount_attr: &MountAttr<'_>,
+) -> io::Result<()> {
+    path.into_with_c_str(|path| {
+        backend::mount::syscalls::mount_setattr(dir_fd, path, flags, mount_attr)
+    })
+}
--- a/src/backend/libc/mount/syscalls.rs
+++ b/src/backend/libc/mount/syscalls.rs
@@ -270,3 +270,31 @@ pub(crate) fn fsconfig_reconfigure(fs_fd: BorrowedFd<'_>) -> io::Result<()> {
         ))
     }
 }
+
+#[cfg(linux_kernel)]
+#[cfg(feature = "mount")]
+pub(crate) fn mount_setattr(
+    dir_fd: BorrowedFd<'_>,
+    path: &CStr,
+    flags: crate::fs::AtFlags,
+    mount_attr: &crate::mount::MountAttr<'_>,
+) -> io::Result<()> {
+    syscall! {
+        fn mount_setattr(
+            dir_fd: c::c_int,
+            fs_name: *const c::c_char,
+            flags: c::c_uint,
+            mount_attr: *const crate::mount::MountAttr<'_>,
+            size: usize
+        ) via SYS_mount_setattr -> c::c_int
+    }
+    unsafe {
+        ret(mount_setattr(
+            borrowed_fd(dir_fd),
+            c_str(path),
+            flags.bits(),
+            mount_attr as *const _,
+            core::mem::size_of::<crate::mount::MountAttr<'_>>(),
+        ))
+    }
+}
--- a/src/backend/libc/mount/types.rs
+++ b/src/backend/libc/mount/types.rs
@@ -1,4 +1,5 @@
 use crate::backend::c;
+use crate::fd::BorrowedFd;
 use bitflags::bitflags;
 
 #[cfg(linux_kernel)]
@@ -150,9 +151,10 @@ pub(crate) enum FsConfigCmd {
 #[cfg(feature = "mount")]
 #[cfg(linux_kernel)]
 bitflags! {
-    /// `MOUNT_ATTR_*` constants for use with [`fsmount`].
+    /// `MOUNT_ATTR_*` constants for use with [`fsmount`, `mount_setattr`].
     ///
     /// [`fsmount`]: crate::mount::fsmount
+    /// [`mount_setattr`]: crate::mount::mount_setattr
     #[repr(transparent)]
     #[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
     pub struct MountAttrFlags: c::c_uint {
@@ -338,3 +340,13 @@ bitflags! {
 
 #[cfg(linux_kernel)]
 pub(crate) struct MountFlagsArg(pub(crate) c::c_ulong);
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+#[allow(missing_docs)]
+pub struct MountAttr<'a> {
+    pub attr_set: MountAttrFlags,
+    pub attr_clr: MountAttrFlags,
+    pub propagation: MountPropagationFlags,
+    pub userns_fd: BorrowedFd<'a>,
+}
--- a/src/backend/linux_raw/mount/syscalls.rs
+++ b/src/backend/linux_raw/mount/syscalls.rs
@@ -237,3 +237,23 @@ pub(crate) fn fsconfig_reconfigure(fs_fd: BorrowedFd<'_>) -> io::Result<()> {
         ))
     }
 }
+
+#[cfg(feature = "mount")]
+#[inline]
+pub(crate) fn mount_setattr(
+    dir_fd: BorrowedFd<'_>,
+    path: &CStr,
+    flags: crate::fs::AtFlags,
+    mount_attr: &crate::mount::MountAttr<'_>,
+) -> io::Result<()> {
+    unsafe {
+        ret(syscall_readonly!(
+            __NR_mount_setattr,
+            dir_fd,
+            path,
+            flags,
+            mount_attr as *const crate::mount::MountAttr<'_>,
+            crate::backend::conv::size_of::<crate::mount::MountAttr<'_>, _>()
+        ))
+    }
+}
--- a/src/mount/mod.rs
+++ b/src/mount/mod.rs
@@ -10,10 +10,12 @@
 
 #[cfg(feature = "mount")]
 mod fsopen;
+mod misc;
 mod mount_unmount;
 mod types;
 
 #[cfg(feature = "mount")]
 pub use fsopen::*;
+pub use misc::*;
 pub use mount_unmount::*;
 pub use types::*;


src/backend/linux_raw/mount/types.rs
```diff
@@ -330,3 +332,13 @@ bitflags! {
 
 #[repr(transparent)]
 pub(crate) struct MountFlagsArg(pub(crate) c::c_uint);
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+#[allow(missing_docs)]
+pub struct MountAttr<'a> {
+    pub attr_set: MountAttrFlags,
```
sunfishcode, line 340: According to Linux's documentation, `attr_set` is a `__u64`, while `MountAttrFlags` is currently a `c_uint`. Could you add a test testing that the layout matches, following one of the "layouts" tests in the tree?

src/mount/misc.rs
```diff
@@ -0,0 +1,24 @@
+//! Miscellaneous mount APIs
+
+use crate::backend::mount::types::MountAttr;
+use crate::fd::BorrowedFd;
+use crate::fs::AtFlags;
+use crate::{backend, io, path};
+
+/// `mount_setattr(dir_fd, path, flags, mount_attr)`
```
sunfishcode, line 8: Since there is no `mount_setattr` declaration exposed in C, could you change this comment to `syscall(SYS_mount_setattr, ...)`?

sunfishcode:
> I leave the `size` non-exported. Should we export this?

If you don't have a need for it at this time, then we don't need to. If someone later has a need for it, we can add another function that exposes it.

src/backend/linux_raw/mount/types.rs
```diff
@@ -330,3 +332,13 @@ bitflags! {
...
+    pub attr_clr: MountAttrFlags,
+    pub propagation: MountPropagationFlags,
+    pub userns_fd: BorrowedFd<'a>,
```
ayosec, line 343: `userns_fd` field is expected as a `__u64`, but `BorrowedFd` is 4 bytes. Maybe it is safer to declare a private struct with the exact same fields of the C API, and translate the `MountAttr` fields to it before calling `mount_setattr`.

```rust
#[repr(C)]
pub(crate) struct mount_attr {
    pub attr_set: u64,
    pub attr_clr: u64,
    pub propagation: u64,
    pub userns_fd: u64,
}
```

---

Also, `userns_fd` is needed only when `attr_set` includes `MOUNT_ATTR_IDMAP`.  The field could be declared as an `Option<BorrowedFd<'a>>`, and send it as `-EBADFD` to the syscall if it is `None`.

src/mount/misc.rs
ayosec: glibc added `mount_setattr` in 2.36 ([commit](https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=af1aa36c617d72e80ff6b268bf927b6642d90823)), though the `libc` crate has no support for it.

In my system, the declaration is visible in `/usr/include/x86_64-linux-gnu/sys/mount.h`

```c
/* Change the mount properties of the mount or an entire mount tree.  If
   PATH is a relative pathname, then it is interpreted relative to the
   directory referred to by the file descriptor dirfd.  Otherwise if DFD is
   the special value AT_FDCWD then PATH is interpreted relative to the current
   working directory of the calling process.  */
extern int mount_setattr (int __dfd, const char *__path, unsigned int __flags,
                          struct mount_attr *__uattr, size_t __usize)
  __THROW;
```

ayosec (commented):
I would like to help to finish this pull-request, in case @yujincheng08 can't continue it.

src/backend/linux_raw/mount/types.rs
```diff
@@ -330,3 +332,13 @@ bitflags! {
 
 #[repr(transparent)]
 pub(crate) struct MountFlagsArg(pub(crate) c::c_uint);
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+#[allow(missing_docs)]
+pub struct MountAttr<'a> {
```
rusty-snake, line 339: This struct should be non_exhaustive.

src/backend/linux_raw/mount/types.rs
rusty-snake: > Maybe it is safer to declare a private struct with the exact same fields of the C API, and translate the MountAttr fields to it before calling mount_setattr.

FWIW: https://codeberg.org/crabjail/lnix/src/commit/092defd573bab18f32c466dd7f774a03236babb6/src/mount/mountfd.rs#L453-L488

ayosec: I think that the way `mount_attr` is defined in `lnix` is better than exposing the fields directly.

The usage would be something like this:

```rust
mount_setattr(
    dir_fd,
    c"", 
    MountSetattrFlags::RECURSIVE | MountSetattrFlags::EMPTY_PATH,
    MountAttr::new().set_flags(MountAttrFlags::RDONLY),
)?;
```

@sunfishcode Do you want to take the same approach from `lnix`?

I can implement the changes if needed.

src/mount/misc.rs
sunfishcode: Ok, cool. As long as it's in glibc, for the purposes of the documentation comment here, that's good enough.

src/backend/linux_raw/mount/types.rs
sunfishcode: How is `userns_fd` exposed in the lnix API?

In general, I'm in favor of encapsulating these fields like this, provided we can do so without restricting useful functionality.

ayosec: > How is `userns_fd` exposed in the lnix API?

It seems that it does not support setting values to `userns_fd` (there is a [`TODO: ID-mapped mounts`](https://codeberg.org/crabjail/lnix/src/commit/2d7123794846848421b79b29b1784a1e84c4d620/src/mount/mountfd.rs#L482), and `userns_fd` only appears in the struct definition).

`mount_setattr()` will not close the file descriptor. I guess that `MountAttr` should take the ownership of the file descriptor, so it can guarantee that the descriptor is valid when `mount_setattr` is called, and closed when `MountAttr` is not needed anymore. 

Maybe something like this:

```rust
pub struct MountAttr {
    mount_attr: sys::mount_attr,
    userns_fd: Option<OwnedFd>, 
}

impl MountAttr {

    pub fn set_userns_fd(&mut self, userns_fd: OwnedFd) -> &mut Self {
        self.mount_attr.userns_fd = userns_fd.as_raw_fd() as u64;
        self.userns_fd = Some(userns_fd);
        self
    }

    // ...
```

rusty-snake: > I guess that MountAttr should take the ownership of the file descriptor, so it can guarantee that the descriptor is valid when mount_setattr is called, and closed when MountAttr is not needed anymore.

A BorrowedFd would work too.

While `BorrowedFd`/`OwnedFd` are ffi safe, this does not help us here because we need a u64.

You should first find answers to the following questions.
- Is a lifetime parameter on `MountAttr` ok?
- Where on the raw-opinionated scalar should rustix' implementation be.
  - `userns_fd` is ignored unless `set_flags` contains `MOUNT_ATTR_IDMAP`.
- Do you want a zero-copy `repr(C)` struct or a copy `repr(Rust)` struct?

If a lifetime is ok, you could use a phantom.

```rust
#[repr(C)]
struct mount_attr<'fd> {
    attr_set: u64, // MountAttrFlags is repr(transparent) but not u64.
    attr_clr: u64,
    propagation: u64,
    userns_fd: u64,
    _userns_fd_phantom: PhantomData<BorrowedFd<'fd>>,
}
```

Otherwise, you can also take ownership.

```rust
#[repr(C)]
struct mount_attr<'fd> {
    attr_set: u64, // MountAttrFlags is repr(transparent) but not u64.
    attr_clr: u64,
    propagation: u64,
    userns_fd: u64,
    _userns_fd: Option<OwnedFd>,
}
```

ayosec: > A `BorrowedFd` would work too.

I didn't it consider because it requires a change in the lifetime, but  now I think it makes more sense to use `BorrowedFd` instead of an `OwnedFd`.

I think that the main issue with `BorrowedFd` is how to call the functions to set the values (in `lnix`: `set_flags`, `clear_flags`, and `propagation`). Those functions receive and return a `&mut self`. However, `userns_fd` must use the lifetime of its parameter, so it returns a new type.

For example, we could define `MountAttr` like this:

```rust
#[repr(C)]
#[allow(non_camel_case_types)]
#[derive(Default, Debug)]
struct mount_attr {
    attr_set: u64,
    attr_clr: u64,
    propagation: u64,
    userns_fd: u64,
}

#[derive(Debug)]
pub struct MountAttr<'a> {
    attr: mount_attr,
    userns_fd: PhantomData<BorrowedFd<'a>>,
}
```

Then, the default constructor would return a `'static` lifetime:

```rust
impl MountAttr<'static> {
    pub fn new() -> Self {
        MountAttr {
            attr: mount_attr::default(),
            userns_fd: PhantomData,
        }
    }
}
```

But for `userns_fd` it must return the lifetime of the `BorrowedFd`, so it can't return `Self`. The implementation could be something like this:

```rust
impl<'a> MountAttr<'a> {

    pub fn userns_fd<'b>(self, userns_fd: BorrowedFd<'b>) -> MountAttr<'b> {
        MountAttr {
            attr: mount_attr {
                userns_fd: userns_fd.as_raw_fd() as u64,
                ..self.attr
            },
            userns_fd: PhantomData,
        }
    }
```

This is important because it is inconsistent with the two ways to build an instance of `MountAttr`:

```rust
// This two blocks are identical:

foo(MountAttr::new().set(MountAttrFlags::MOUNT_ATTR_FOO));


let mut attr = MountAttr::new();
attr.set(MountAttrFlags::MOUNT_ATTR_FOO);
foo(&attr);


// But these blocks are different:

foo(MountAttr::new()
    .set(MountAttrFlags::MOUNT_ATTR_IDMAP)
    .userns_fd(fd.as_fd()));


let mut attr = MountAttr::new();
attr.set(MountAttrFlags::MOUNT_ATTR_IDMAP); // Updates `attr`.
attr.userns_fd(fd.as_fd());                 // Returns a new instance.
foo(&attr);                                 // Call `foo()` with no `userns_fd`.
```

Maybe, a way to avoid the possible confussion is to take the ownership of `MountAttr` to build its value, and always return an owned instance (instead of a reference).

```rust
impl<'a> MountAttr<'a> {
    pub fn set(mut self, flags: MountAttrFlags) -> Self {
        self.attr.attr_set = flags.bits() as u64;
        self
    }

    pub fn clear(mut self, flags: MountAttrFlags) -> Self {
        self.attr.attr_clr = flags.bits() as u64;
        self
    }

    pub fn propagation(mut self, propagation: MountPropagationFlags) -> Self {
        self.attr.propagation = propagation.bits() as u64;
        self
    }

    pub fn userns_fd<'b>(self, userns_fd: BorrowedFd<'b>) -> MountAttr<'b> {
        MountAttr {
            attr: mount_attr {
                userns_fd: userns_fd.as_raw_fd() as u64,
                ..self.attr
            },
            userns_fd: PhantomData,
        }
    }
}
```

I made a simple [proof of concept](https://gist.github.com/ayosec/94ee445c8c686c58d73060f4c2a5d2bb) with this idea.

rusty-snake: > Where on the raw-opinionated scalar should rustix' implementation be. 

- `mount_setattr(MountAttr)`
- `mount_setattr_idmap(MountAttr, Fd)`

ayosec: > * `mount_setattr(MountAttr)`
> * `mount_setattr_idmap(MountAttr, Fd)`

If we add specific functions for common cases, we could add specific constructors for `MountAttr`.

The [`rustix::mount` module](https://docs.rs/rustix/0.38.36/rustix/mount/index.html) has multiple `mount_*` functions for common use-cases of `mount(2)`, like [`mount_bind`](https://docs.rs/rustix/0.38.36/rustix/mount/fn.mount_bind.html) or [`mount_change`](https://docs.rs/rustix/0.38.36/rustix/mount/fn.mount_change.html), so it makes sense to support specific cases for `mount_setattr`.

I checked some projects using `mount_setattr(2)`:

* When an ID-mapped mount is created, the most common usage of `mount_setattr` is with just `attr_set = MOUNT_ATTR_IDMAP`:
    * In [composefs](https://github.com/containers/composefs), [`lcfs-mount.c:lcfs_mount_erofs`](https://github.com/containers/composefs/blob/47042e810aeac438f187ab678f4ec14a44a74e0f/libcomposefs/lcfs-mount.c#L554-L560).
    * In [containerd](https://github.com/containerd/containerd), [`mount_idmapped_linux.go:IDMapMount`](https://github.com/containerd/containerd/blob/e013322eed0524019f60577666ea74a304c0f38d/core/mount/mount_idmapped_linux.go#L70-L83).
    * In [lxc](https://github.com/lxc/lxc), [`conf.c:lxc_idmapped_mounts_parent`](https://github.com/lxc/lxc/blob/bad6d6f180da45b2d3ae06d44201197df38504d7/src/lxc/conf.c#L3657-L3662).
    * In [runc](https://github.com/opencontainers/runc), [`mount_linux.go:mountFd`](https://github.com/opencontainers/runc/blob/961b8031f6614be300d3b10eec9d2f54323026d9/libcontainer/mount_linux.go#L257-L260).
    * In [systemd](https://github.com/systemd/systemd), [`mount-util.c:remount_idmap_fd`](https://github.com/systemd/systemd/blob/6c8e9ba2f0adb7a66fe520a3f58f36d69bc8db43/src/shared/mount-util.c#L1423-L1427).
    * In [util-linux](https://github.com/util-linux/util-linux), [`hook_idmap.c:hook_mount_post`](https://github.com/util-linux/util-linux/blob/a3bf59a6f497a9314d4671ddd481235e2eecbbec/libmount/src/hook_idmap.c#L302-L344).

* `attr_clr` seems to be very uncommon.

    The only use-case I was able to find is to change the access-time setting, since it is a requirement of the syscall. From [`mount_setattr(2)`](https://www.man7.org/linux/man-pages/man2/mount_setattr.2.html):

    > Note that, since the access-time values are an enumeration rather than bit values, a caller wanting to  transition to a different access-time setting cannot simply specify the access-time setting in `attr_set`, but must also include `MOUNT_ATTR__ATIME` in the `attr_clr` field. The kernel will verify that `MOUNT_ATTR__ATIME` isn't partially set in `attr_clr` \[...]


* Both `attr_set` and `propagation` are used either alone or together.

I also checked other types in rustix that contain file descriptors:

* [`rustix::process::PrctlMmMap`](https://docs.rs/rustix/0.38.36/rustix/process/struct.PrctlMmMap.html#structfield.exe_fd) has a `pub exe_fd: u32` field.

    The struct is used in [`rustix::process::configure_virtual_memory_map`](https://docs.rs/rustix/0.38.36/rustix/process/fn.configure_virtual_memory_map.html), which is marked as _unsafe_.

* [`rustix::io_uring::io_uring_params`](https://docs.rs/rustix/0.38.36/rustix/io_uring/struct.io_uring_params.html) has a `pub wq_fd: u32` field, which is used when `io_uring_setup` is called with `IORING_SETUP_ATTACH_WQ`.

    Unlike the previous function, [`rustix::io_uring::io_uring_setup`](https://docs.rs/rustix/0.38.36/rustix/io_uring/fn.io_uring_setup.html) is not marked as _unsafe_.

* [`rustix::event::PollFd`](https://docs.rs/rustix/0.38.36/rustix/event/struct.PollFd.html) contains a `BorrowedFd` field, which is private.

    This struct seems to be close to what is needed for `MountAttr`.

To keep consistency, we could define `MountAttr` similar to `PollFd`: keep all fields private, and provide constructors from common cases. For example:

```rust
pub struct MountAttr<'fd> {
    attr: c::mount_attr,
    userns_fd: PhantomData<BorrowedFd<'fd>>,
}

impl<'fd> MountAttr<'fd> {
    pub fn new<Fd: AsFd>(
        set: MountAttrFlags,
        clear: MountAttrFlags,
        propagation: MountAttrPropagation,
        userns_fd: Option<&'fd Fd>,
    ) -> Self {
        // If there is no `userns_fd`, use `fs::CWD` instead of `0`, so the
        // kernel will reply with `EBADF` if `IDMAP` is set.
        let userns_fd = userns_fd.map(|fd| fd.as_fd()).unwrap_or(fs::CWD);

        // ...
    }

    pub fn new_with_idmap<Fd: AsFd>(fd: &'fd Fd) -> Self {
        new(
            MountAttrFlags::IDMAP,
            MountAttrFlags::empty(),
            MountAttrPropagation::empty(),
            Some(fd),
        )
    }

    // ...
}

impl MountAttr<'static> {
    pub fn new_with_set(set: MountAttrFlags) -> Self {
        new(
            set,
            MountAttrFlags::empty(),
            MountAttrPropagation::empty(),
            None,
        )
    }

    pub fn new_with_propagation(propagation: MountAttrPropagation) -> Self {
        new(
            MountAttrFlags::empty(),
            MountAttrFlags::empty(),
            propagation,
            None,
        )
    }
}
```

allisonkarlitskaya: It seems to me like the current `struct`-based approach was done as an implementation detail at the syscall level to allow extensibility in the future.  I wouldn't see anything wrong with a simple implementation of this API using only (an admittedly large number of) function parameters: dirfd, pathname, flags, set, clear, propagation, userns.

We could even imagine specialized cases similar to what is done for other mount APIs (like how the mount syscall is exposed also as mount_remount, or there are many typesafe fsconfig variants).

rusty-snake: > It seems to me like the current struct-based approach was done as an implementation detail at the syscall level to **allow extensibility in the future**. I wouldn't see anything wrong with a simple implementation of this API using only (an admittedly large number of) function parameters: dirfd, pathname, flags, set, clear, propagation, userns.

Well this would mean rustix needs a semver bump or a new function if the kernel adds a new field.

> We could even imagine specialized cases similar to what is done for other mount APIs (like how the mount syscall is exposed also as mount_remount, or there are many typesafe fsconfig variants).

There is one difference, the mount functions you listed have an multiplexed interface that got demultiplexed by rustix. `mount_setattr` isn't a multiplexed syscall (in that form) so splitting would mean you need multiple syscalls instead of one.

---

Summarizing my thought, we should first decided whether we want extensibility in the future (w/o breaking) or not.

allisonkarlitskaya: I see it as a matter of pragmatism: this would be an easy way to resolve this issue in the short term...

Should the API ever change, then there would be a lot of options at that point, including adding a complex API that takes a struct...

ayosec: > It seems to me like the current `struct`-based approach was done as an implementation detail at the syscall level to allow extensibility in the future.

An advantage of using a struct is that it is easier to omit default fields with dedicated builder functions. For example, with the draft in <https://github.com/bytecodealliance/rustix/pull/1002#discussion_r1747931183>, an ID-mapped mount could be created with:


```rust
use MountSetAttrFlags as Flags;

mount_setattr(
    dir_fd,
    c"", 
    Flags::RECURSIVE | Flags::EMPTY_PATH,
    MountAttr::new_with_idmap(userns),
)?;
```

rusty-snake: - `openat2` has a future extensibility struct that rustix decided to inline into the function signature.
- `clone3` has a future extensibility struct but is impossible to be implemented by rustix.
- `mount_setattr`, `sched_setattr`, `landlock_create_ruleset`, `statmount`/`listmount`, `perf_event_open` have a future extensibility struct and are not yet implemented by rustix.

So we are not dealing with a single syscall but also with the design future syscalls in rustix might follow.

cyphar: Having been one of the people who pushed for this newer syscall design in Linux ([I even gave a talk about it a few years ago](https://www.youtube.com/watch?v=ggD-eb3yPVs)), in my view there are two Rust-y ways of supporting this kind of extensibility:

 1. Require a builder pattern for `copy_struct_from_user` APIs -- this is The Rust Way(TM) of doing things but it doesn't work for `copy_struct_to_user` APIs and the proliferation of getters and setters feels out of place in a low-level library like `rustix` (IMHO).
 2. Expose the structure with `#[non_exhaustive]` (and `Default`) and accept an `AsRef` of the structure -- this does make some parts of the API less ergonomic (it really sucks you can't do `{ ..Default::default() }` for `#[non_exhaustive]`-tagged structs) but usage would end up just looking like a bit like C usage, which isn't too bad. `rustix` would set the structure size automatically though.

In the PR for `openat2` support, I was advocating for doing it this way but sadly we ended up with the non-extensible API in `rustix` (as a result, in my own crates I continue to use my own syscall wrappers). To be fair, I have not ended up adding as many extensions to `openat2` as I planned initially, so you could argue that the `rustix` approach was more pragmatic (though I am working on them again now).

However, APIs like `mount_setattr` have seen very important extensions added post-merge (id-mapped mounts being the main example) and it would be a grave mistake to not take that into consideration when adding wrappers for other APIs.

Unfortunately, APIs like `statmount` involve self-referential pointers so I'm not sure what exactly the API would look like in Rust (irrespective of the extensible struct thing).

livingsilver94: I was writing my own attempt at adding support for `mount_setattr()` when I stumbled upon this conversation.

My proposal is to define a `MountAttr` struct that wraps `struct mount_attr`.
We'll also define the `MountAttrSimpleFlags` bitmap for flags without arguments, plus `MountAttrParamFlags` for flags with arguments, like `MOUNT_ATTR_IDMAP`.
This `MountAttrParamFlags` will be of course tightly coupled with `MountAttr`, but I don't see it as a problem as it'll be only meaningful when used together with the latter.

This proposal doesn't take into consideration the lifetime of the file descriptor yet, but I'm open to suggestions unless I'll manage to propose my own first.

Any future extension in the C API *should* be hidden by the `MountAttrParamFlags::apply` function. That is the part we'll adapt.

Below I removed some docstrings and macros for brevity.

```rust
bitflags! {
    /// All `MOUNT_ATTR_*` flags that don't carry a parameter.
    pub struct MountAttrSimpleFlags: u64 {
        const MOUNT_ATTR_RDONLY =  c::MOUNT_ATTR_RDONLY;
        const MOUNT_ATTR_NOSUID = c::MOUNT_ATTR_NOSUID;
        const MOUNT_ATTR_NODEV = c::MOUNT_ATTR_NODEV;
        const MOUNT_ATTR_NOEXEC = c::MOUNT_ATTR_NOEXEC;
        const MOUNT_ATTR__ATIME = c::MOUNT_ATTR__ATIME;
        const MOUNT_ATTR_RELATIME = c::MOUNT_ATTR_RELATIME;
        const MOUNT_ATTR_NOATIME = c::MOUNT_ATTR_NOATIME;
        const MOUNT_ATTR_STRICTATIME =c::MOUNT_ATTR_STRICTATIME;
        const MOUNT_ATTR_NODIRATIME = c::MOUNT_ATTR_NODIRATIME;
        const MOUNT_ATTR_NOSYMFOLLOW = c::MOUNT_ATTR_NOSYMFOLLOW;
        const _ = !0;
    }
}

/// `MOUNT_ATTR_*` flags that also carry a parameter.
pub enum MountAttrParamFlags {
    /// `MOUNT_ATTR_IDMAP`, which carries a user namespace file descriptor.
    IdMap(u64),
}

impl MountAttrParamFlags {
    fn apply(self, mount_attr: &mut MountAttr, operation: MountAttrOperation) {
        let flags = match operation {
            MountAttrOperation::Set => &mut mount_attr.0.attr_set,
            MountAttrOperation::Clear => &mut mount_attr.0.attr_clr,
        };
        match self {
            MountAttrParamFlags::IdMap(userns_fd) => {
                *flags |= MountAttrSimpleFlags::from_bits_retain(c::MOUNT_ATTR_IDMAP).bits();
                mount_attr.0.userns_fd = userns_fd;
            }
        }
    }
}

/// `struct mount_attr`
pub struct MountAttr(pub(crate) c::mount_attr);

impl MountAttr {
    /// Create a `MountAttr` with the given simple fields.
    pub fn new(
        attr_set: MountAttrSimpleFlags,
        attr_clr: MountAttrSimpleFlags,
        propagation: MountPropagationFlags,
    ) -> Self {
        Self(c::mount_attr {
            attr_set: attr_set.bits(),
            attr_clr: attr_clr.bits(),
            propagation: propagation.bits(),
            userns_fd: 0,
        })
    }

    /// Create a zero-initialized `MountAttr`.
    pub fn zeroed() -> Self {
        Self(c::mount_attr {
            attr_set: 0,
            attr_clr: 0,
            propagation: 0,
            userns_fd: 0,
        })
    }

    /// Set a parameterized flag.
    pub fn set_param_flag(mut self, param: MountAttrParamFlags) -> Self {
        param.apply(&mut self, MountAttrOperation::Set);
        self
    }

    /// Clear a parameterized flag.
    pub fn clear_param_flag(mut self, param: MountAttrParamFlags) -> Self {
        param.apply(&mut self, MountAttrOperation::Clear);
        self
    }
}

enum MountAttrOperation {
    Set,
    Clear,
}
```

rusty-snake: No matter how the API looks like, it has to be an rustic API:

> rustix provides efficient ***memory-safe and I/O-safe*** wrappers to POSIX-like, Unix-like, Linux, and Winsock syscall-like APIs, with configurable backends. It uses Rust references, slices, and return values instead of raw pointers, and ***I/O safety types instead of raw file descriptors***, providing memory safety, I/O safety, and provenance. It uses Results for reporting errors, ***bitflags instead of bare integer flags***, an Arg trait with optimizations to efficiently accept any Rust string type, and several other efficient conveniences.

Issue #918, opened 2023-11-03
Support abritrary socket adresses


niklaswimmer:
Rustix can currently not be used to bind sockets which are not unix or ip. The only real requirement on a socket address is that it starts with an u16 (the address family) - judging from the types used in the definition of the bind syscall.

My specific use case is binding bluetooth sockets and my current workaround is to just call [`libc::bind`](https://docs.rs/libc/0.2.149/libc/fn.bind.html) instead (using the fd returned by rustix's `socket_with`).

I am unsure what a good API would look like, but ideally it would not require me to construct a `[u8]` array myself and instead work with structs which are then automatically converted into byte arrays when it comes to invoking the syscall. For my use case I wrote [a helper module](https://github.com/niklaswimmer/raw-bluetooth-mgmt/blob/main/src/address.rs) that uses unions to convert some struct into a [libc::sockaddr](https://docs.rs/libc/0.2.149/libc/struct.sockaddr.html) if you want to take a look.

This issue also blocks usage of [eyra](https://docs.rs/eyra) for any programs that need to work with some non-well-known sockets.

I would be happy to help implementing this API, but as noted above I am not sure about the design and would require some input.

sunfishcode:
Interesting question.

I think my preference here is to add a `Bluetooth` arm to `SocketAddrAny` for bluetooth, containing a `SocketAddrBluetooth`, similar to the `Unix` arm, which encapsulates a bluetooth address, and then teach the encode/decode logic to handle it.

That said, at a quick search, I found about 40 `sockaddr_*` types in Linux's headers. I don't have any idea how commonly used they all are. If rustix is going to end up supporting a lot of them, it may make sense to switch over to just exposing arbitrary bytes and letting users or other crates sort it out.

However, my guess is, for now, it makes sense to stick with typed interfaces and adding address families such as `SocketAddrBluetooth` when people have a need for them, and we'll see how far we get.

Does that make sense?

sunfishcode:
The purpose of having `connect_v4`/`connect_v6`/`connect_unix` instead of just `connect`/`connect_any` is just to avoid the construction of a `SocketAddr`/`SocketAddrAny` in common cases. Compared to V4/V6/Unix, I assume all other socket address types are far less commonly used, so it seems ok to just add entries to `SocketAddrAny` and use `connect_any`, and not add new specialized `connect_*` etc. routines. Does that sound feasible?

nbdd0121:
[...]

PR #1004, opened 2024-01-29
Add trait to reduce duplicated code between socket address types, implement netlink addresses

kevinmehall:
https://github.com/bytecodealliance/rustix/issues/918

This adds a trait `SockAddr`, and implements it for `SocketAddrV4`, `SocketAddrV6`, `SocketAddrUnix`, `SocketAddrXdp`, `SocketAddr`,  `SocketAddrAny`, and the newly-added `SocketAddrNetlink`. The syscalls that take a `sockaddr` are reimplemented in terms of this trait, removing hundreds of lines of duplicated code. It also simplifies the public interface, allowing `rustix::net::{bind, connect, sendto, sendmsg_addr}` to take an `(addr: &impl SockAddr)`, rather than needing separate functions `bind_v4`, `bind_v6`, `bind_unix`, `bind_xdp`, `bind_any` etc. 

The `SocketAddrAny` enum is replaced by a struct wrapping `SocketAddrStorage`.  This allows `SocketAddrAny` be fully generic and allows syscalls returning addresses to be used with any address type not known to rustix. Address types use `.into()` / `.try_into()` to convert to / from `SocketAddrAny`.


sunfishcode:
As a quick update here, I am looking forward to reviewing this; I've just been busy.

src/net/socket_address.rs
```diff
@@ -0,0 +1,72 @@
...
+/// A trait abstracting over the types that can be passed as a `sockaddr`.
+///
+/// Safety: by implementing this trait, you assert that the values returned
+/// by the trait methods can be passed to the system calls that accept `sockaddr`.
+pub unsafe trait SocketAddress {
+    /// The corresponding C `sockaddr_*` type.
+    type CSockAddr;
+
+    /// Convert to the C type.
+    fn encode(&self) -> Self::CSockAddr;
```
niklaswimmer, line 18: I find it a bit annoying/confusing that this trait uses default fns. We could instead utilize a small custom enum that stores either a reference to or an owned `CSockAddr`. Kinda like a `Cow` but for nostd, something along the lines of

```rust
/// Borrowed Or oWned
pub enum Bow<'a, B> {
    /// Stores a reference to the value.
    Borrowed(&'a B),
    /// Owns the value.
    Owned(B),
}

impl<'a, B> Bow<'a, B> {
    /// Get a reference to the value stored in this Bow.
    pub fn as_ref(&self) -> &B {
        match self {
            Bow::Borrowed(val) => val,
            Bow::Owned(val) => &val,
        }
    }
}
```

Then only `encode` would be enough and the rest could be defined as helper functions or similar.

Let me know what you think

src/backend/linux_raw/net/mod.rs
```diff
@@ -6,4 +6,3 @@ pub(crate) mod read_sockaddr;
 pub(crate) mod send_recv;
 pub(crate) mod sockopt;
 pub(crate) mod syscalls;
-pub(crate) mod write_sockaddr;
```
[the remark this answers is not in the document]
kevinmehall: Good call, replaced the duplicated implementations with ones shared between linux_raw and libc that call into the existing `write_sockaddr` functions.

src/net/socket_address.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// the memory pointed to by `storage`, and returns the number of
+    /// bytes used.
+    ///
+    /// # Safety
+    ///
+    /// `storage` must point to valid memory for encoding the socket
+    /// address.
+    unsafe fn write_sockaddr(&self, storage: *mut SocketAddrStorage) -> usize {
+        let encoded = self.encode();
+        core::ptr::write(storage.cast(), encoded);
```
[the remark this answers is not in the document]
kevinmehall: Removed this method entirely, but fixed an `unsafe_fn_in_fn` elsewhere.

src/net/socket_address.rs
kevinmehall: I removed the `write_sockaddr` method, which had a default implementation that was not overridden for any of the types. It was only used in one place, for `SocketAddrAny::write`.

The purpose of the `with_sockaddr` method is to avoid the copying that you mention, and is overridden in the implementation for `SocketAddrUnix` which contains a `sockaddr_un`.

Before simplifying this trait further, I'd like to hear what @sunfishcode thinks about the `CSockAddr` associated type. If it's [undesirable to expose the underlying C types](https://github.com/bytecodealliance/rustix/pull/956), I think the way to go would be to remove `CSockAddr` and `encode`, leaving only `with_sockaddr`. This is a messier interface than simply returning a value, but a private helper function could be used to implement this in terms of the `encode_sockaddr_*` functions for the majority of the address types.

kevinmehall:
Given that `sockaddr` was included on the list of libc types to remove in https://github.com/bytecodealliance/rustix/pull/1063, I slimmed down the `SockAddr` trait further to only `with_sockaddr` and made the raw pointer type an opaque struct to avoid all libc types in the API.

I also rebased to clean up the messy history for easier review. Full history archived [here](https://github.com/bytecodealliance/rustix/compare/main...kevinmehall:rustix:socket-addr-trait-history).

al8n:
Hi, any updates on this PR? Looking forward `rustix` supporting netlink.

sunfishcode:
Thanks for the pings here, and your patience! I'll make an effort to review this before 1.0.

src/net/socket_address.rs
```diff
@@ -0,0 +1,81 @@
...
+/// `sockaddr`-compatible C type, and then cast back to a `sockaddr` pointer
+/// to be passed to a system call.
+#[repr(C)]
+pub struct SockAddrRaw {
+    _data: [u8; 0],
+}
+
+/// A trait abstracting over the types that can be passed as a `sockaddr`.
+///
+/// Safety: implementers of this trait must ensure that `with_sockaddr` calls
```
sunfishcode, line 26: As a minor style nit, this should be a `# Safety` section header.

src/net/socket_address.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// Opaque type equivalent to `sockaddr` in C.
+///
+/// This is always used behind a raw pointer that is cast from a pointer to a
+/// `sockaddr`-compatible C type, and then cast back to a `sockaddr` pointer
+/// to be passed to a system call.
+#[repr(C)]
+pub struct SockAddrRaw {
+    _data: [u8; 0],
+}
```
sunfishcode, line 22: What would you think of renaming `SockAddrRaw` to `SocketAddrOpaque`?

There are two changes there:
 - `Sock` -> `Socket`, because if we have some things prefixed `Socket` and some prefixed `Sock` I expect I'll frequently forget which things use which prefixes.
 - `Raw -> `Opaque`, which I think better captures that this struct type doesn't describe the data layout.

src/net/mod.rs
```diff
@@ -10,6 +10,7 @@
 mod send_recv;
 mod socket;
 mod socket_addr_any;
+mod socket_address;
```
sunfishcode, line 13: Could we make this a `pub mod`, and perhaps name it `addr`, to keep `SockAddr` and `SockAddrRaw` from cluttering the rest of the `net` API? It seems like most users won't ever need to use those directly.

src/net/socket_address.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// A trait abstracting over the types that can be passed as a `sockaddr`.
+///
+/// Safety: implementers of this trait must ensure that `with_sockaddr` calls
+/// `f` with a pointer that is readable for the passed length, and points
+/// to data that is a valid socket address for the system calls that accept
+/// `sockaddr` as a const pointer.
+pub unsafe trait SockAddr {
```
sunfishcode, line 30: And, what would you think of renaming this to `SocketAddrArg`? `Sock` -> `Socket` for the same reason as above, and `Arg` since it is for describing arguments; return values will need to use some other mechanism as you noted in the top comment.

sunfishcode (commented):
This looks really nice!

I was initially worried about the safety of the `f` called by the `with_sockaddr` function, however I now see that the safety requirements are covered by the `pub unsafe trait`, so it looks good.

And I was initially worried about generic functions in src/backend, to keep code duplication down, however they're in fact replacing manually duplicated code, so that doesn't seem to be a problem.

I agree that it'd be good to continue to evolve this as you discussed. https://github.com/bytecodealliance/rustix/pull/1171 adds a `RawSocketAddr` type which is very close to what you describe for return values, and sounds like a good general direction.

I have have a few relatively minor comments, plus this needs a rebase. If you'd like help with the rebase please let me know.

src/net/mod.rs
kevinmehall: Is it confusing that `rustix::net::addr` then only contains those two, and not any of the other address types (`SocketAddrUnix`, `SocketAddrV4`, `SocketAddrV6`, `SocketAddr`, `SocketAddrAny`, `SocketAddrStorage`) that are one level up? Seems like at least `SocketAddrStorage` and `SocketAddrOpaque` should be next to each other since they represent similar C types.

Maybe move them all into `rustix::net::addr`?

sunfishcode: Good idea to keep `SocketAddrStorage` next to `SocketAddrOpaque`.

It seems nice to keep `SocketAddrV4`, `SocketAddrV6`, `SocketAddrUnix`, in `rustix::net` where all the rest of the API that uses them are though, because that's the "regular" API.

What if we put `SocketAddrOpaque`, `SocketAddrArg`, and `SocketAddrStorage` into a `pub mod` named `rustix::net::abstract`?

kevinmehall: Sounds good. `abstract` might be confused with the abstract namespace for unix sockets, but I don't have a better name to propose.

kevinmehall: Also, `abstract` is a Rust reserved keyword.

kevinmehall:
Rebased and made the requested changes. I named the new module `rustix::net::addr` since `abstract` is a reserved keyword, and I moved `SocketAddrStorage` into it. It could use a better name if we can come up with one.

I'll take a look at updating `SocketAddrAny` with https://github.com/bytecodealliance/rustix/pull/1171's  `RawSocketAddr`.

Since it looks like you're merging breaking changes to main for 1.0, should I go ahead and remove the address type specific functions?
* `bind_any`, `bind_unix`, `bind_v4`, `bind_v6`, `bind_xdp` in favor of `bind`,
* `connect_any`, `connect_unix`, `connect_v4`, `connect_v6` in favor of `connect` (leaving address-less `connect_unspec`)
* `sendmsg_v4`, `sendmsg_v6`, `sendmsg_unix`, `sendmsg_xdp`, `sendmsg_any` in favor of `sendmsg_addr` (leaving address-less `sendmsg`)
* `sendto_any`, `sendto_v4`, `sendto_v6`, `sendto_unix`, `sendto_xdp` in favor of `sendto`

sunfishcode:
> Rebased and made the requested changes. I named the new module `rustix::net::addr` since `abstract` is a reserved keyword, and I moved `SocketAddrStorage` into it. It could use a better name if we can come up with one.

I haven't thought of a better name yet, but I'll keep thinking :smile: .

> I'll take a look at updating `SocketAddrAny` with #1171 `RawSocketAddr`.

:+1: 

> Since it looks like you're merging breaking changes to main for 1.0, should I go ahead and remove the address type specific functions?
> 
>     * `bind_any`, `bind_unix`, `bind_v4`, `bind_v6`, `bind_xdp` in favor of `bind`,
> 
>     * `connect_any`, `connect_unix`, `connect_v4`, `connect_v6` in favor of `connect` (leaving address-less `connect_unspec`)
> 
>     * `sendmsg_v4`, `sendmsg_v6`, `sendmsg_unix`, `sendmsg_xdp`, `sendmsg_any` in favor of `sendmsg_addr` (leaving address-less `sendmsg`)
> 
>     * `sendto_any`, `sendto_v4`, `sendto_v6`, `sendto_unix`, `sendto_xdp` in favor of `sendto`

Yes, that'd be great!

kevinmehall:
Removed the address-type-specific variants of `bind`, `connect`, `sendmsg`, and `sendto`, and changed `SocketAddrAny` from an enum to a struct wrapping `SocketAddrStorage`.

src/backend/libc/net/read_sockaddr.rs
```diff
@@ -97,260 +99,160 @@ pub(crate) unsafe fn read_sa_family(storage: *const c::sockaddr) -> u16 {
...
-                Ipv6Addr::from(in6_addr_s6_addr(decode.sin6_addr)),
-                u16::from_be(decode.sin6_port),
-                u32::from_be(decode.sin6_flowinfo),
-                sockaddr_in6_sin6_scope_id(decode),
-            ))
+    if len < offsetof_sun_path {
+        return Err(Errno::INVAL);
+    }
+    if len == offsetof_sun_path {
+        SocketAddrUnix::new(&[][..])
```
sunfishcode, line 186: I'm looking to figure out all the ways that `TryFrom<SocketAddrAny>` can fail, and `SocketAddrUnix::new` here can fail with `NAMETOOLONG`. However the only time it fails is if the path is longer than `sun_path.len()`, which we've already checked here, so I think that can't happen. Would it make sense to make this do `Ok(SocketAddrUnix::new(&[][...]).unwrap())`, or add a new infallible `SocketAddrUnix` constructor?

And similar for the `SocketAddrUnix::new_abstract_name` and `SocketAddrUnix::new` calls below?

src/net/socket_addr_any.rs
```diff
@@ -1,122 +1,263 @@
...
-    Unix(SocketAddrUnix),
-    /// `struct sockaddr_xdp`
-    #[cfg(target_os = "linux")]
-    Xdp(SocketAddrXdp),
+/// A type that can hold any kind of socket address, as a safe abstraction for
+/// `sockaddr_storage`.
+///
+/// Socket addresses can be converted to `SocketAddrAny` via the [`From`] and
+/// [`Into`] traits. `SocketAddrAny` can be converted back to a specific socket
+/// address type with [`TryFrom`] and [`TryInto`].
```
sunfishcode, line 23: Could you document here that the `TryFrom` functions fail with `Errno::AFNOSUPPORT` if the address family is not the requested one?

An unfortunate thing about using `Errno` is that it doesn't describe which errors could actually occur. However, it might be worth it here, because a convenient thing about using `Errno` is that these functions will usually be called from within functions that return `Result<..., Errno>`, so users can just do `?` on the result and it'll do something reasonable.

Looking at the code, it looks like these `TryFrom<SocketAddrAny>` functions can fail with:
 - `AFNOSUPPORT` if the address family isn't the requested one
 -  `INVAL` if the length is wrong, though perhaps those should be panics?
 - `NAMETOOLONG` if it's Unix-domain and the name doesn't fit in `sun_path`, but that shouldn't be possible

sunfishcode (commented):
Cool, overall this looks great!

src/backend/libc/net/sockopt.rs
```diff
@@ -832,11 +829,8 @@ pub(crate) fn ipv6_original_dst(fd: BorrowedFd<'_>) -> io::Result<SocketAddrV6>
 
     getsockopt_raw(fd, level, optname, &mut value, &mut optlen)?;
 
-    let any = unsafe { SocketAddrAny::read(value.as_ptr(), optlen as usize)? };
-    match any {
-        SocketAddrAny::V6(v6) => Ok(v6),
-        _ => unreachable!(),
-    }
+    let any = unsafe { SocketAddrAny::read(value.as_ptr().cast(), optlen as usize) };
```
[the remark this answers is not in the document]
kevinmehall: I tried several different ways to construct the `SocketAddrAny`,  looking at the generated optimized assembly for a function that calls `getsockname` and returns `SocketAddrAny`, to see how the value is returned, but I wasn't able to get full return value optimization.

Previous version of this PR:
```rs
let mut addrlen = core::mem::size_of::<sockaddr_storage>() as socklen_t;
let mut storage = MaybeUninit::<sockaddr_storage>::uninit();
ret(syscall!(
    __NR_getsockname,
    fd,
    &mut storage,
    by_mut(&mut addrlen)
))?;
Ok(SocketAddrAny::read(
    storage.as_ptr().cast(),
    addrlen.try_into().unwrap(),
))
```
-> Compiles to a call to memcpy with variable length. The `copy_nonoverlapping` lets the optimizer see that only `len` bytes of the buffer matter, and can ignore the rest even if it were potentially written by the syscall.

`MaybeUninit::<SocketAddrAny>`:
```rs
let mut addr = MaybeUninit::<SocketAddrAny>::uninit();
let mut addrlen = core::mem::size_of::<SocketAddrStorage>() as socklen_t;
ret(syscall!(
    __NR_getsockname,
    fd,
    &raw mut (*addr.as_mut_ptr()).storage,
    by_mut(&mut addrlen)
))?;
*(&raw mut (*addr.as_mut_ptr()).len) = NonZeroUsize::new_unchecked(addrlen as _);
Ok(addr.assume_init())
```
-> call to memcpy with fixed length of 136

That doesn't avoid the memcpy, but rather makes it larger. The compiler can no longer see that we don't care about the uninitialized tail of the buffer.

I realized that since the callers all have a `MaybeUninit<SocketAddrStorage>`, which is the same as stored in `SocketAddrAny`, a constructor `pub unsafe fn new(storage: MaybeUninit<SocketAddrStorage>, len: usize) -> Self` could be cleaner than the `read` signature with a raw pointer:

```rs
let mut addrlen = core::mem::size_of::<SocketAddrStorage>() as SocketAddrLen;
let mut storage = MaybeUninit::<SocketAddrStorage>::uninit();
ret(syscall!(
    __NR_getsockname,
    fd,
    &mut storage,
    by_mut(&mut addrlen)
))?;
Ok(SocketAddrAny::new(storage, addrlen as _))
```
-> This resulted in an inlined memcpy of 128 bytes (8x pairs of `movups`)

I tested replacing the syscall with a fake to see if there was anything besides the syscall itself preventing return value optimization:
```rs
let mut storage = MaybeUninit::<SocketAddrStorage>::uninit();
let addrlen = 4;
storage
    .as_mut_ptr()
    .cast::<u8>()
    .copy_from_nonoverlapping([1, 2, 3, 4].as_ptr(), 4);
Ok(SocketAddrAny::new(storage, addrlen))
```
-> This wrote directly to the return pointer.

Maybe since the pointer address is exposed to the inline asm, it's not allowed to use a different address and prevents return value optimization?

To clean up the boilerplate creation of the `addr` and `addrlen` variables, I wrapped the two temporaries into a new `SocketAddrBuf` struct that is basically a `SocketAddrAny` without invariants. `SocketAddrBuf::new()` initially sets its length to `size_of::<SocketAddrStorage>()` as is passed into the syscall:

```rs
let mut addr = SocketAddrBuf::new();
ret(syscall!(
    __NR_getsockname,
    fd,
    &mut addr.storage,
    by_mut(&mut addr.len)
))?;
Ok(addr.into_any())
```
-> codegen is about the same (inlined memcpy)

`SocketAddrBuf` seemed more like an internal helper for the syscalls than a public API, so I kept it `pub(crate)`. I'm not sure if the inlined 128 byte memcpy or the call to memcpy with (usually shorter) variable length would ultimately perform better, but with `SocketAddrBuf` that can be updated in one place instead of at every call site.

The good news is that after adding a few `#[inline]`s, `SocketAddrV4::try_from(rustix::net::getsockname(fd)?)` does inline the two functions together to read from the stack and convert without any large copies, so it's not necessarily worse than the status quo for anything that consumes the address in the same function rather than returning `SocketAddrAny`.

src/net/socket_addr_any.rs
```diff
@@ -1,122 +1,263 @@
...
+    ///
+    /// # Panics
+    ///
+    /// if `len` is smaller than the sockaddr header or larger than `SocketAddrStorage`.
+    ///
+    /// # Safety
+    ///
+    /// `storage` must point to valid memory for decoding a socket
+    /// address. `len` bytes must be initialized.
+    pub unsafe fn read(src: *const SocketAddrOpaque, len: usize) -> Self {
```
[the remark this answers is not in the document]
kevinmehall: Changed the internal callers to construct it through a new `SocketAddrBuf` that handles this. Not sure if the new public constructor `pub unsafe fn new(storage: MaybeUninit<SocketAddrStorage>, len: usize) -> Self` should take `usize` or `socklen_t` -- I lean towards `usize` as the convention in Rust, and to avoid needing to re-export `socklen_t` (`SocketAddrLen`?)

src/net/socket_addr_any.rs
```diff
@@ -1,122 +1,263 @@
...
+    ///
+    /// # Panics
+    ///
+    /// if `len` is nonzero but smaller than the sockaddr header or larger than `SocketAddrStorage`.
+    ///
+    /// # Safety
+    ///
+    /// `storage` must point to valid memory for decoding a socket
+    /// address. `len` bytes must be initialized.
+    pub unsafe fn read_option(src: *const SocketAddrOpaque, len: usize) -> Option<Self> {
```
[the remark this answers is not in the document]
kevinmehall: Removed the public method in favor of an internal API that handles the length in a different way, see above.

src/net/socket_addr_any.rs
```diff
@@ -1,122 +1,263 @@
...
+    /// # Panics
+    ///
+    /// if `len` is smaller than the sockaddr header or larger than `SocketAddrStorage`.
+    ///
+    /// # Safety
+    ///
+    /// `storage` must point to valid memory for decoding a socket
+    /// address. `len` bytes must be initialized.
+    pub unsafe fn read(src: *const SocketAddrOpaque, len: usize) -> Self {
+        assert!(len >= core::mem::size_of::<backend::c::sa_family_t>());
```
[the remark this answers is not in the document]
kevinmehall: Replaced it with `size_of::<read_sockaddr::sockaddr_header>()` which accounts for that. This check exists in order to make sure we can read the address family, and `read_sa_family`  dereferences it as `sockaddr_header`.

src/net/addr.rs
```diff
@@ -0,0 +1,83 @@
+//! Socket address types.
```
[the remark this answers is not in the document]
kevinmehall: Updated to "Types for implementers of socket address types or code that is generic over address types."

src/backend/libc/net/read_sockaddr.rs
```diff
@@ -97,260 +99,160 @@ pub(crate) unsafe fn read_sa_family(storage: *const c::sockaddr) -> u16 {
...
+    (*storage.cast::<sockaddr_header>()).sa_family = c::AF_UNSPEC as _;
+}
+
+pub(crate) fn read_sockaddr_v4(addr: &SocketAddrAny) -> Result<SocketAddrV4, Errno> {
+    if addr.address_family() != AddressFamily::INET {
+        return Err(Errno::AFNOSUPPORT);
+    }
+    if addr.len() < size_of::<c::sockaddr_in>() {
+        return Err(Errno::INVAL);
+    }
```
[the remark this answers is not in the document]
kevinmehall: Changed all these to `assert!()`. In theory it's even covered by the `SocketAddrAny::new()` safety invariant requiring a valid address and doesn't need a check, but it's probably worth the assert in case the kernel gives us something unanticipated.

src/backend/libc/net/read_sockaddr.rs
kevinmehall: Since `SocketAddrUnix` holds a `sockaddr_un`, this doesn't really need to parse and reconstruct the address and can just memcpy it.

In reviewing this, I found an edge case: On Linux, a [it's permitted to use 108-character path without a nul-terminator, but it comes back from the kernel with a nul, pushing the length past the size of sockaddr_un](https://man7.org/linux/man-pages/man7/unix.7.html#BUGS). This currently causes a bounds check panic in both backends, unchanged by this PR. Since the OS considers it a valid address, it seems like this should not be rejected here, but then `SocketAddrUnix` would have to handle it. Maybe `SocketAddrUnix::path` should return `&Path` instead of `&CStr` to be able to represent a path with no null terminator, and probably a better type for it anyway.

I don't necessarily want to redesign `SocketAddrUnix` in this PR, though. I turned the initial length check into an `assert!()` but otherwise left the errors from `SocketAddrUnix` for now, which I agree should be unreachable.

src/net/socket_addr_any.rs
kevinmehall: Yeah, I was debating whether it should be `()` or `Errno` but went with `Errno` because some of the tests were matching on the old enum and  returning `AFNOSUPPORT` if it was the wrong address type.

Added documentation here and on the `TryFrom` impls.

src/backend/linux_raw/net/syscalls.rs
```diff
@@ -178,36 +171,28 @@ pub(crate) fn accept_with(fd: BorrowedFd<'_>, flags: SocketFlags) -> io::Result<
...
             __NR_accept,
             fd,
-            &mut storage,
-            by_mut(&mut addrlen)
+            &mut addr.storage,
+            by_mut(&mut addr.len)
         ))?;
-        Ok((
-            fd,
-            maybe_read_sockaddr_os(&storage.assume_init(), addrlen.try_into().unwrap()),
```
kevinmehall, line 191: Also, I noticed that prior to this PR, these `assume_init` calls were all undefined behavior, because the `sockaddr_storage` is not necessarily fully initialized. That's why `SocketAddrAny` contains `MaybeUninit`.

src/backend/libc/net/read_sockaddr.rs
sunfishcode: Wow. I've now filed https://github.com/bytecodealliance/rustix/issues/1316 to track that SocketAddrUnix bug.

src/net/socket_addr_any.rs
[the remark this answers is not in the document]
sunfishcode: The thing that has me leaning toward `socklen_t` is seeing places like [this](https://github.com/bytecodealliance/rustix/pull/1004/files#diff-ef53992d92c529ae3cbce72d22be289c6fb319446a1f7e445dc4d404852264b9R88)  that silently cast the `usize` back to `socklen_t`. Perhaps we could make those `.try_into().unwrap()` to eliminate the (admittedly remote, but perhaps not impossible) chance of silently truncating an address, but it seems cleaner to keep socket lengths in the type that the OS them to us in.

And yeah, I think it would be `SocketAddrLen`. If you like, I can make that change.

sunfishcode:
Thanks! The socklen_t question is my only outstanding question now.

colinmarc:
Should I rebase #1171 onto this?

sunfishcode:
@colinmarc Yes, please do rebase it on top of this.

I've now added the `SocketAddrLen` patch, so this should be ready to land soon.

sunfishcode:
Interesting. It appears this PR breaks code like `bind(s, &"[::1]:0".parse().unwrap())` no longer compiles, because Rust can't infer the type of the `.parse()` due to the `impl SocketAddrArg`. And by quite a coincidence, https://github.com/bytecodealliance/rustix/pull/132 just landed today adding tests which do just that.

I've now changed the tests to avoid doing this. It's a little unfortunate that this common idiom doesn't work, but my guess is it's ok.

sunfishcode:
Thanks!

PR #1005, opened 2024-01-30
Add `init_module` and `finit_module` syscalls

yujincheng08:


sunfishcode:
The CI failures here are unrelated and should be fixed in #1010.

I think the `system` module would be a better fit for these functions than the `runtime` module. `runtime` is for things that require knowledge of libc internals to use correctly. If I understand correctly, `init_module`/`finit_module`/etc. are things that need to be used with care, but don't require any special coordination with libc.

yujincheng08:
@sunfishcode moved to `system`.

sunfishcode:
Thanks!

PR #1006, opened 2024-02-01
Add `setgroups` and fix `getgroups`

yujincheng08:


sunfishcode:
The CI failure here is unrelated.

The Linux setgroups syscall only operates on the current thread, which differs from the POSIX behavior of operating on the entire process, so these functions should go in `thread` rather than `process`. And the libc backend will need to use the syscall! macro rather than libc to achieve the same behavior. See the other `set_*` functions in src/thread/id.rs for examples of this.

yujincheng08:
@sunfishcode I also considered the same thing as I saw the implementation of other `set_*` of `id.rs` but the thing is i) there's no `setgroups` in POSIX; ii) in Linux it's actually setting for the whole process instead of threads.

See the following codes for testing:

```c++
#include <sys/types.h>
#include <thread>
#include <grp.h>
#include <unistd.h>
#include <stdio.h>

int main() {
    std::thread t([]() {
        gid_t gids[] = {1000, 27, 109};
        setgroups(3, gids);
    });
    t.join();
    gid_t gids[10];
    int n = getgroups(10, gids);
    for (int i = 0; i < n; i++) {
        printf("%d\n", gids[i]);
    }
    return 0;
}
```

sunfishcode:
Ah, I was mistaken about POSIX. Still, the Linux libc function sets the groups for all threads, but it appears the syscall only sets the groups for one thread. That means that in the current PR, the linux_raw backend which uses the syscall behaves differently from the libc backend which uses the libc function. Rustix does not currently have the infrastructure for setting the ids on all threads, and in the absence of such infrastructure, so we should provide the single-thread behahvior, in the `thread` module.

yujincheng08:
You are right. The syscall itself is done for the thread only. But glibc uses nptl to set for the whole process.

```c++
int main() {
  std::thread t([]() {
    gid_t gids[] = {1000, 27, 109};
    syscall(SYS_setgroups, 3, gids);
    int n = syscall(SYS_getgroups, 10, gids);
    for (int i = 0; i < n; i++) {
      printf("in thread: %d\n", gids[i]);
    }
  });
  t.join();
  gid_t gids[10];
  int n = syscall(SYS_getgroups, 10, gids);
  for (int i = 0; i < n; i++) {
    printf("in main %d\n", gids[i]);
  }
  return 0;
}
```

yujincheng08:
Modified as requested.

Also, I fixed `getgroups`:

> EINVAL size is less than the number of supplementary group IDs,
>        but is not zero.

> If size is zero, list is not modified, but the total number of supplementary group IDs for the process is returned.  This allows the caller to determine the size of a dynamically allocated list to be used in a further call to getgroups().

yujincheng08:
@sunfishcode Gentle ping.

sunfishcode:
Looks good, thanks!

PR #1008, opened 2024-02-05
Support processing ancillary outside rustix infrastructure

SUPERCILEX:
Specifically, I'm trying to parse control data from io_uring responses.

### DIFF
--- a/src/net/send_recv/msg.rs
+++ b/src/net/send_recv/msg.rs
@@ -440,8 +440,7 @@ impl<'buf> RecvAncillaryBuffer<'buf> {
     pub fn drain(&mut self) -> AncillaryDrain<'_> {
         AncillaryDrain {
             messages: messages::Messages::new(&mut self.buffer[self.read..][..self.length]),
-            read: &mut self.read,
-            length: &mut self.length,
+            read_and_length: Some((&mut self.read, &mut self.length)),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


SUPERCILEX:
Test failures appear unrelated.

notgull:
We had this error on `softbuffer` as well, it can be fixed by pinning `half` to v2.2.1

SUPERCILEX:
@sunfishcode I don't think this requires a semver bump, right? I'm only changing private APIs and adding one new one.

sunfishcode:
It looks like the change to `AncillaryDrain` to combine `read` and `length` into `read_and_lenghth` would be a semver bump.

SUPERCILEX:
Hmmm, maybe I'm missing something, but all of those fields are private.

sunfishcode:
Oh wow you're totally right.

sunfishcode (commented):
Looks good. Two observations:

Looking at it now, it looks like a mistake to have `AncillaryDrain` hold `read` and `length` members. Rust standard library things like `Vec::drain` remove their elements even if the iterator is dropped without being iterated, so we should ideally do the same, and if we did, that would also eliminate this `Option`. However, we can do that in the next semver bump, because it is a subtle behavior change.

In general, it's a goal of `rustix` to provide safe APIs wherever possible, so this unsafe `parse` function is unforuntate, however a safe RecvMsg API for io_uring would be non-trivial, and possibly the best way thing to do within rustix is to just expose the unsafe primitives and look to higher-level crates to provide the safe abstractions on top of them.

PR #1010, opened 2024-02-05
In Rust 1.63 builds, pin the half crate to v2.2.1.

sunfishcode:
half 2.3.1 appears to depend on newer Rust vesions, so in the Rust 1.63 build use half 2.2.1.


sunfishcode:
It appears [Cirrus has dropped FreeBSD 12](https://cirrus-ci.org/guide/FreeBSD/), so this removes that configuration as well.

PR #1011, opened 2024-02-06
Add `io_uring_register` flags

dmitry-salin:
Adds `IoringRegisterFlags` to `rustix::io_uring::io_uring_register`.

```
int io_uring_register(unsigned int fd, unsigned int opcode, void *arg, unsigned int nr_args);
```

`fd` is the file descriptor returned by a call to `io_uring_setup`. If opcode has the flag **IORING_REGISTER_USE_REGISTERED_RING** ored into it, `fd` is instead the index of a registered ring fd.

Resources: [liburing](https://github.com/axboe/liburing/blob/liburing-2.5/src/register.c#L17), [kernel](https://github.com/torvalds/linux/blob/v6.7/io_uring/io_uring.c#L4621)

### DIFF
--- a/src/backend/libc/io_uring/syscalls.rs
+++ b/src/backend/libc/io_uring/syscalls.rs
@@ -4,7 +4,7 @@ use crate::backend::c;
 use crate::backend::conv::{borrowed_fd, ret_owned_fd, ret_u32};
 use crate::fd::{BorrowedFd, OwnedFd};
 use crate::io;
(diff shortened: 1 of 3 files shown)


ryanseipp:
We now have `io_uring_register_with` merged in #1115 that adds support for `IORING_REGISTER_USE_REGISTERED_RING`

sunfishcode:
Yes, `io_uring_register_with` covers this. I'll add a documentation comment to `io_uring_register` mentioning this.

PR #1013, opened 2024-02-07
Add support for Packet MMAP

DBLouis:
This is WIP

A bit of guidance would be greatly appreciated :)

sunfishcode/linux-raw-sys#103


DBLouis:
What is the difference between `#[cfg(linux_kernel)]` and `#[cfg(target_os = "linux"]`? When should I use which?

DBLouis:
I wonder if it is worth it to backport my changes to linux-raw-sys to an earlier new release like 0.4.13 instead of 0.6.4

sunfishcode:
You can submit a PR for 0.4.x if you like, but if you just care about docs.rs, 0.6.x is sufficient.

DBLouis:
> You can submit a PR for 0.4.x if you like, but if you just care about docs.rs, 0.6.x is sufficient.

Sorry I meant the changes in this sunfishcode/linux-raw-sys@783bd07b0253116a04d8b29cfcc05566c4a00fcc

DBLouis:
I have added a example high-level safe "wrapper". I'm thinking that is could be moved to it's own crate (e.g. `rustix-packet-mmap`)

shrey-subcom:
whats going on with this pr? I too am in need of a tpacket_v3 implementation in rustix.

PR #1019, opened 2024-02-16
Add pidfd_send_signal syscall

SUPERCILEX:
I didn't bother supporting a non-null siginfo_t because that will require a re-work of the WaitidStatus API. I think if somebody needs that support they can make the breaking changes necessary.


sunfishcode:
Thanks!

PR #1020, opened 2024-02-17
Expose `OFlags::LARGEFILE`

oxalica:
This flag can be reported by `fcntl_getfl` but is ignored currently. It is also used in FUSE requests. This should also simplify tests, but I don't have `illumos` platforms to test on.

### DIFF
--- a/src/backend/libc/c.rs
+++ b/src/backend/libc/c.rs
@@ -91,6 +91,16 @@ pub(crate) const MSG_DONTWAIT: c_int = libc::MSG_NONBLOCK;
 #[cfg(feature = "net")]
 pub(crate) const SO_NOSIGPIPE: c_int = 0x0800;
 
+// It is defined as 0 in libc under 64-bit platforms, but is automatically set by kernel.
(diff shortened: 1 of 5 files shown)


sunfishcode:
Rustix's current design is to transparently always set `LARGEFILE` internally, rather than have it be a flag that users have to set. Could you say more about what motivates this PR?

oxalica:
> Rustix's current design is to transparently always set `LARGEFILE` internally

The motivation is to define it, so users can use it when they really want (when you are writing a FUSE server that process users' LARGEFILE flag, or if the user is rustix itself), and they can see its existence from `fcntl_getfl` rather than a magic hexadecimal number. Passing them transparently does not mean to hide their existence from all users.

sunfishcode:
Thinking about this more, I think you're right. We should add `OFlags::LARGEFILE`. It'd be good to document that it's not necessary for most users, as rustix functions set it automatically.

oxalica:
> It'd be good to document that it's not necessary for most users, as rustix functions set it automatically.

Added.

sunfishcode:
Thanks!

PR #1021, opened 2024-02-21
Fix fallocate for apple platforms.

c-t-k:
The `fallocate` implementation for apple platforms was not using the `F_ALLOCATEALL` flag during the first call to fcntl. This allows the call to allocate fewer bytes than the passed amount.

It's possible this shim still has different semantics from `posix_fallocate` or linux's `fallocate`, given newer versions of apple's libc also include this flag (the newest man page for `fcntl` seems to indicate that the filesystem is allowed to free unused blocks that were previously allocated during a close and that this flag may be passed as a hint to avoid that):

```
F_ALLOCATEPERSIST  Allocate space that is not freed when close(2) is called. (Note that the file system may ignore this request.)
```

That said, I haven't been able to trigger this scenario on a Macbook Pro M1 running Sonoma with APFS. My guess is this only applies to blocks allocated outside the file length, which the call to `ftruncate` here should prevent.

### DIFF
--- a/src/backend/libc/fs/syscalls.rs
+++ b/src/backend/libc/fs/syscalls.rs
@@ -1662,7 +1662,7 @@ pub(crate) fn fallocate(
 
     let new_len = offset.checked_add(len).ok_or(io::Errno::FBIG)?;
     let mut store = c::fstore_t {
-        fst_flags: c::F_ALLOCATECONTIG,
+        fst_flags: c::F_ALLOCATECONTIG | c::F_ALLOCATEALL,
         fst_posmode: c::F_PEOFPOSMODE,
         fst_offset: 0,
         fst_length: new_len,


sunfishcode:
Thanks for the report! Your analysis sounds reasonable.

Looking at this code now, I'm not sure we need `F_ALLOCATECONTIG` at all. It seems like it would be fine to just let the filesystem driver decide how to allocate the space.

Also, I agree, it sounds like `F_ALLOCATEPERSIST` is something we should be using, to better match the `fallocate` semantics.

Would you mind updating the PR to remove `F_ALLOCATECONTIG` and the extra `fcntl` call, and add `F_ALLOCATEPERSIST`? If not, I can also do it.

c-t-k:
Noticing some other off things about this.

`fst_length` is set to this:

```rust
let new_len = offset.checked_add(len).ok_or(io::Errno::FBIG)?;
```

The man page just says this field sets the length of the preallocated region, so I don't believe it denotes an end offset in the file (and this should initially allocate more blocks than requested by the caller):
```
off_t     fst_length;     /* IN: size of the region */
```

More problematic though is the call to `ftruncate` also uses this length without ensuring the end position is further than the current end of the file. This will cause it to truncate the file if the region ends before the end of the file, which [isn't what `posix_fallocate` does](https://pubs.opengroup.org/onlinepubs/007904975/functions/posix_fallocate.html).

I can also fix the length set for `fst_length` and throw in an `fstat` to ensure we don't truncate the file here if that solution works for you.

PR #1024, opened 2024-02-28
Add mount2 syscall

morr0ne:
Adds a mount2 syscall.

This is actually the same as the current mount syscall. The main difference being the function parameters which are arguably wrong in the original implementation since it did not account for certain values being null.

This pr is incomplete since the parameters have been changed from `Arg ` to `&CStr`. This is not inline with other apis and should be fixed but I could not figure out how to convert the types properly. I temporarily used a different type because I needed to use the function. Help fixing the parameters would be appreciated.

### DIFF
--- a/src/mount/mount_unmount.rs
+++ b/src/mount/mount_unmount.rs
@@ -1,5 +1,7 @@
 //! Linux `mount`.
 
+use std::ffi::CStr;
+
 use crate::backend::mount::types::{
     InternalMountFlags, MountFlags, MountFlagsArg, MountPropagationFlags, UnmountFlags,
 };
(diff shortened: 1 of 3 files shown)


sunfishcode:
To fix the `Arg` types, would it work to add something like this to src/path/arg.rs and then use that in place of `.into_with_c_str`?

```rust
pub fn option_into_with_c_str<T, F, A: Arg>(a: Option<A>, f: F) -> io::Result<T>
where
    A: Sized,
    F: FnOnce(Option<&CStr>) -> io::Result<T>,
{
    if let Some(a) = a {
        a.into_with_c_str(|p| f(Some(p)))
    } else {
        f(None)
    }
}
```

morr0ne:
Should be good to go

morr0ne:
While testing I realized that that data should not be an `Arg`, the data is interpreted differently based on the filesystem being mounted so it can't be generalized to anything "Path-like". I changed it to a CStr witch is more inline with the syscall api

SUPERCILEX:
Sorry for resurrecting an old PR, but I don't think this is correct. When are these parameters null during a mount? @morr0ne I think you misunderstood the API. If you want `source` to be null for example, you should be using https://docs.rs/rustix/latest/rustix/mount/fn.mount_remount.html. Though I guess maybe it's fine to have a "raw" version of the API.

morr0ne:
> .

It has been quite a while since I opened this pr so I do not properly remember what my needs were. I believe this is what I needed this for: https://github.com/morr0ne/woven/blob/48e5e68c736fb7578c58fb847cb4e00dd73aba2e/system/src/raminit.rs#L86

SUPERCILEX:
Wouldn't the empty string work?

morr0ne:
> Wouldn't the empty string work?

It might. I followed examples of how c programs used the api and the manual pages. Both pointed to using null and it does work

SUPERCILEX:
Alrighty, I've submitted https://github.com/bytecodealliance/rustix/pull/1112 to deprecate that API while also allow passing in null for options.

PR #1031, opened 2024-03-05
Mark posix_fallocate as supported on AIX

ecnelises:
AIX actually supports posix_fallocate procedure, although the fully functioning fallocate is not available.


sunfishcode:
Thanks!

PR #1040, opened 2024-03-28
Replace `Ioctl::OPCODE` with an `Ioctl::opcode()` method

ids1024:
This is a breaking API change.

This is needed for something like `SPI_IOC_MESSAGE(N)`, which takes an array of `N` `spi_ioc_transfer` structs. Which nix this ioctl would use `ioctl_write_buf!`. I'm not sure what other ioctls use this.

It probably should also have a helper type implementing `Ioctl` for this use case. Though it would help to know that it works for more than one ioctl...

### DIFF
--- a/src/ioctl/mod.rs
+++ b/src/ioctl/mod.rs
@@ -85,7 +85,7 @@ use bsd as platform;
 #[inline]
 pub unsafe fn ioctl<F: AsFd, I: Ioctl>(fd: F, mut ioctl: I) -> Result<I::Output> {
     let fd = fd.as_fd();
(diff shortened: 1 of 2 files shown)


ids1024:
Right. It's not urgent, but seems good to make sure some change like this gets in whenever there is another API break.

I think with this, all ioctls should be possible to support. Though it's hard to be sure.

sunfishcode:
I apologize for losing track of this, but this change was recently implemented in https://github.com/bytecodealliance/rustix/pull/1286.

PR #1041, opened 2024-03-30
Add ELF aux vector AT_MINSIGSTKSZ

ur4t:
https://github.com/rust-lang/rust/commit/9da004ea1997ac0eedbb882fa52960417725bd48 uses `AT_MINSIGSTKSZ` which is not supported by `rustix`. `AT_MINSIGSTKSZ` was introduced since glibc 2.35 on all architectures.


sunfishcode:
Thanks!

PR #1045, opened 2024-04-15
Pass null instead of empty slices to getxattr and fgetxattr on macos

Stebalien:
fixes the remainder of #957
fixes https://github.com/Stebalien/xattr/issues/57

### DIFF
--- a/src/backend/libc/fs/syscalls.rs
+++ b/src/backend/libc/fs/syscalls.rs
@@ -2254,15 +2254,24 @@ pub(crate) fn getxattr(path: &CStr, name: &CStr, value: &mut [u8]) -> io::Result
     }
 
     #[cfg(apple)]
-    unsafe {
-        ret_usize(c::getxattr(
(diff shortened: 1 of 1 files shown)


src/backend/libc/fs/syscalls.rs
```diff
@@ -2318,15 +2327,24 @@ pub(crate) fn fgetxattr(fd: BorrowedFd<'_>, name: &CStr, value: &mut [u8]) -> io
...
-        ret_usize(c::fgetxattr(
-            borrowed_fd(fd),
-            name.as_ptr(),
-            value_ptr.cast::<c::c_void>(),
-            value.len(),
-            0,
-            0,
-        ))
+    {
+        // Passing an empty to slice to getxattr leads to ERANGE on macOS. Pass null
```
Stebalien, line 2331: This one wasn't reported, but I assume it has the same behavior.

sunfishcode:
Thanks!

PR #1056, opened 2024-05-05
Add io_uring setup flags `NO_MMAP` and `REGISTERED_FD_ONLY`

ryanseipp:
Adds the following setup flags:
 * `IORING_SETUP_NO_MMAP` - liburing 2.5, kernel 6.5
 * `IORING_SETUP_REGISTERED_FD_ONLY` - liburing 2.5, kernel 6.5

Upgrades linux-raw-sys to 0.6.4 to provide the constants with the linux-raw backend. This upgrade came with three renamed fields which were previously reserved by padding. The flags field in `io_uring_buf_reg` is currently unused by liburing, and doesn't have a good definition that can be ported to bitflags.

https://github.com/axboe/liburing/blob/master/src/include/liburing/io_uring.h#L188


ryanseipp:
It looks like linux-raw-sys 6.4 has a misconfigured cfg macro for `c_char` on the sparc platform. I can take a look at submitting a PR to fix that tomorrow, to fix the degraded results from the tier 3 check.

sunfishcode:
We can't update to linux-raw-sys 0.6 yet because it's semver-exposed. But I have now added https://github.com/sunfishcode/linux-raw-sys/pull/122 to add the needed constants to the linux-raw-sys 0.4 branch.

sunfishcode:
That patch is now in linux-raw-sys 0.4.14.

sunfishcode:
Would you be able to downgrade this PR to linux-raw-sys 0.4.14, so that we don't need a semver bump here?

ryanseipp:
Sorry, haven't been able to get back to this yet. I'll give the downgrade a go this evening.

sunfishcode:
Thanks!

sunfishcode:

PR #1061, opened 2024-05-20
Update linux-raw-sys to 0.6

oherrala:
This required changing fields of some structs under `io_uring` thus probably braking semantic versioning.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ once_cell = { version = "1.5.2", optional = true }
 # libc backend can be selected via adding `--cfg=rustix_use_libc` to
 # `RUSTFLAGS` or enabling the `use-libc` cargo feature.
 [target.'cfg(all(not(rustix_use_libc), not(miri), target_os = "linux", any(target_endian = "little", target_arch = "s390x"), any(target_arch = "arm", all(target_arch = "aarch64", target_pointer_width = "64"), target_arch = "riscv64", all(rustix_use_experimental_asm, target_arch = "powerpc64"), all(rustix_use_experimental_asm, target_arch = "s390x"), all(rustix_use_experimental_asm, target_arch = "mips"), all(rustix_use_experimental_asm, target_arch = "mips32r6"), all(rustix_use_experimental_asm, target_arch = "mips64"), all(rustix_use_experimental_asm, target_arch = "mips64r6"), target_arch = "x86", all(target_arch = "x86_64", target_pointer_width = "64"))))'.dependencies]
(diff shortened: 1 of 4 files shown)


sunfishcode:
Yes, this is unfortunately semver breaking. If there's a specific feature you need in linux-raw-sys 0.6, we can backport or add compatible definitions it to 0.4.

raftario:
I'm personally interested in this for the futex support in io_uring

sunfishcode:
0.38.x releases are now being done in a dedicated 0.38 branch, so `main` can now accept semver-breaking changes.

The CI failure looks like rustix's `io_uring_sync_cancel_reg` needs to be updated to match the changes in Linux.

raftario:
@oherrala Do you plan on updating this to fix CI ? If not I'd be happy to take over

sunfishcode:
The CI failure here looks due to changes in the layout of `xdp_umem_reg`, so it looks like rustix's `XdpUmemReg` needs to be updated to match.

sunfishcode:
The compile errors are fixed on main in #1173. However the linux-i686 failure [here](https://github.com/bytecodealliance/rustix/actions/runs/10971306900/job/30466286482?pr=1061) looks relevant:
```
thread 'net::types::test_sizes' panicked at src/net/types.rs:1817:5:
assertion `left == right` failed
  left: 32
 right: 28
```

arctic-alpaca:
> However the linux-i686 failure [here](https://github.com/bytecodealliance/rustix/actions/runs/10971306900/job/30466286482?pr=1061) looks relevant:

This is caused by the `libc` crate still using the previous/removed version of the struct: 
* https://github.com/rust-lang/libc/blob/main/src/unix/linux_like/linux/gnu/mod.rs#L396-L402 
* https://github.com/rust-lang/libc/blob/main/src/unix/linux_like/linux/musl/mod.rs#L308-L314

arctic-alpaca:
Starting with [`libc` version 0.2.165](https://github.com/rust-lang/libc/releases/tag/0.2.165), the struct is updated and this PR shouldn't be blocked on this anymore.

oherrala:
Rebased my changes on top of ef39477129e2cffbef8928324cf7264a79018f5d. Let's see if the tests pass now.

cgwalters:
We hit on the listxattr call exposing `c_char` and hence I think this PR would roll in https://github.com/sunfishcode/linux-raw-sys/commit/769c98892ae7053cf1ddb4c9f8bfc6deeeeff98d#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R20 which would fix that

sunfishcode:
Thanks!

PR #1064, opened 2024-05-26
Fix hurd build

sthibaul:


sthibaul:
(the macos cancellation is unrelated)

sunfishcode:
Thanks!

sunfishcode:

Issue #1076, opened 2024-06-17
Feature request: Add `fcntl_getlk` support


al8n:
Hi, may I require supporting `fcntl_getlk`?

PR #1077, opened 2024-06-17
Implement `fcntl_getlk`

al8n:
Solve #1076

### DIFF
--- a/src/fs/ioctl.rs
+++ b/src/fs/ioctl.rs
@@ -79,7 +79,10 @@ unsafe impl ioctl::Ioctl for Ficlone<'_> {
     type Output = ();
 
     const IS_MUTATING: bool = false;
-    const OPCODE: ioctl::Opcode = ioctl::Opcode::old(c::FICLONE as ioctl::RawOpcode);
+
+    fn opcode(&self) -> ioctl::Opcode {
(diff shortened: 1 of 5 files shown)


src/backend/linux_raw/fs/syscalls.rs
```diff
@@ -1090,6 +1090,60 @@ pub(crate) fn fcntl_lock(fd: BorrowedFd<'_>, operation: FlockOperation) -> io::R
...
+
+    let mut lock = flock {
+        l_type: F_WRLCK as _,
+
+        // When `l_len` is zero, this locks all the bytes from
+        // `l_whence`/`l_start` to the end of the file, even as the
+        // file grows dynamically.
+        l_whence: SEEK_SET as _,
+        l_start: 0,
+        l_len: 0,
```
notgull, line 1109: It might be nice to give the user control over these fields via parameters to the function call.

notgull (changes requested):

src/backend/linux_raw/fs/syscalls.rs
sunfishcode: I agree that it'd be nice to expose this functionality, however the existing `fcntl_lock`  doesn't currently give the user control over these flags either, so this is ok for now.

sunfishcode:
Overall this looks good, though there are some test failures that look relevant.

PR #1079, opened 2024-06-20
Add support for Netlink sockets

ids1024:
I noticed that `msg_namelen` is being set to the size of `SocketAddr*`, which seems to be incorrect. So I also added a commit changing that for the existing socket address types.

I wonder if some of the code here could be less redundant, but that isn't changed here.


ids1024:
Ah, I guess this (and making the socket code less redundant) is covered in https://github.com/bytecodealliance/rustix/pull/1004. I didn't really expect anyone else would have tried adding Netlink support recently.

Though I guess the `size_of` change is still needed.

sunfishcode:
Closing, as I believe #1004 now implements this, and removes the code where the `size_of` fix was needed.

PR #1080, opened 2024-06-27
Fix soundness issues in `dup2_stdout` et al

pekkarr:
The `dup2_std*` functions had a few soundness issues that allowed safe code to close the stdio file descriptors.

1. `AsFd::as_fd()` was called twice, possibly returning a different `BorrowedFd` each time. This would allow safe code to bypass the `fd.as_fd().as_raw_fd() != c::STDOUT_FILENO` check.
2. `AsFd::as_fd()` was called between `let mut target = unsafe { take_stdout() };` and `forget(target);`. If `as_fd()` panics, then `target` would be automatically dropped without the call to `forget`. This violates [exception safety](https://doc.rust-lang.org/nomicon/exception-safety.html).
3. If `backend::io::syscalls::dup2` returns an error, this was immediately returned from the function using the `?` operator, so `forget(target);` wasn't called resulting in `target` getting dropped closing the stdio file descriptor.

Fix these issues by only calling `AsFd::as_fd()` once and by using [`ManuallyDrop`](https://doc.rust-lang.org/stable/core/mem/struct.ManuallyDrop.html) instead of `forget`.

### DIFF
--- a/src/stdio.rs
+++ b/src/stdio.rs
@@ -17,7 +17,7 @@ use backend::fd::{BorrowedFd, FromRawFd, RawFd};
 use {
     crate::io,
     backend::fd::{AsFd, AsRawFd},
-    core::mem::forget,
+    core::mem::ManuallyDrop,
 };
 
 /// `STDIN_FILENO`—Standard input, borrowed.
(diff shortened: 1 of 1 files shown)


sunfishcode:
Great catches, thanks!

sunfishcode:

PR #1082, opened 2024-06-27
Fix `dup2_stdio` tests to test all three functions

pekkarr:
These tests were added in #1069, but despite the names suggesting that they test all of stdin, stdout and stderr, the tests were identical and only tested stdout.


sunfishcode:
Good catch! The Windows failures here are unrelated and should be fixed on main in #1081.

PR #1086, opened 2024-07-06
Add loongarch64 support

zhaixiaojuan:


sunfishcode:
Rustix doesn't use `__NR_setrlimit` or `__NR_getrlimit` to implement `getrlimit` and `setrlimit`. Linux on loongarch64 does appear to have `__NR_prlimit64`, so could the `getrlimit` and `setrlimit` functions be supported?

sunfishcode:
> @sunfishcode Hi, I found that getrlimit and setrlimit functions have been implemented using prlimit64 in the file src/backend/linux_raw/process/syscalls.rs. Maybe loongarch64 does not need additional processing.
> 
> But I am not sure which file the following code calls in the file src/process/rlimit.rs:
> 
> ```
> pub fn setrlimit(resource: Resource, new: Rlimit) -> io::Result<()> {
> backend::process::syscalls::setrlimit(resource, new)
> ```
> 
> Because both src/backend/linux_raw/process/syscalls.rs and src/backend/libc/process/syscalls.rs define the two functions setrlimit and getrlimit.

Rustix has two "backends", one that uses libc calls and one that uses raw linux syscalls. The code in src/process/rlimit.rs will call either the code in src/backend/linux_raw/process/syscalls.rs or src/backend/libc/process/syscalls.rs depending on which backend the user has configured. However, I'd expect both backends to work with loongarch64, because rustix doesn't use the `__NR_setrlimit` or `__NR_getrlimit` syscalls in either case.

src/fs/fd.rs
```diff
@@ -161,6 +162,7 @@ pub fn fchown<Fd: AsFd>(fd: Fd, owner: Option<Uid>, group: Option<Gid>) -> io::R
 /// [Linux]: https://man7.org/linux/man-pages/man2/fstat.2.html
 /// [`Mode::from_raw_mode`]: Mode::from_raw_mode
 /// [`FileType::from_raw_mode`]: crate::fs::FileType::from_raw_mode
+#[cfg(not(target_arch = "loongarch64"))]
 #[inline]
 pub fn fstat<Fd: AsFd>(fd: Fd) -> io::Result<Stat> {
```
sunfishcode, line 167: I assume loongarch64 should support the public `fstat` API. There may be no `__NR_fstat` but if so then src/backend/linux_raw/fs/syscalls.rs' `fstat` can probably use `statx` instead.

src/process/rlimit.rs
```diff
@@ -35,6 +36,7 @@ pub fn getrlimit(resource: Resource) -> Rlimit {
 ///
 /// [POSIX]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/setrlimit.html
 /// [Linux]: https://man7.org/linux/man-pages/man2/setrlimit.2.html
+#[cfg(not(target_arch = "loongarch64"))]
 #[inline]
 pub fn setrlimit(resource: Resource, new: Rlimit) -> io::Result<()> {
```
sunfishcode, line 41: Similarly, I assume loongarch64 should support the public `setrlimit` API even if there is no `__NR_setrlimit`, as it can use `prlimit64` instead.

sunfishcode (commented):
Since loongarch64 has Tier-2 targets in Rust, adding support for raw syscalls to rustix should also add testing using qemu in the CI scripts, following the other architectures.

xtqqczze:
@zhaixiaojuan needs to be rebased

PR #1087, opened 2024-07-11
rawdir: Use `ignore`, not `notrust`

cgwalters:
- Fixes the problem that the prelude comments show up in the docs, including the confusing comment " # // The `notrust` above can be removed when we can depend on Rust 1.65." because the reader can't *see* the `notrust`
- I parsed `notrust` as "no-trust" and thought there was some security thing

The `ignore` attribute AFAICS is intended for these situations.


sunfishcode:
Sounds good, thanks!

PR #1088, opened 2024-07-17
setns: Add a doc alias

cgwalters:
At first I thought this wasn't bound in rustix until I went to search the source code to look at adding it.


sunfishcode:
Thanks!

PR #1089, opened 2024-07-19
Fix mising MREMAP_FIXED for mremap_fixed for linux_raw

zanglg:
According to https://man7.org/linux/man-pages/man2/mremap.2.html, mremap_fixed missed flag of MREMAP_FIXED. This flag should be applied automatically like libc-backend.


sunfishcode:
Thanks!

sunfishcode:

PR #1090, opened 2024-07-20
Enable linux_raw support for android

yujincheng08:
I have no idea why we need use libc backend, but definitely, we should not prevent users from using linux_raw backend.

CI test fails because some constants are missing from libc, and the fix is rust-lang/libc#3779.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ once_cell = { version = "1.5.2", optional = true }
 # addition to the libc backend. The linux_raw backend is used by default. The
 # libc backend can be selected via adding `--cfg=rustix_use_libc` to
 # `RUSTFLAGS` or enabling the `use-libc` cargo feature.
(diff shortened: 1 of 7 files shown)


sunfishcode:
I can't find a link at the moment, but I recall discussions where people believed Android doesn't guarantee raw syscall behavior, and expects users use its libc. Do you know if that's the case?

My other concern here is that we don't currently have any CI testing on Android, so I'm hesitant to enable linux_raw by default without a way of testing it.

yujincheng08:
I don't think that's the way. Android also uses a Linux kernel, and many apps (like Google Chrome) are using raw syscall with libc.

yujincheng08:
For CI, we can use Android emulator for testing (see https://github.com/rust-lang/libc/blob/main/.github/workflows/full_ci.yml)

notgull:
Is there any significant difference between Linux libc and Android libc?

yujincheng08:
The libc of Andorid is bonic (a implementation rather than gnu libc, musl, or llvm-libc).

See https://cs.android.com/android/platform/superproject/main/+/main:bionic/

sunfishcode:
I will not be able to add Android emulator support to rustix's CI myself, so if this is to happen, someone else will need to to this.

I also don't have confidence in my ability to make an appropriate judgement here, so I've opened https://github.com/bytecodealliance/rustix/issues/1095 in hopes of receiving more input.

yujincheng08:
Thanks.

yujincheng08:
Then we will stick to [our fork](https://github.com/Kernel-SU/rustix) instead.

https://github.com/tiann/KernelSU/blob/59ba84eb2b45c7ee7cded8774c918100d62a2712/userspace/ksud/Cargo.toml#L51-L53

PR #1092, opened 2024-07-21
Do some todos

primoly:
- add `MOVE_MOUNT_BENEATH`
- use libc’s `SO_NOSIGPIPE`
- use libc’s `AT_MINSIGSTKSZ`

### DIFF
--- a/src/backend/libc/mount/types.rs
+++ b/src/backend/libc/mount/types.rs
@@ -227,7 +227,7 @@ bitflags! {
         /// `MOVE_MOUNT__MASK`
         const MOVE_MOUNT_SET_GROUP = 0x0000_0100;
 
-        /// `MOVE_MOUNT_BENEATH`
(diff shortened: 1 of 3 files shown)


sunfishcode:
For libc, we can just bump the version in Cargo.toml.

For linux-raw-sys, it looks like this needs a newer Linux version, however bindings generated from newer Linux versions are semver-incompatible with the version rustix is currently using, and this is unfortunately exposed in rustix's public API. A workaround would be to add the constants manually to [general.h](https://github.com/sunfishcode/linux-raw-sys/blob/main/gen/modules/general.h) though, which would allow them to be released in linux-raw-sys 0.4.x.

sunfishcode:
I took the liberty of rebasing this, as some of these todos are now fixed on main, and I added a temporary workaround for `MOVE_MOUNT_BENEATH`, so we can include this in 0.38.35.

PR #1093, opened 2024-07-21
Disable UWP windows target in CI for now

notgull:
It appears this target is broken in upstream Rust.

cc rust-lang/rust#100400

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -219,7 +219,8 @@ jobs:
     # Omit --all-targets on haiku because not all the tests build yet.
     # Temporarily disable this because the target appears to have breakage on nightly.
     #- run: cargo check -Z build-std --target x86_64-unknown-haiku --features=all-apis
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -219,7 +219,7 @@ jobs:
     # Omit --all-targets on haiku because not all the tests build yet.
     # Temporarily disable this because the target appears to have breakage on nightly.
     #- run: cargo check -Z build-std --target x86_64-unknown-haiku --features=all-apis
-    - run: cargo check -Z build-std --target x86_64-uwp-windows-msvc --all-targets --features=all-apis
+    #- run: cargo check -Z build-std --target x86_64-uwp-windows-msvc --all-targets --features=all-apis
```
sunfishcode, line 222: Could you add a comment like the one on line 220??

notgull: Fixed!

PR #1097, opened 2024-07-26
Add missing futex operations

danielschemmel:
This PR simply exposes already existing futex operations to the user. I recently ran into this issue when attempting to port a futex module to use rustix as its backend, where I ended up missing `FUTEX_CMP_REQUEUE_PI`. Note that other PI operations (such as `FUTEX_LOCK_PI`) are already exposed.

The additions are symmetric across the backends.


danielschemmel:
Included in #1098

Issue #753, opened 2023-07-19
1.0 release planning


sunfishcode:
Here's a planning issue for a 1.0 release, similar to the one for the [0.38 release](https://github.com/bytecodealliance/rustix/issues/578). Hopefully we won't need a semver bump for a good long while, but when we do, here's the list of changes we can make:

 - [x] Remove the deprecated mount functions in the `fs` module [(details)](https://github.com/bytecodealliance/rustix/pull/752#issuecomment-1642328674)
 - [x] Make the underlying type of `rustix::fs::Access` consistent between linux_raw and libc backends (https://github.com/bytecodealliance/rustix/pull/1152)
 - [x] Rename `mount_recursive_bind` to `mount_bind_recursive`? https://github.com/bytecodealliance/rustix/pull/763#issuecomment-1662756184 (https://github.com/bytecodealliance/rustix/pull/1273)
 - [x] Change `SeekFrom::Hole` and `SeekFrom::Data`'s field from `i64` to `u64`. (https://github.com/bytecodealliance/rustix/pull/1266)
 - [x] rename `Shutdown::ReadWrite` to `Shutdown::Both` to align with std (https://github.com/bytecodealliance/rustix/pull/1264)
 - [x] remove the `features = ["extra_traits"]` from the libc dependency (#1168)
 - [x] move (or perhaps alias?) `sched_yield` and other `sched_*` functions from `process` to `thread`, since they apply to threads rather than processes on Linux etc. (https://github.com/bytecodealliance/rustix/pull/1265)
 - [x] `stx_mtime` and similar members of `Statx` should be `i64`, because Unix timestamps can be negative.
 - [x] `futex`'s first argument should be `Atomic`
 - [x] Remove `get_` from sockopt getters, following [Rust convention](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter). Consider removing `get_` in other things too. (https://github.com/bytecodealliance/rustix/pull/1269)
 - [x] The `udata` argument to `rustix::event::kqueue::Event::new` should be a pointer instead of an `isize` in order to better preserve provenance. (https://github.com/bytecodealliance/rustix/pull/1270)
 - [x] The `stx_attributes_mask` field of `Statx` with flags like `libc::STATX_ATTR_MOUNT_ROOT` should ideally be a bitflags, which requires making `Statx` a custom type (which should match layout with `libc::statx`). (https://github.com/bytecodealliance/rustix/pull/1282)
[...]

Issue #214, opened 2022-02-04
Split `futex` into separate functions


sunfishcode:
Rustix's [futex function](https://docs.rs/rustix/latest/rustix/thread/fn.futex.html) can perform several different operations, as determined by a `FutexOperation` argument. However, as documented in the Arguments section of [Linux's man page](https://man7.org/linux/man-pages/man2/futex.2.html), in some operations, some of the arguments are unused. We should consider splitting `futex` into multiple functions, one per signature at least, so that we can provide customized signatures for each operation.

A related crate in this space is the [linux-futex crate]. Another option here would be to simply port linux-futex to use rustix's existing low-level futex API internally.

[linux-futex crate]: https://docs.rs/linux-futex/latest/linux_futex/

sunfishcode:
This is [implemented](https://docs.rs/rustix/latest/rustix/thread/futex/index.html#functions).

PR #1098, opened 2024-07-26
Futex Rework

danielschemmel:
This PR splits `futex` into a zoo of `futex_*` functions with different and correctly typed arguments.

There are still a few open issues and questions:
1. The `futex` syscall treats one of its arguments as _either_ a _nullable_ `*const Timespec` _or_ a `u32`, depending on the context. The `futex_time64` fallback to `futex_old` uses `NOSYS` as its deciding factor, which can also be generated in other ways. This may lead to `futex_old` dereferencing an invalid pointer (because it is null, or not a pointer in the first place). As we now have different entrypoints that know whether the value should be a nullable pointer or a `u32`, I would suggest splitting up the backend `futex` functions into two variants (each) with different signatures. This would prevent a fallback to `futex_old` happening when the argument is not a pointer. Adding the null check is then trivial.
2. There are two flags that may sometimes change the function behavior. The first flag, `FUTEX_PRIVATE_FLAG` basically just says whether the lock(s) are guaranteed to be process-private (for some kernel optimizations). It is available for every single function. The second flag `FUTEX_CLOCK_REALTIME` changes how the timeout is interpreted in a few functions. Removing those flags by way of introducing more functions is viable, but I am unsure if it really makes sense to add that many more functions. Since their meaning is very different, I would suggest splitting them into two enums (`FutexPrivacy::Shared` and `FutexPrivacy::ProcessPrivate`, and `FutexClock::Realtime` and `FutexClock::Monotone`) and making them two arguments. It might even make sense to take them as compile time parameters. What do you think?
3. There are (already) a lot of functions. Instead of having them be global functions, there could be a `struct Futex<'a>(&'a AtomicU32);` (and similarly `FutexPI`), which then could show only the functions applicable to that scenario. E.g., `Futex(&word).cmp_requeue(...)` and `FutexPI(&word).cmp_requeue(...)` would call `futex_cmp_requeue` and `futex_cmp_requeue_pi` respectively without polluting each other's namespace. While this would make the API a bit nicer, I fear it would suggest that it does more than just the syscalls. What do you prefer?
4. While I have not done so right now, it should be possible to retain the old `futex` API, so as to be semver compatible - ideally with a `#[deprecated]`. Would this be important to you? If so, should `FutexOperations` be retained in a way that does not add any new variants?

This PR implements #214 and one point from #753 (`futex`'s first argument should be `Atomic`) and builds on the changes from #1097.

### DIFF
--- a/src/backend/libc/thread/syscalls.rs
+++ b/src/backend/libc/thread/syscalls.rs
@@ -597,7 +597,10 @@ unsafe fn futex_old_timespec(
     } else {
         Some(linux_raw_sys::general::__kernel_old_timespec {
             tv_sec: (*timeout).tv_sec.try_into().map_err(|_| io::Errno::INVAL)?,
-            tv_nsec: (*timeout).tv_nsec.try_into().map_err(|_| io::Errno::INVAL)?,
+            tv_nsec: (*timeout)
+                .tv_nsec
+                .try_into()
+                .map_err(|_| io::Errno::INVAL)?,
         })
     };
     ret_usize(futex(
--- a/src/backend/libc/thread/syscalls.rs
+++ b/src/backend/libc/thread/syscalls.rs
@@ -8,12 +8,14 @@ use crate::thread::{NanosleepRelativeResult, Timespec};
 #[cfg(all(target_env = "gnu", fix_y2038))]
 use crate::timespec::LibcTimespec;
 use core::mem::MaybeUninit;
+use core::sync::atomic::AtomicU32;
 #[cfg(linux_kernel)]
 use {
+    super::futex::FutexOperation,
     crate::backend::conv::{borrowed_fd, ret_c_int, ret_usize},
     crate::fd::BorrowedFd,
     crate::pid::Pid,
-    crate::thread::{FutexFlags, FutexOperation},
+    crate::thread::FutexFlags,
     crate::utils::as_mut_ptr,
 };
 #[cfg(not(any(
@@ -418,12 +420,12 @@ pub(crate) fn setresgid_thread(
 // TODO: This could be de-multiplexed.
 #[cfg(linux_kernel)]
 pub(crate) unsafe fn futex(
-    uaddr: *mut u32,
+    uaddr: *const AtomicU32,
     op: FutexOperation,
     flags: FutexFlags,
     val: u32,
     utime: *const Timespec,
-    uaddr2: *mut u32,
+    uaddr2: *const AtomicU32,
     val3: u32,
 ) -> io::Result<usize> {
     #[cfg(all(
@@ -437,11 +439,11 @@ pub(crate) unsafe fn futex(
 
         syscall! {
             fn futex_time64(
-                uaddr: *mut u32,
+                uaddr: *const AtomicU32,
                 futex_op: c::c_int,
                 val: u32,
                 timeout: *const Timespec,
-                uaddr2: *mut u32,
+                uaddr2: *const AtomicU32,
                 val3: u32
             ) via SYS_futex_time64 -> c::ssize_t
         }
@@ -473,11 +475,11 @@ pub(crate) unsafe fn futex(
     {
         syscall! {
             fn futex(
-                uaddr: *mut u32,
+                uaddr: *const AtomicU32,
                 futex_op: c::c_int,
                 val: u32,
                 timeout: *const linux_raw_sys::general::__kernel_timespec,
-                uaddr2: *mut u32,
+                uaddr2: *const AtomicU32,
                 val3: u32
             ) via SYS_futex -> c::c_long
         }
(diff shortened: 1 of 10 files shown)


notgull:
It might be better to try this against the `1.0-staging` branch, which already has some breaking changes.

danielschemmel:
My bad, would it make more sense to ensure there are no breaking changes, or to target `1.0-staging`?

src/backend/libc/thread/syscalls.rs
```diff
@@ -418,12 +420,12 @@ pub(crate) fn setresgid_thread(
 // TODO: This could be de-multiplexed.
```
sunfishcode, line 420: This PR implements this TODO comment, so it can be removed.

src/backend/linux_raw/thread/syscalls.rs
```diff
@@ -206,12 +208,12 @@ pub(crate) fn gettid() -> Pid {
 // TODO: This could be de-multiplexed.
```
sunfishcode, line 208: This TODO comment as well!

sunfishcode (commented):
Thanks for working on this! I'll need more time to fully review this, but at a first glance this looks like a good direction.

> * The futex syscall treats one of its arguments as either a nullable *const Timespec or a u32, depending on the context. The futex_time64 fallback to futex_old uses NOSYS as its deciding factor, which can also be generated in other ways. This may lead to futex_old dereferencing an invalid pointer (because it is null, or not a pointer in the first place). As we now have different entrypoints that know whether the value should be a nullable pointer or a u32, I would suggest splitting up the backend futex functions into two variants (each) with different signatures. This would prevent a fallback to futex_old happening when the argument is not a pointer. Adding the null check is then trivial.

This sounds like a good idea to me.

> * There are two flags that may sometimes change the function behavior. The first flag, `FUTEX_PRIVATE_FLAG` basically just says whether the lock(s) are guaranteed to be process-private (for some kernel optimizations). It is available for every single function. The second flag `FUTEX_CLOCK_REALTIME` changes how the timeout is interpreted in a few functions. Removing those flags by way of introducing more functions is viable, but I am unsure if it really makes sense to add that many more functions. Since their meaning is very different, I would suggest splitting them into two enums (`FutexPrivacy::Shared` and `FutexPrivacy::ProcessPrivate`, and `FutexClock::Realtime` and `FutexClock::Monotone`) and making them two arguments. It might even make sense to take them as compile time parameters. What do you think?

Could you say more about what the advantage of this would be over the current `FutexFlags` approach?

> * There are (already) a lot of functions. Instead of having them be global functions, there could be a `struct Futex<'a>(&'a AtomicU32);` (and similarly `FutexPI`), which then could show only the functions applicable to that scenario. E.g., `Futex(&word).cmp_requeue(...)` and `FutexPI(&word).cmp_requeue(...)` would call `futex_cmp_requeue` and `futex_cmp_requeue_pi` respectively without polluting each other's namespace. While this would make the API a bit nicer, I fear it would suggest that it does more than just the syscalls. What do you prefer?

I like how global functions makes it clear that they're "just functions"; as you say, it also feels like a wrapper type might suggest it does more. For namespacing purposes, what if we created a new `pub mod futex` and put all these new functions in it? Perhaps we could even drop the `futex_` prefix from them, so users could do `futex::wait` instead of `futex_wait`.

> * While I have not done so right now, it should be possible to retain the old `futex` API, so as to be semver compatible - ideally with a `#[deprecated]`. Would this be important to you? If so, should `FutexOperations` be retained in a way that does not add any new variants?

Yes, I think it would be good to retain the old API and remain semver-compatible. A `#[deprecated]` makes sense to me too. I don't have an opinion about whether to hide new variants in `FutexOperations`.

src/backend/libc/thread/syscalls.rs
danielschemmel: Done

src/backend/linux_raw/thread/syscalls.rs
danielschemmel: Done

danielschemmel:
All of @sunfishcode's suggestions should be implemented - thanks!
- `futex` is now a deprecated function as well as a module that contains functions like `wake` and `cmp_requeue_pi`.
- the backends are split into `futex_val2` and `futex_timespec`, which should both handle the argument without potentially segfaulting.
  - the `futex` functions now performs dynamic dispatch to pick the right backend

As to splitting the `PRIVATE` and `CLOCK` flags:
The `PRIVATE` flag applies to each function, as it is a property of the futex word, while the `CLOCK` flag applies to the timeout parameter that only 5 out of 14 functions even take.
The drawback is increasing the number of arguments to 5 out of 14 functions by one.

danielschemmel:
I have added some tests (and fixed a stupid bug), and cleaned up the naming a bit.

One thing I also noticed is that the new functions probably don't really need any `unsafe` anymore: as they now take and return safe types, they can only really cause a deadlock - which a normal rust function with a `loop { }` can just as well.

sunfishcode:
> One thing I also noticed is that the new functions probably don't really need any `unsafe` anymore: as they now take and return safe types, they can only really cause a deadlock - which a normal rust function with a `loop { }` can just as well.

This sounds right. Deadlocks are not a safety issue in Rust. At the moment it appears functions like `wait` are still unsafe in this PR. Are you planning to remove the unsafe?

danielschemmel:
> > One thing I also noticed is that the new functions probably don't really need any `unsafe` anymore: as they now take and return safe types, they can only really cause a deadlock - which a normal rust function with a `loop { }` can just as well.
> 
> This sounds right. Deadlocks are not a safety issue in Rust. At the moment it appears functions like `wait` are still unsafe in this PR. Are you planning to remove the unsafe?

Yes, I'll do so right away, I just wanted to have your input before removing the unsafe.

danielschemmel:
Done.

One more note, libc [includes FUTEX_WAITERS/FUTEX_OWNER_DIED](https://docs.rs/libc/latest/libc/constant.FUTEX_OWNER_DIED.html) as of 0.2.158, so if you are OK with updating the dependency, we can also use that in the libc backend.

sunfishcode:
Yes, updating the libc dependency to the latest 0.2 is fine.

danielschemmel:
Ah well, turns out libc's "linux-like" and rustix's "linux_kernel" are just different enough that this fails on android.

sunfishcode:
Looks good, thanks!

sunfishcode:

PR #1099, opened 2024-08-01
Add missing epoll links

SUPERCILEX:


sunfishcode:
That comment about `delete` was an attempt to summarize the behavior documented [here](https://man7.org/linux/man-pages/man7/epoll.7.html) under the question "Will closing a file descriptor cause it to be removed from all epoll interest lists?". Specifically the idea of a `Weak<dyn SystemResource>` describes this behavior of it holding onto the reference as long as there are 1 or more (strong) file descriptors open for it. That said, reading it now, that first sentence does appear to be incorrect. Would you be interested in rewording that first sentence to capture the documented behavior, or would you prefer I write up something?

sunfishcode:
Looks good. I hadn't seen that "#:~:text=" syntax before and it doesn't currently work in Firefox, but I looked it up and confirmed that Mozilla's stance on it is [positive](https://github.com/mozilla/standards-positions/blob/0aa0de92a1edddddf6fb312f7c5ce454651536da/activities.json#L1636), so I'll go with it :smile: .

SUPERCILEX:
Lol, it's ugly as sin for sure, but worst case scenario it doesn't do anything in a browser that doesn't support it and hopefully gets people to the right place in one that does.

Anyway, thanks!

PR #1105, opened 2024-08-05
Unify inotify public interface across backends and accept `impl AsFd`

SUPERCILEX:
It currently only accepts a `BorrowedFd` which is annoying. Pretty sure this is a backwards compatible change.

### DIFF
--- a/src/backend/libc/fs/inotify.rs
+++ b/src/backend/libc/fs/inotify.rs
@@ -1,9 +1,6 @@
 //! inotify support for working with inotifies
 
 use crate::backend::c;
-use crate::backend::conv::{borrowed_fd, c_str, ret, ret_c_int, ret_owned_fd};
-use crate::fd::{BorrowedFd, OwnedFd};
-use crate::io;
 use bitflags::bitflags;
 
 bitflags! {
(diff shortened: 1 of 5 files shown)


src/fs/inotify.rs
```diff
@@ -24,12 +24,12 @@ pub fn inotify_init(flags: CreateFlags) -> io::Result<OwnedFd> {
 /// this method may result in it returning the same watch descriptor. An
 /// application should keep track of this externally to avoid logic errors.
 #[inline]
-pub fn inotify_add_watch<P: crate::path::Arg>(
-    inot: BorrowedFd<'_>,
+pub fn inotify_add_watch<Fd: AsFd, P: crate::path::Arg>(
```
SUPERCILEX, line 27: Actually I think adding a generic isn't backwards compatible if someone explicitly specified `P`. Then again, I don't think anybody uses this code? https://github.com/search?q=inotify_add_watch+language%3ARust+rustix+-is%3Afork&type=code At least not publicly.

sunfishcode: Would it by chance work to use `<P: crate::path::Arg, Fd: AsFd>`?

SUPERCILEX: I think it would complain with a missing generic and you'd have to put an underscore. Maybe we could do `impl AsFd` and put a todo to make it a generic later?

sunfishcode: Making it `impl AsFd` and adding a TODO sounds reasonable to me.

SUPERCILEX: Sounds good, done!

src/fs/inotify.rs
```diff
@@ -0,0 +1,43 @@
...
+///
+/// This registers or updates a watch for the filesystem path `path` and
+/// returns a watch descriptor corresponding to this watch.
+///
+/// Note: Due to the existence of hardlinks, providing two different paths to
+/// this method may result in it returning the same watch descriptor. An
+/// application should keep track of this externally to avoid logic errors.
+#[inline]
+pub fn inotify_add_watch<P: crate::path::Arg>(
+    inot: impl AsFd,
```
sunfishcode, line 28: Could you add the TODO comments about changing this `impl AsFd` to a generic argument in the future?

src/backend/libc/fs/inotify.rs
```diff
@@ -79,53 +76,3 @@ bitflags! {
...
     }
 }
-
-/// `inotify_init1(flags)`—Creates a new inotify object.
-///
-/// Use the [`CreateFlags::CLOEXEC`] flag to prevent the resulting file
-/// descriptor from being implicitly passed across `exec` boundaries.
-#[doc(alias = "inotify_init1")]
-pub fn inotify_init(flags: CreateFlags) -> io::Result<OwnedFd> {
-    // SAFETY: `inotify_init1` has no safety preconditions.
```
sunfishcode, line 89: Could you restore these three `SAFETY` comments? I know they're not super interesting, but they're there for the benefit of people auditing the code. They record what we believed the relevant safety considerations are, so that if it ever turns out that we missed something, it's easier to understand the intent of the code.

sunfishcode (commented):
Looks good, just a few minor details:

src/backend/libc/fs/inotify.rs
SUPERCILEX: Makes sense, restored.

src/fs/inotify.rs
SUPERCILEX: See https://github.com/bytecodealliance/rustix/issues/753#issuecomment-2291869835. I feel like it'll be more effective to run a global pass on the whole lib rather than trying to remember the spots impls need to be fixed. If you disagree, I'll be happy to add the TODOs here.

sunfishcode: Ah, makes sense.

sunfishcode:

Issue #1107, opened 2024-08-14
1.0-staging was rebased incorrectly


SUPERCILEX:
@sunfishcode it looks like two 1.0 PRs were wiped off the map accidentally. https://github.com/bytecodealliance/rustix/pull/996 and https://github.com/bytecodealliance/rustix/pull/1063 don't exist anymore.

1.0 PRs: https://github.com/bytecodealliance/rustix/pulls?q=is%3Apr+base%3A1.0-staging+is%3Aclosed

sunfishcode:
Oops, thanks for catching that. I've now cherry-picked those patches back into the 1.0-staging tree.

SUPERCILEX:
Sweet!

PR #1108, opened 2024-08-15
Add uninit buffer ancillary APIs

### LINKED ISSUE #1103: Make RecvAncillaryBuffer and SendAncillaryBuffer use MaybeUninits
Any objections to this? Would be nice to avoid having to pull the buffer out of hot loops.

SUPERCILEX:
Pretty sure I managed to maintain backcompat via a conversion trait.

Closes #1103

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -267,5 +267,11 @@ argument optionallly containing a raw file descriptor.
 
 [`rustix::io_uring::io_uring_setup`]: https://docs.rs/rustix/1.0.0/rustix/io_uring/fn.io_uring_setup.html
 
+The buffer for [`SendAncillaryBuffer`] and [`RecvAncillaryBuffer`] is now
+a `[MaybeUninit<u8>]` instead of a `[u8]`.
+
+[`SendAncillaryBuffer`]: https://docs.rs/rustix/1.0.0/rustix/net/struct.SendAncillaryBuffer.html
+[`RecvAncillaryBuffer`]: https://docs.rs/rustix/1.0.0/rustix/net/struct.RecvAncillaryBuffer.html
+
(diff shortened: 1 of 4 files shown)


SUPERCILEX:
Nevermind, I don't think backcompat is possible without implementing a trait for both `DerefMut<Target=[u8]>` and `DerefMut<Target=[MaybeUninit<u8>]>` which I don't think you can do. Targeting the staging branch instead.

SUPERCILEX:
Staging needs a rebase and also https://github.com/bytecodealliance/rustix/issues/1107

SUPERCILEX:
Annnnnd I decided to make it backwards compatible again. The public API should be identical except there's now a `From` impl for `MaybeUninit`. 1.0 should break the `::new` impl to take uninits instead.

SUPERCILEX:
Not sure why the tests are so borked.

SUPERCILEX:
Ok so the idea with this one is that the `::new` methods would be swapped to use `MaybeUninit`s for 1.0 and any way to construct a buffer with `[u8]` would be removed. I'm not really happy with the deprecation story as it feels like we'll just be springing this change on people, but I also don't see a way to allow seamless migration while preserving the name `new`.

sunfishcode:
I've now rebased this and fixed the merge conflicts.

SUPERCILEX:
Can we remove the non MaybeUninit versions? I personally don't think you should be using u8s for these buffers since we take care of giving you a nice well typed iterator.

sunfishcode:
Yep, I was just reading your comment above about that too.

SUPERCILEX:
Brilliant, this looks great!

sunfishcode:
Thanks!

Issue #1100, opened 2024-08-01
Weird epoll comment around close


SUPERCILEX:
https://docs.rs/rustix/latest/rustix/event/epoll/fn.add.html says that closing a FD doesn't do anything which directly contradicts the man pages: https://man7.org/linux/man-pages/man7/epoll.7.html#:~:text=Will%20closing%20a%20file%20descriptor%20cause%20it%20to%20be%20removed%20from%20all%0A%20%20%20%20%20%20%20%20%20%20epoll%20interest%20lists%3F

@notgull is there context I'm missing or can we just remove that comment in favor of the man pages?

SUPERCILEX:
Hmmm, also confused about why `EventVec` is a thing. Shouldn't that just use the same API as https://docs.rs/rustix/latest/rustix/fs/fn.readlink.html? If you and @sunfishcode code are ok with that, I'd like to make the change. Also the `event_list.events.set_len(0);` is very suspicious—like yes everything is copy so whatever, but why not just call `clear`?

SUPERCILEX:
Actually the `Into` interface seems annoying since an `&mut Vec<Event>` would do. Though we can go a step further and take an `&mut [MaybeUninit<Event>]` and return an `&mut [Event]` so you can pass in stack arrays. The interface would look like this:

```rust
fn wait<I: Into<SliceOrVec>>(..., event_buf: I) -> io::Result<&mut [Event]>

enum SliceOrVec<'a> {
  Slice(&'a mut [MaybeUninit<Event>]),
  Vec(&'a mut Vec<Event>),
}

impl From slice/vec for SliceOrVec

// Usage
for event in wait(..., &mut stack OR vec.spare_capacity_mut()) {}

let mut vec = ...;
wait(..., vec.spare_capacity_mut());
do_stuff(&vec)
```

Thoughts? This seems way more flexible and less code than the current version.

SUPERCILEX:
Also I don't think the API should clear/set_len(0) the vec at all, that's the caller's problem (use drain) and they might want to only partially process events for some reason. I guess maybe it's fine? But seems weird to kill stuff in your buffer.

SUPERCILEX:
Ooo, pretty sure this is backwards compatible with a from impl for EventVec! So we could do all this in 0.38 with a deprecation and only drop EventVec later.

sunfishcode:
`EventVec` was removed in #1304, and #1290 implements support for no-alloc and uninitialized buffers.

[...]

PR #1110, opened 2024-08-15
Modernize epoll API

SUPERCILEX:
This PR:
- Unifies the epoll backends
- Implements the ideas in #1100, meaning the user can now pass in a Vec and uninits with the same API
- Enables using `epoll::wait` in non-alloc crates

cc @notgull for extra review

@sunfishcode this buffer trait idea seems pretty cool, maybe we should use this pattern in other areas that take uninits?

All of this should be fully backwards compatible.

### DIFF
--- a/src/event/epoll.rs
+++ b/src/event/epoll.rs
@@ -71,6 +71,7 @@
 
 #![allow(unsafe_code)]
 #![allow(unused_qualifications)]
+#![allow(deprecated)]
 
 use super::epoll;
 use crate::backend::c;
@@ -80,6 +81,7 @@ use crate::fd::{AsFd, OwnedFd};
 use crate::io;
 #[cfg(feature = "alloc")]
 use alloc::vec::Vec;
+pub use buf::EventBuffer;
 use core::ffi::c_void;
 use core::hash::{Hash, Hasher};
 use core::slice;
@@ -191,20 +193,15 @@ pub fn delete(epoll: impl AsFd, source: impl AsFd) -> io::Result<()> {
 #[cfg(feature = "alloc")]
 #[cfg_attr(docsrs, doc(cfg(feature = "alloc"), alias = "epoll_wait"))]
 #[inline]
-pub fn wait(epoll: impl AsFd, event_list: &mut EventVec, timeout: c::c_int) -> io::Result<()> {
-    // SAFETY: We're calling `epoll_wait` via FFI and we know how it
-    // behaves.
+pub fn wait<B: EventBuffer>(
+    epoll: impl AsFd,
+    mut events: B,
+    timeout: c::c_int,
+) -> io::Result<B::Out> {
     unsafe {
-        event_list.events.clear();
-        let nfds = syscalls::epoll_wait(
-            epoll.as_fd(),
-            event_list.events.spare_capacity_mut(),
-            timeout,
-        )?;
-        event_list.events.set_len(nfds);
+        let nfds = syscalls::epoll_wait(epoll.as_fd(), events.convert(buf::Internal), timeout)?;
+        Ok(events.filled(nfds, buf::Internal))
     }
-
-    Ok(())
 }
 
 /// An iterator over the `Event`s in an `EventVec`.
@@ -342,6 +339,7 @@ struct SixtyFourBitPointer {
 
 /// A vector of `Event`s, plus context for interpreting them.
 #[cfg(feature = "alloc")]
+#[deprecated(note = "Use an array or vec directly instead.")]
 pub struct EventVec {
     events: Vec<Event>,
 }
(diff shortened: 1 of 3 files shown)


SUPERCILEX:
Ok, I think the remaining questions are:
- Should we clear events in the `Vec` API? I have a slight preference against this, but I can also see it leading to weird bugs.
- What do we do if the Vec has zero capacity? Do we auto-alloc for the user or let the syscall error out. I'm leaning towards let the syscall error out, but again this could be confusing so I could see wanting to make the API friendlier.

src/event/epoll.rs
```diff
@@ -0,0 +1,495 @@
...
+    check_renamed_struct_renamed_field!(Event, epoll_event, data, data);
+}
+
+mod buf {
+    use super::Event;
+    #[cfg(feature = "alloc")]
+    use alloc::vec::Vec;
+    use core::mem::MaybeUninit;
+
+    pub trait EventBuffer {
```
notgull, line 444: Probably good to have an impl for `&mut [Event]` as well.

src/event/epoll.rs
```diff
@@ -0,0 +1,495 @@
...
+/// For each event of interest, an element is written to `events`. On
+/// success, this returns the number of written elements.
+///
+/// # References
+///
+///  - [Linux]
+///
+/// [Linux]: https://man7.org/linux/man-pages/man2/epoll_wait.2.html
+#[inline]
+pub fn wait<B: buf::EventBuffer>(
```
notgull, line 195: I would keep the `epoll_` prefix here. It's a bit weird to have freestanding functions just named "wait".

notgull (changes requested):
Big fan of the idea here, but a few notes.

src/event/epoll.rs
SUPERCILEX: I think the idea is to call it with the module prefix like `epoll::wait` (implemented here: https://github.com/bytecodealliance/rustix/commit/ac4a66f05f2976b21a603135da31eb9d04f6b2a5). But anyway, can't change that as it would be a breaking change.

src/event/epoll.rs
SUPERCILEX: Sounds good, done.

src/event/epoll.rs
SUPERCILEX: Hmmm, looking at the rest of our APIs I think yours is more consistent. Filed https://github.com/bytecodealliance/rustix/issues/753#issuecomment-2292528359

src/event/epoll.rs
```diff
@@ -0,0 +1,508 @@
...
+#[inline]
+#[doc(alias = "epoll_create1")]
+pub fn create(flags: CreateFlags) -> io::Result<OwnedFd> {
+    syscalls::epoll_create(flags)
+}
+
+/// `epoll_ctl(self, EPOLL_CTL_ADD, data, event)`—Adds an element to an epoll
+/// object.
+///
+/// This registers interest in any of the events set in `events` occurring on
```
sunfishcode, line 105: This is a pre-existing bug, but this should say `event_flags` instead of `events`.

sunfishcode (commented):
I need some more time to think about this.

The EventBuffer trait as a way to migrate away from EventVec and toward just using a plain Vec or array makes sense.

That said, I have reservations about the use of `B::Out` return value, which means that the value you get back, which determines how you use this function, changes significantly depending on what type of thing you pass in, and it's not clear from the public interface how users would figure out this relationship.

I like how eg. `read` and `read_uninit` have intuitive signatures that don't require you to learn about a trait and look up what types implement it, and I like how they give simple error messages if you use the wrong type.

src/event/epoll.rs
SUPERCILEX: Fixed.

SUPERCILEX:
Aw, I thought the dependent type theory stuff was the cool part. :grin: I do agree that it's pretty high on the complexity scale though. I've tried tweaking it so the docs are actually usable:
image

And the error looks like this which is honestly pretty decent IMO:
```
error[E0277]: the trait bound `{integer}: EventBuffer` is not satisfied
   --> src/event/mod.rs:33:22
    |
33  |     epoll::wait(CWD, 0, 0);
    |     -----------      ^ the trait `EventBuffer` is not implemented for `{integer}`
    |     |
    |     required by a bound introduced by this call
    |
    = help: the following other types implement trait `EventBuffer`:
              &'a mut [Event]
              &'a mut [MaybeUninit<Event>]
              &mut EventVec
              &mut Vec<Event>
```

notgull:
cc https://github.com/bytecodealliance/rustix/pull/908, which implemented a trait similar to `EventBuffer` but more generic in scope

SUPERCILEX:
Doh! I'd totally forgotten about that, sorry. And here I thought this was a new idea lol.

So if this space was already explored, then maybe we should just add a `wait_uninit` method instead and expect people to use `spare_capacity_mut`? I'm not a huge fan of that approach for Vecs because it means the caller has to take on the unsafety of setting the length if they want to pass the vec along rather than using the returned slice.

If we do want to try the fancy pantsy approach, I can make it look a little more like @notgull's original proposal while maintaining the Internal type so nobody can use it but us.

SUPERCILEX:
I've been working with `getxattr` and @notgull's approach seems more and more convenient. It'd be really nice to be able to pass in both a MaybeUninit and plain buffer depending on the needs. I think I'm in favor of resurrecting some variant of the dependent typing stuff... thoughts?

notgull:
I still advocate for my `Buffer` trait approach. It simplifies and extends so many use cases.

sunfishcode:
I've now updated #1290 to implement the basic idea here.

sunfishcode:
I'm going to go ahead and close this as #1290 contains the generalization of this code for general-purpose use. Thanks for the idea here, and thanks for your patience!

SUPERCILEX:
Sounds good, I'm still reviewing #1290.

PR #1112, opened 2024-08-15
Remove incorrect mount API and allow null data arg

SUPERCILEX:
Added in error in https://github.com/bytecodealliance/rustix/pull/1024 due to a misunderstanding of how the API works. I've made a tweak to the existing API to allow passing in null options (should be backwards compatible!) as some poorly designed FS might require this, but in practice you could have just passed in the empty string.

### DIFF
--- a/src/mount/mount_unmount.rs
+++ b/src/mount/mount_unmount.rs
@@ -14,12 +14,12 @@ use crate::{backend, io};
 ///
 /// [Linux]: https://man7.org/linux/man-pages/man2/mount.2.html
 #[inline]
-pub fn mount<Source: path::Arg, Target: path::Arg, Fs: path::Arg, Data: path::Arg>(
+pub fn mount<'a, Source: path::Arg, Target: path::Arg, Fs: path::Arg>(
     source: Source,
     target: Target,
     file_system_type: Fs,
     flags: MountFlags,
-    data: impl Into<Option<Data>>,
+    data: impl Into<Option<&'a CStr>>,
 ) -> io::Result<()> {
     source.into_with_c_str(|source| {
         target.into_with_c_str(|target| {
@@ -33,41 +33,6 @@ pub fn mount<Source: path::Arg, Target: path::Arg, Fs: path::Arg, Data: path::Ar
                         data,
                     )
                 })
-            })
-        })
-    })
-}
-
-/// `mount2(source, target, filesystemtype, mountflags, data)`
-///
-/// This is same as the [`mount`], except it adds support for the source,
-/// target, and data being omitted, and the data is passed as a `CStr` rather
-/// than a `path::Arg`.
-///
-/// # References
-///  - [Linux]
-///
-/// [Linux]: https://man7.org/linux/man-pages/man2/mount.2.html
-#[inline]
-#[deprecated(note = "This API was added in error, use the expressive mount APIs instead.")]
-#[doc(hidden)]
-pub fn mount2<Source: path::Arg, Target: path::Arg, Fs: path::Arg>(
-    source: Option<Source>,
-    target: Target,
-    file_system_type: Option<Fs>,
-    flags: MountFlags,
-    data: Option<&CStr>,
-) -> io::Result<()> {
-    option_into_with_c_str(source, |source| {
-        target.into_with_c_str(|target| {
-            option_into_with_c_str(file_system_type, |file_system_type| {
-                backend::mount::syscalls::mount(
-                    source,
-                    target,
-                    file_system_type,
-                    MountFlagsArg(flags.bits()),
-                    data,
-                )
             })
         })
     })
--- a/src/mount/mount_unmount.rs
+++ b/src/mount/mount_unmount.rs
@@ -19,18 +19,18 @@ pub fn mount<Source: path::Arg, Target: path::Arg, Fs: path::Arg, Data: path::Ar
     target: Target,
     file_system_type: Fs,
     flags: MountFlags,
-    data: Data,
+    data: impl Into<Option<Data>>,
 ) -> io::Result<()> {
     source.into_with_c_str(|source| {
         target.into_with_c_str(|target| {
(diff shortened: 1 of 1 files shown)


morr0ne:
I have to disagree with the assertion that the API introduced in #1024 is flawed. The current implementation adheres to the documented requirements and allows necessary functionality. It is ultimately the responsibility of the user to utilize the kernel apis appropriately. The crate offers safe wrappers, and the mount2 function does not introduce unsound behavior.

That said, I am not against this changes to the api, provided they maintain backward compatibility. However, it appears that deprecating mount2 offers no tangible benefit.

Additionally, I would like to point out that I am unsure whether the proposed use of impl syntax is supported by the current msrv. As such, this change may not actually be backward compatible.

SUPERCILEX:
> It is ultimately the responsibility of the user to utilize the kernel apis appropriately.

That's not how rustix works: we don't add APIs that directly mirror the kernel syscalls. Instead, rustix APIs are "actions" that may or may not use the same underlying syscall. For examples, look at epoll, fcntl, ioctl, the entire net module, etc etc etc. If you want user responsibility, use C. :)

> impl syntax

We've been using the impl trait syntax for ages, it's fine.

morr0ne:
> That's not how rustix works: we don't add APIs that directly mirror the kernel syscalls. Instead, rustix APIs are "actions" that may or may not use the same underlying syscall. For examples, look at epoll, fcntl, ioctl, the entire net module, etc etc etc. If you want user responsibility, use C. :)

I am familiar with rustix's functionality. However, I fail to identify any significant flaws in the mount2 api. The suggestion to switch to C is neither helpful nor professional. I added an api tailored to an use case which rustix did not address. These changes were approved, and I have not been presented with sufficient evidence to justify their deprecation.

SUPERCILEX:
`mount2` is a raw API and in the same spirit as writing C. There is no valid mount configuration where `source` and `file_system_type` are null, therefore it is not an appropriate rustix API.

SUPERCILEX:
One thing I forgot to address is the `path::Arg` to `&CStr` change. I think `&CStr` is indeed more correct from a theoretical standpoint but annoying in practice. I'd argue `path::Arg` is a bad name as it's really just a "gimme CStr plz" trait which means the current API allows passing in a byte array, whereas if we go with the more theoretically correct `&CStr` approach, users will be on their own when it comes to conversions.

SUPERCILEX:
Are there other examples of c-like rustix interfaces? Off the top of my head an obvious one is open, but there's an item in the 1.0 release to consider splitting it between create and open so the Mode argument isn't meaningless in non-create cases. In general, I think rustix should strive to avoid meaningless arguments as much as possible—IIRC the nix crate was more focused on matching the c APIs so I kinda see folks using either crate depending on their preferred approach.

I don't have a strong preference regarding not using `path::Arg`, but I do think it would be quite unfortunate to lose easy byte to CStr conversions. Maybe `Arg` should just be moved to a `convert` module? And maybe pull out a super trait that doesn't have PathBuf/Path conversions (not sure that's necessary)?

SUPERCILEX:
Actually I kind of like the idea of leaving `path::Arg` where it is and creating a super trait `convert::Arg` (or maybe `ffi::Arg`) which has implementations for everything not in `std::path` (so no components, DecInt, etc). That makes the path vs not distinction clearer while still giving you useful conversions.

Also I just saw the futex rework landed! That further reinforces the idea that rustix is moving away from c like APIs—`mount2` is a step backwards.

src/mount/mount_unmount.rs
```diff
@@ -19,18 +19,18 @@ pub fn mount<Source: path::Arg, Target: path::Arg, Fs: path::Arg, Data: path::Ar
     target: Target,
     file_system_type: Fs,
     flags: MountFlags,
-    data: Data,
+    data: impl Into<Option<Data>>,
```
sunfishcode, line 22: I'm curious why this is `impl Into<Option<...>>` instead of just `Option<...>`.

src/mount/mount_unmount.rs
```diff
@@ -49,6 +49,8 @@ pub fn mount<Source: path::Arg, Target: path::Arg, Fs: path::Arg, Data: path::Ar
 ///
 /// [Linux]: https://man7.org/linux/man-pages/man2/mount.2.html
 #[inline]
+#[deprecated(note = "This API was added in error, use the expressive mount APIs instead.")]
+#[doc(hidden)]
```
sunfishcode, line 53: Since we're nearing a 1.0 release, we can just delete this function instead of deprecating it.

src/mount/mount_unmount.rs
sunfishcode, line 22: Also, could you change `Data` to `&CStr`?

I know that's not quite as fun as `path::Arg`, but as discussed earlier in the issue, that's for filesystem paths. The idea of splitting an `ffi::Arg` out seems fun, but there seem to be very few things in rustix that would use it, so it's not clearly worth it. And now that Rust has C-string literals, plain `&CStr` is pretty ergonomic in common cases.

sunfishcode (commented):
I think this makes sense to do.

src/mount/mount_unmount.rs
SUPERCILEX: Just because it's annoying to write `Some(thing)` when you're primarily going to pass in `thing` all the time. I don't have a strong opinion there though.

SUPERCILEX: Fair enough, done.

sunfishcode: Ah, for some reason it didn't occur to me that you could do that. This sounds fine to me.

sunfishcode:
Thanks!

SUPERCILEX:
Oh wait shoot, sorry I forgot to move the `impl` to a normal generic.

SUPERCILEX:
Fixed in https://github.com/bytecodealliance/rustix/pull/1317

Issue #538, opened 2023-02-16
Feature request: Support for inotify


VorpalBlade:
Having support for `inotify` would be nice. I only target Linux in my current project, but perhaps it would make sense to abstract this across `inotify` and whatever the *BSDs and OS X use.

sunfishcode:
https://github.com/bytecodealliance/rustix/pull/479 added inotify support; it's in the 0.37.0-alpha.0 prerelease and will be in a 0.37 soon.

VorpalBlade:
Oh, sorry, I just looked on docs.rs for it, was it wasn't there yet. My bad! Closing the issue.

VorpalBlade:
Actually, reading that PR it seems to be just the low level API, not a high level API like epoll has at https://docs.rs/rustix/latest/rustix/io/epoll/struct.Epoll.html

So reopening for now?

VorpalBlade:
Clarifying the above comment a bit:

For a rust-ified API I would expect something like a function to read an iterator of pending inotify events, not having to read the raw inotify bytes from the fd and convert it into a `repr(C)` struct myself, which is what PR #479  seems to have added.

nix has this for example: https://docs.rs/nix/latest/nix/sys/inotify/struct.Inotify.html

sunfishcode:
My sense is that an API like `read_events` returning a `Vec<InotifyEvent>`, or an iterator that would want to implicitly manage a buffer, is more high-level than makes  sense for rustix. Higher-level APIs are already pretty well covered by eg. the [inotify crate](https://lib.rs/crates/inotify) or the higher-level [notify crate](https://lib.rs/crates/notify) and others.

It is true that rustix did add a somewhat higher-level Epoll, but that was motivated by pursuit of a particular I/O safety property, and ended up being awkward. In general, my thought is to keep rustix fairly low-level where possible, and encourage higher-level APIs to live in separate crates.

sunfishcode:
Inotify support is now released in 0.37.

SUPERCILEX:
@sunfishcode would you be opposed to a raw dir style iterator implementation? Seems like the same ish API just with different fields.

sunfishcode:
@SUPERCILEX Would that be an iterator over `inotify_event`s? If so then yeah, that'd be good to have, as they contain fds, so an iterator could add I/O safety.

SUPERCILEX:
[...]

PR #1113, opened 2024-08-16
Implement InotifyReader for raw inotify event iteration

SUPERCILEX:
From https://github.com/bytecodealliance/rustix/issues/538#issuecomment-2076539826


sunfishcode:
This looks good; would it be practical to add a testcase that exercises the basic functionality?

sunfishcode:
Thanks! I think that test looks good enough for now.

sunfishcode:

PR #1115, opened 2024-08-17
Add io_uring flags available in linux-raw-sys

ryanseipp:
Adds flags to io_uring that are available in the current version of linux-raw-sys.

* `IORING_SETUP_NO_SQARRAY`: Linux 6.6
* `IORING_FEAT_REG_REG_RING`: Linux 6.3
* `IORING_REGISTER_USE_REGISTERED_RING`: Linux 6.3
* `xattr_flags`: https://github.com/axboe/liburing/blob/liburing-2.7/src/include/liburing/io_uring.h#L69

There are more flags available, like new operations and a new feature, but I'm unsure what it'd take to get them into linux-raw-sys, in a way this project can consume. I believe they were recently added in Linux 6.10 though, so I'm not in any rush to get those out the door.

### DIFF
--- a/src/io_uring.rs
+++ b/src/io_uring.rs
@@ -35,7 +35,9 @@ use linux_raw_sys::net;
 pub use crate::event::epoll::{
     Event as EpollEvent, EventData as EpollEventData, EventFlags as EpollEventFlags,
 };
-pub use crate::fs::{Advice, AtFlags, Mode, OFlags, RenameFlags, ResolveFlags, Statx, StatxFlags};
+pub use crate::fs::{
+    Advice, AtFlags, Mode, OFlags, RenameFlags, ResolveFlags, Statx, StatxFlags, XattrFlags,
+};
 pub use crate::io::ReadWriteFlags;
 pub use crate::net::{RecvFlags, SendFlags, SocketFlags};
 pub use crate::timespec::Timespec;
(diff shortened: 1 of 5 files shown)


src/io_uring.rs
```diff
@@ -894,6 +902,9 @@ pub const IORING_OFF_CQ_RING: u64 = sys::IORING_OFF_CQ_RING as _;
 #[allow(missing_docs)]
 pub const IORING_OFF_SQES: u64 = sys::IORING_OFF_SQES as _;
 
+/// Indicate that the fd passed to [`io_uring_register`] is the index of a registered ring fd.
+pub const IORING_REGISTER_USE_REGISTERED_RING: u32 = sys::IORING_REGISTER_USE_REGISTERED_RING as _;
```
sunfishcode, line 906: As I understand it, this constant doesn't work with the current API because `io_uring_register` take as `IoringRegisterOp` which is an enum and doesn't support additional flags being or'd in.

Would it work to add a new `io_uring_register_with` function which is just like `io_uring_register` except that it has an extra `flags: IoringRegisterFlags` argument, where `IoringRegisterFlags` is a new `bitflags` type that currently just contains one flag, `USE_REGISTERED_RING`, and then the implementation could or that flags into the opcode value?

sunfishcode (commented):
Thanks for submitting this! Overall this looks good; just one comment:

sunfishcode:
Thanks!

sunfishcode:

PR #1127, opened 2024-08-24
Implement the new module structure for shm and inotify

SUPERCILEX:
Let's help folks transition before 1.0.

### DIFF
--- a/src/backend/libc/shm/types.rs
+++ b/src/backend/libc/shm/types.rs
@@ -2,9 +2,9 @@ use crate::backend::c;
 use bitflags::bitflags;
 
 bitflags! {
-    /// `O_*` constants for use with [`open`].
+    /// `O_*` constants for use with [`shm::open`].
     ///
-    /// [`open`]: crate:shm::open
+    /// [`shm::open`]: crate:shm::open
     #[repr(transparent)]
     #[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
     pub struct ShmOFlags: u32 {
--- a/src/backend/libc/shm/types.rs
+++ b/src/backend/libc/shm/types.rs
@@ -2,9 +2,9 @@ use crate::backend::c;
 use bitflags::bitflags;
 
 bitflags! {
-    /// `O_*` constants for use with [`shm_open`].
+    /// `O_*` constants for use with [`open`].
     ///
-    /// [`shm_open`]: crate:shm::shm_open
+    /// [`open`]: crate:shm::open
     #[repr(transparent)]
     #[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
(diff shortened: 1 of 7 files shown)


src/backend/libc/shm/types.rs
```diff
@@ -2,9 +2,9 @@ use crate::backend::c;
 use bitflags::bitflags;
 
 bitflags! {
-    /// `O_*` constants for use with [`shm_open`].
+    /// `O_*` constants for use with [`open`].
```
sunfishcode, line 5: I think it'd help avoid confusion to change the text here from \`open\` to \`shm::open\`. I think that matches how users will want to use it, as `use rustix::shm; ... shm::open` rather than `use rustix::shm::open; ... open`.

And similar for \`init\` and other functions in this PR.

src/shm.rs
```diff
@@ -21,8 +30,9 @@ pub use crate::backend::shm::types::ShmOFlags;
 ///
 /// [POSIX]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/shm_open.html
 /// [Linux]: https://man7.org/linux/man-pages/man3/shm_open.3.html
+#[doc(alias = "shm_open")]
 #[inline]
-pub fn shm_open<P: path::Arg>(name: P, flags: ShmOFlags, mode: Mode) -> io::Result<OwnedFd> {
+pub fn open<P: path::Arg>(name: P, flags: OFlags, mode: Mode) -> io::Result<OwnedFd> {
```
sunfishcode, line 35: What would you think of doing something like `use crate::shm;` at the top of the file here, and then changing `OFlags` to `shm::OFlags`, so that users always see the `OFlags` presented to them as `shm::OFlags`?

sunfishcode (commented):
Seeing this change in context, I'm a little worried about potential confusion and maybe even import conflicts between `shm::open` and `fs::open`.

The reason I like this style is that I like the `use rustix::shm;` approach where the `shm::` qualifier is used for everything in the `shm` module. However, that is kind of in conflict with how a lot of people use other things in rustix. For example, most code I've seen does `use rustix::fs::open; ... open` rather than `use rustix::fs; ... fs::open`. So I wonder if there's something we can do to signal that `inotify`, `shm` and maybe also `futex` are meant to be used in this different style.

src/backend/libc/shm/types.rs
SUPERCILEX: Seems like a good improvement! This confusion is part of the reason I prefer `shm_open` (I also think autocomplete support is better with just the function name to deal with), but if we display all uses in the docs with the module prefix then that probably gets us most of the way there.

SUPERCILEX: Yeah ok this looks a lot better now, nice!

sunfishcode:

PR #1132, opened 2024-08-26
Add more `epoll::` qualifiers to epoll's public API.

sunfishcode:
Qualify more types and functions in epoll's public API with `epoll::` qualifiers.


sunfishcode:

PR #1147, opened 2024-09-03
Fall back to `TCGETS` if `TCGETS2` fails

sunfishcode:
Switch `tcgetattr`/`tcsetattr` from using `TCGETS2`/`TCSETS2` first to using `TCGETS`/`TCSETS` first. Have `tcgetattr` fall back to `TCGETS2` if the `TCGETS` flags indicate that a custom speed is in used.

glibc and musl have not yet migrated to `TCGETS2`/`TCSETS2`, and as a result, seccomp sandboxes and Linux-like environments such as WSL don't always support them.

Also, fix some bugs in QEMU related to the handling of termios syscalls. This eliminates the need for having rustix do extra fixups on PowerPC.

This is expected to fix crossterm-rs/crossterm#912.


sunfishcode:
This is now released in rustix 0.38.36.

PR #1155, opened 2024-09-07
Add `io_uring_buf_ring` struct

### LINKED ISSUE #1143: Feature request: Export `io_uring_buf_ring`
io_uring utilizes a struct called `io_uring_buf_ring` to register a ring buffer of `io_uring_buf` structs that contain a reference to slices of buffers that can be filled in certain operations. One operation that can utilize this is `IoringOp::Recv` with the `IoringRecvFlags::MULTISHOT` flag set. This can make for much more performant operations, especially for those using direct io, as the kernel can choose a buffer to utilize, allowing the application to have many more `recv` operations in flight than available buffers.

This struct was added to linux 5.19, and is defined [here](https://github.com/torvalds/linux/blob/v5.19/include/uapi/linux/io_uring.h#L507).

`linux_raw_sys` 0.4 exports this struct [here](https://github.com/sunfishcode/linux-raw-sys/blob/0.4/src/x86_64/io_uring.rs#L400), which contains an anonymous union and C flexible array member. `libc` does not export this type.

I see we hand-code structs instead of re-exporting them, seemingly because `libc` doesn't export those structs. So, I'm primarily looking for guidance on how to define the flexible array member. I'd be happy to contribute this struct if someone doesn't go for it first.

ryanseipp:
Closes #1143.

Adds a new type `io_uring_buf_ring`, defined in the kernel [here](https://github.com/torvalds/linux/blob/v5.19/include/uapi/linux/io_uring.h#L507) as:
```c
struct io_uring_buf_ring {
	union {
		/*
		 * To avoid spilling into more pages than we need to, the
		 * ring tail is overlaid with the io_uring_buf->resv field.
		 */
		struct {
			__u64	resv1;
			__u32	resv2;
			__u16	resv3;
			__u16	tail;
		};
		struct io_uring_buf	bufs[0];
	};
};
```

linux-raw-sys exports the type [here](https://github.com/sunfishcode/linux-raw-sys/blob/0.4/src/x86_64/io_uring.rs#L400), so we can verify alignment and sizes of the types. Additionally, an integration test was added to ensure we can use the structs, anonymous union type, and flexible array member correctly.


sunfishcode:
Looks good, thanks!

sunfishcode:
This is now released in rustix 0.38.41.

PR #1159, opened 2024-09-09
Fix `recv_uninit` to handle the `TRUNC` flag.

### LINKED ISSUE #1153: `recvmsg_uninit` does not work with `MSG_TRUNC`
[`recvfrom(2)`](https://man7.org/linux/man-pages/man2/recvfrom.2.html) has a `MSG_TRUNC` flag that makes it "return the real length of the packet or datagram, even when it was longer than the passed buffer."

This leads to panic since `rustix::net::recvmsg_uninit` tries to split the return buffer according to the returned length.

```rs
// Pass in a really small buffer
let (buf, _, _) = rustix::net::recvfrom_uninit(&socket, &mut buf, RecvFlags::TRUNC)?;
```

```
thread 'main' panicked at ~/.cargo/git/checkouts/rustix-8fafe793515fde97/9b432db/src/buffer.rs:18:30:
mid > len
stack backtrace:
   2: core::slice::<impl [T]>::split_at_mut
             at /rustc/0d634185dfddefe09047881175f35c65d68dcff1/library/core/src/slice/mod.rs:1906:21
   3: rustix::buffer::split_init
             at ~/.cargo/git/checkouts/rustix-8fafe793515fde97/9b432db/src/buffer.rs:18:26
   4: rustix::net::send_recv::recvfrom_uninit
             at ~/.cargo/git/checkouts/rustix-8fafe793515fde97/9b432db/src/net/send_recv/mod.rs:169:35
```

`rustix::net::recvfrom` doesn't suffer from this issue since it just returns the length directly. But it also doesn't support uninit buffers.

sunfishcode:
With the `TRUNC` flag, `recv` returns the untruncated length, which may be longer than the provided buffer. Fix `recv_uninit` and `recvfrom_uninit` to handle this case.

This mean that `recv_uninit` and `recvfrom_uninit` are unable to return the full untruncated length returned by `recv` with the `TRUNC` flag, however it appears that fixing that may require API changes.

Fixes #1153.

### DIFF
--- a/src/net/send_recv/mod.rs
+++ b/src/net/send_recv/mod.rs
@@ -10,6 +10,7 @@ use crate::net::SocketAddrUnix;
 use crate::net::{SocketAddr, SocketAddrAny, SocketAddrV4, SocketAddrV6};
 use crate::{backend, io};
 use backend::fd::{AsFd, BorrowedFd};
+use core::cmp::min;
 use core::mem::MaybeUninit;
 
 pub use backend::net::send_recv::{RecvFlags, SendFlags};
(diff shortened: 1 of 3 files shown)


andylizi:
While `recvfrom` indeed works, would this be a good candidate for #753?

I feel that avoiding zero-initialization is especially advantageous in this case, because UDP socket buffers are often quite large. Their recommended size can even reach multiple megabytes (as seen in [quic-go](https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes), for example).

sunfishcode:
This is now released in rustix 0.38.37.

@andylizi Sounds good, I've now added a todo box to #753 for it.

PR #1162, opened 2024-09-14
Implement `select`.

sunfishcode:
Add a `select` function, defined only on platforms where it doesn't have an `FD_SETSIZE` limitation.

### DIFF
--- a/src/backend/libc/event/windows_syscalls.rs
+++ b/src/backend/libc/event/windows_syscalls.rs
@@ -2,7 +2,7 @@
 
 use crate::backend::c;
 use crate::backend::conv::ret_c_int;
-use crate::event::PollFd;
+use crate::event::{FdSetElement, PollFd};
 use crate::io;
 
 pub(crate) fn poll(fds: &mut [PollFd<'_>], timeout: c::c_int) -> io::Result<usize> {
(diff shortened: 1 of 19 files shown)


src/backend/libc/event/windows_syscalls.rs
```diff
@@ -14,3 +15,63 @@ pub(crate) fn poll(fds: &mut [PollFd<'_>], timeout: c::c_int) -> io::Result<usiz
     ret_c_int(unsafe { c::poll(fds.as_mut_ptr().cast(), nfds, timeout) })
         .map(|nready| nready as usize)
 }
+
+pub(crate) fn select(
+    nfds: i32,
+    readfds: Option<&mut [FdSetElement]>,
+    writefds: Option<&mut [FdSetElement]>,
+    exceptfds: Option<&mut [FdSetElement]>,
```
notgull, line 23: Is this tested on Win32? It looks like it has a [slightly different](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/ns-winsock2-fd_set) implementation of `FD_SET` than Unix does, that isn't implemented in the bitfield way.

sunfishcode: Ooh, good spot. I had not yet tested it on Windows. I'll just disable Windows support for now.

notgull: I'm worried that the current way it's implemented would make Windows support impossible in the future without a breaking change.

sunfishcode: Yes. I was thinking of just skipping the Windows tax here until someone shows up with a use case that needs it. If that happens, we could do like a `select_ng` or whatever until the next semver break.

sunfishcode: I changed my mind here, and decided to re-introduce Windows support, partly because WASI also uses a count+array fd_set like Windows does, and partly because I thought of a way to support this that doesn't radically change the public API.

PR #1164, opened 2024-09-15
Fix `fcntl_setpipe_size` return value

### LINKED ISSUE #1163: fcntl_setpipe_size returns pipe size on success, not 0
```
rustix::pipe::fcntl_setpipe_size(&w, 1024 * 1024).unwrap();
```
is reported to fail, due to the return value is 1024 * 1024 = 1048576.

jjyyxx:
`fcntl` man page says that:
> F_SETPIPE_SZ ... The actual capacity (in bytes) that is set is
> returned as the function result.

But the current `fcntl_setpipe_size` function assumes that the return value is 0 (success) or negative (failure). This makes the function panic due to `debug_assert!` in debug mode, and it returns `Err` on success in release mode.

This commit fixes the return value type and its checking, and adds a test for it.

Closes #1163


sunfishcode:
Thanks!

Issue #1156, opened 2024-09-09
`sendmmsg` / `recvmmsg`


colinmarc:
It would be great to support `sendmmsg` and `recvmmsg` in `rustix::net`.

sunfishcode:
As a heads up to anyone working on this, musl and glibc have different types for the msg_iovlen, msg_controllen, and cmsg_len fields, which is something to watch out for.

PR #1171, opened 2024-09-18
Add support for sendmmsg(2) on linux

colinmarc:
https://man7.org/linux/man-pages/man2/sendmmsg.2.html

Partially addresses #1156. I would've liked to add `recvmmsg` in the same PR, but it's actually much more complicated.


colinmarc:
My test is unfortunately flaky, because of course the kernel can consolidate stream messages. Will fix tomorrow.

src/net/send_recv/msg.rs
```diff
@@ -591,6 +597,74 @@ impl<'buf> Iterator for AncillaryDrain<'buf> {
...
+#[repr(transparent)]
+pub struct MMsgHdr<'a> {
+    raw: c::mmsghdr,
+    _phantom: PhantomData<&'a mut ()>,
+}
+
+#[cfg(target_os = "linux")]
+impl<'a> MMsgHdr<'a> {
+    /// Constructs a new message with no destination address.
+    pub fn new(iov: &[IoSlice<'a>], control: &mut SendAncillaryBuffer<'_, '_, '_>) -> Self {
```
sunfishcode, line 612: Can you comment on why `control` needs to be a `mut` reference here?

tests/net/v4.rs
```diff
@@ -194,3 +196,94 @@ fn test_v4_msg() {
     client.join().unwrap();
     server.join().unwrap();
 }
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_v4_sendmmsg() {
```
sunfishcode, line 202: Could you also add an equivalent test using IPv6 to v6.rs?

src/net/send_recv/msg.rs
```diff
@@ -591,6 +597,74 @@ impl<'buf> Iterator for AncillaryDrain<'buf> {
...
+            raw: c::mmsghdr {
+                msg_hdr,
+                msg_len: 0,
+            },
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Returns the number of bytes sent by [sendmmsg].
+    pub fn bytes(&self) -> usize {
```
sunfishcode, line 663: `bytes()` is often the name of a function that returns a `&[u8]` or similar. What would you think about renaming this to `num_bytes_sent()`?

src/net/send_recv/msg.rs
```diff
@@ -591,6 +597,74 @@ impl<'buf> Iterator for AncillaryDrain<'buf> {
...
+        Self {
+            raw: c::mmsghdr {
+                msg_hdr,
+                msg_len: 0,
+            },
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Returns the number of bytes sent by [sendmmsg].
```
sunfishcode, line 662: Could you add a comment mentioning that this isn't meaningful to call until after the `sendmmsg` call?

src/net/send_recv/msg.rs
colinmarc: Sure - I think I was just trying to match [`RecvMsgReturn::bytes`](https://docs.rs/rustix/latest/rustix/net/struct.RecvMsgReturn.html#structfield.bytes)

colinmarc:
I think I addressed all your comments.

src/net/send_recv/msg.rs
```diff
@@ -591,6 +597,75 @@ impl<'buf> Iterator for AncillaryDrain<'buf> {
...
+        with_noaddr_msghdr(iov, control, Self::wrap)
+    }
+
+    /// Constructs a new message to a specific IPv4 address.
+    pub fn new_v4(
+        addr: &SocketAddrV4,
+        iov: &[IoSlice<'a>],
+        control: &mut SendAncillaryBuffer<'_, '_, '_>,
+    ) -> Self {
+        with_v4_msghdr(addr, iov, control, Self::wrap)
```
notgull, line 622: Isn't this unsound? `with_v4_msghdr` allocates the V4 socket address on the stack and then passes that stack address in the `msghdr` in `Self::wrap`. So when we eventually call into the system call it is pointing to invalid stack memory.

notgull (changes requested):

src/net/send_recv/msg.rs
colinmarc: Yes, you're right. Maybe `SockAddrStorage` can be used for this?

colinmarc: In order to avoid passing the length in here, we'll need to add a `RawSockAddr` type, I think, created with `SockAddrV4::as_raw` etc. It'll be slightly unergonomic, as the borrow checker won't let you do this:

```rust
MMsgHdr::new_with_addr(my_v4_addr.as_raw(), ...)
```

Since that would create a temporary value which is dropped. But it should work fine otherwise.

src/net/send_recv/msg.rs
```diff
@@ -591,6 +597,75 @@ impl<'buf> Iterator for AncillaryDrain<'buf> {
...
+#[repr(transparent)]
+pub struct MMsgHdr<'a> {
+    raw: c::mmsghdr,
+    _phantom: PhantomData<&'a mut ()>,
+}
+
+#[cfg(target_os = "linux")]
+impl<'a> MMsgHdr<'a> {
+    /// Constructs a new message with no destination address.
+    pub fn new(iov: &[IoSlice<'a>], control: &mut SendAncillaryBuffer<'_, '_, '_>) -> Self {
```
colinmarc, line 612: Just leaving a note for myself that I think the lifetimes are also being elided incorrectly here - because technically by using `Self` there's no elision on the output lifetime. So I think this works out to:

```rust
pub fn new(iov: &'x [IoSlice<'a>], control: &'y mut SendAncillaryBuffer<'_, '_, '_>) -> MMsgHdr<'a> {}
```

src/net/send_recv/msg.rs
colinmarc: Add a `RawSocketAddr` type in my latest push (it's a separate commit, so can be reviewed independently).

colinmarc:
Hi, anything needed from me on this?

src/net/send_recv/msg.rs
```diff
@@ -781,6 +826,58 @@ pub fn sendmsg_any(
     }
 }
 
+/// `sendmsg(msghdr)`—Sends a message on a socket to a specific address.
```
sunfishcode, line 829:
```suggestion
/// `sendmsg(msghdr)`—Sends a message on a socket to a specific raw socket address.
```

src/backend/linux_raw/net/syscalls.rs
```diff
@@ -439,6 +445,67 @@ pub(crate) fn sendmsg_xdp(
     })
 }
 
+#[cfg(not(any(
+    windows,
+    target_os = "espidf",
+    target_os = "redox",
+    target_os = "vita",
+    target_os = "wasi"
+)))]
```
sunfishcode, line 454: This `cfg` attribute is unneeded because we're in the linux_raw backend here.

src/net/send_recv/msg.rs
```diff
@@ -781,6 +826,58 @@ pub fn sendmsg_any(
...
+/// [illumos]: https://illumos.org/man/3SOCKET/sendmsg
+#[inline]
+pub fn sendmsg_raw(
+    socket: impl AsFd,
+    addr: Option<&RawSocketAddr>,
+    iov: &[IoSlice<'_>],
+    control: &mut SendAncillaryBuffer<'_, '_, '_>,
+    flags: SendFlags,
+) -> io::Result<usize> {
+    match addr {
```
sunfishcode, line 857: Instead of matching `addr` here, would it work to have the backend `sendmsg_raw` and `with_raw_msghdr` take an `Option<&RawSocketAddr>` as well, and then in `with_raw_msghdr` convert the `Option` into a pointer for `msg_name` using `crate::utils::option_as_ptr`?

src/net/socket_addr_any.rs
```diff
@@ -107,6 +126,35 @@ impl SocketAddrAny {
     }
 }
 
+/// A raw sockaddr and its length.
+#[repr(C)]
```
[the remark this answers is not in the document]
colinmarc: As of your suggested change above, we do, when we use `option_as_ptr(addr)` with `addr` being `Option<&RawSocketAddr>`. I'm open to changing that to `addr.map(RawSocketAddr::as_ptr).unwrap_or_default()`, or maybe dropping the `as_ptr` helper, or some combination.

colinmarc:
Sorry for pinging and then disappearing - was really sick :)

I rebased and addressed your comments. Please let me know how you'd like to resolve the overlap with #1004.

colinmarc:
Rebased on top of #1004 now that it's in main. The test failures look unrelated.

This patch is a lot smaller now!

src/net/send_recv/msg.rs
colinmarc: I removed `RawSocketAddr` since something similar was upstreamed in #1004.

src/net/send_recv/msg.rs
```diff
@@ -591,6 +593,48 @@ impl<'buf> Iterator for AncillaryDrain<'buf> {
...
+#[cfg(target_os = "linux")]
+impl<'a> MMsgHdr<'a> {
+    /// Constructs a new message with no destination address.
+    pub fn new(iov: &'a [IoSlice<'_>], control: &'a mut SendAncillaryBuffer<'_, '_, '_>) -> Self {
+        with_noaddr_msghdr(iov, control, Self::wrap)
+    }
+
+    /// Constructs a new message to a specific address.
+    pub fn new_with_addr(
+        addr: &'a SocketAddrAny,
```
colinmarc, line 614: I would like to take a `&'a impl SockAddrArg` here, but it doesn't seem like the trait ensures that the pointer passed to the closure lives as long as `SockAddrArg` itself.

tests/net/v4.rs
```diff
@@ -194,3 +194,89 @@ fn test_v4_msg() {
...
+        let addr = SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), port);
+        let data_socket = socket(AddressFamily::INET, SocketType::STREAM, None).unwrap();
+        connect(&data_socket, &addr).unwrap();
+
+        let mut off = 0;
+        while off < 2 {
+            let sent = sendmmsg(
+                &data_socket,
+                &mut [
+                    MMsgHdr::new(&[IoSlice::new(b"hello")], &mut Default::default()),
```
sunfishcode, line 252: Should this use `new_wiith_addr`, like the v6 version does?

src/net/send_recv/msg.rs
sunfishcode: Yes, you're right. Could you add a comment about this? Something like, "this doesn't use `SocketAddrArg` because that creates a temporary `SocketAddrAny` that only lives for the duration of the `with_sockaddr` call, and we need a `SocketAddrAny` that lives for `'a`".

Another possibility would be to add a `with_socket_addr_any` function to `SocketAddrArg` like this:
```rust
    fn with_socket_addr_any<R>(&self, f: impl FnOnce(&SocketAddrAny) -> R) -> R {
        let any = self.as_any();
        f(&any)
    }
```
Then users with a family-specific `addr` could do
```rust
    addr.with_socket_addr_any(|addr| {
        sendmmsg(
            &socket,
            &mut [
                MMsgHdr::new_with_addr(
                    &addr,
                    &[IoSlice::new(b"hello")],
                    &mut Default::default(),
                ),
            ],
            SendFlags::empty(),
        )
    });
```
and then `SocketAddrAny`'s `impl SocketAddrArg` could skip the `as_any()` call:
```rust
    fn with_socket_addr_any<R>(&self, f: impl FnOnce(&SocketAddrAny) -> R) -> R {
        f(self)
    }
```
That way users could avoid calling `as_any` themselves. We could do that, though it's not that much of a simplification.

sunfishcode:
The FreeBSD CI failure is fixed on main in https://github.com/bytecodealliance/rustix/pull/1322.

tests/net/v4.rs
colinmarc: I think my idea was to exercise both code paths. I changed it so that both tests exercise both.

sunfishcode:
Thanks!

sunfishcode:

Issue #1051, opened 2024-04-24
Properly support `ioctl` that takes the last argument as an integer not a pointer


kennytm:
The ioctl `TUNSETOFFLOAD` takes an `unsigned int` as input rather than `unsigned int*`, so currently none of the standard `Ioctl` type work and must be worked-around via an awkward int-to-pointer cast.

<details><summary>Proof of concept</summary>

```rust
#!/usr/bin/env -S cargo +nightly -Zscript
---
[dependencies]
rustix = { version = "0.38.34", features = ["fs"] }
---

// Note: require root permission to run.

use rustix::{
    fs::{open, Mode, OFlags},
    io::Result,
    ioctl::{ioctl, CompileTimeOpcode, Ioctl, IoctlOutput, Opcode, Setter, WriteOpcode},
};
use std::{
    ffi::{c_int, c_uint, c_void},
    marker::PhantomData,
};

// define the ioctl constants
const TUN_F_CSUM: c_uint = 0x01;
type TUNSETIFF = WriteOpcode<b'T', 202, c_int>;
type TUNSETOFFLOAD = WriteOpcode<b'T', 208, c_uint>;

fn main() -> Result<()> {
    let fd = open(
        c"/dev/net/tun",
        OFlags::RDWR | OFlags::CLOEXEC,
        Mode::empty(),
    )?;
    println!("created fd: {:?}", fd);
    unsafe {
        ioctl(
            &fd,
            Setter::<TUNSETIFF, _>::new(*b"example0\0\0\0\0\0\0\0\0\x01\x10"),
        )?;
        println!("tunsetiff successful");
        // ioctl(&fd, Setter::<TUNSETOFFLOAD, _>::new(TUN_F_CSUM))?; // will throw "InvalidInput" here!
        ioctl(&fd, UintSetter::<TUNSETOFFLOAD>::new(TUN_F_CSUM))?; // this works!
        println!("tunsetoffload successful");
    }
    Ok(())
}

// we need to unsafe impl a non-standard Ioctl to make that work :/

struct UintSetter<Oc> {
    input: c_uint,
    _opcode: PhantomData<Oc>,
}

impl<Oc> UintSetter<Oc> {
    pub fn new(input: c_uint) -> Self {
        Self {
            input,
            _opcode: PhantomData,
        }
    }
}

unsafe impl<Oc: CompileTimeOpcode> Ioctl for UintSetter<Oc> {
    type Output = ();
    const IS_MUTATING: bool = true;
[...]

PR #1172, opened 2024-09-22
Add Ioctl helper for integer arg

notgull:
Fixing low-hanging-fruit issue #1051.


sunfishcode:
Would it be feasible to add a test that uses this new helper?

notgull:
Added!

notgull:
@sunfishcode poke

PR #1173, opened 2024-09-22
Fix new warnings on nightly

notgull:
static mut is now discouraged; however it appears that the places we use static mut can be removed pretty easily.

### DIFF
--- a/src/backend/linux_raw/vdso_wrappers.rs
+++ b/src/backend/linux_raw/vdso_wrappers.rs
@@ -317,7 +317,7 @@ fn init_syscall() -> SyscallType {
 /// placeholder type, and cast it as needed.
 struct Function;
 #[cfg(feature = "time")]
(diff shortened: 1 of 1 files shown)


sunfishcode:
Thanks! Could you also remove the "SAFETY:" labels in the comments that no longer precede `unsafe` blocks?

notgull:
> Could you also remove the "SAFETY:" labels in the comments that no longer precede `unsafe` blocks?

These still involve loading pointers to raw functions/weird memory regions, so I figured they should remain.

sunfishcode:
The contents of the comments can stay. But if Rust doesn't consider the code `unsafe` then these aren't "SAFETY" comments per se.

notgull:
> The contents of the comments can stay. But if Rust doesn't consider the code `unsafe` then these aren't "SAFETY" comments per se.

Ok, this has been updated.

sunfishcode:
Thanks!

PR #1174, opened 2024-09-23
Expose socket option SO_RCVBUFFORCE

yshui:
This option allows a privileged user to override the receive buffer size limit.


sunfishcode:
Thanks!

PR #1175, opened 2024-09-25
Don't enable the "fs" module implicitly.

sunfishcode:
Several configurations used to enable the "fs" module as rustix used to use the "procfs" module for reading the AUXV vector on Linux. Rustix no longer does this, as the "procfs" module's extra checks are incompatible with QEMU on one hand, and not needed by most use cases on the other.

So since those configurations no longer use the "fs" module, we can remove the code that enabled it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,7 +156,7 @@ shm = ["fs"]
 time = []
 
 # Enable `rustix::param::*`.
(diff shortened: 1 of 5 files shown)


sunfishcode:
Ah, never mind. This isn't as simple as I thought.

PR #1179, opened 2024-10-04
s390x linux_raw support

taiki-e:
This is currently implemented under rustix_use_experimental_asm since s390x asm is unstable. However, we might want to wait for stabilization (https://github.com/rust-lang/rust/pull/131258) and implement as non-experimental.

Tested downstream crates (tested with qemu-user 9.1.0): async-io, polling, tempfile, cap-std

FYI @uweigand


sunfishcode:
Looks good, thanks!

PR #1184, opened 2024-10-16
Remove potential panic in `split_init`

newpavlov:
The function already implicitly requires that `init` is smaller or equal to `buf`'s length.


sunfishcode:
Thanks!

Issue #1193, opened 2024-10-20
`error[E0433]: failed to resolve: could not find 'mount' in 'backend'`


polarathene:
**TL;DR:** I think this was meant to be removed from `lib.rs` with [this Sep 2024 PR](https://github.com/bytecodealliance/rustix/pull/1154), but was missed?

https://github.com/bytecodealliance/rustix/blob/ec96121bf66b065452bf2d784035400df558aabd/src/lib.rs#L296-L300

---

I wanted to try the recently implemented `select` method, and realized it's not available in a release yet, so I switched to git for `rustix`:

```toml
rustix = { git = "https://github.com/bytecodealliance/rustix.git", features = ["procfs", "event"] }
```

Compiling fails however due to various errors about the backend lacking the `mount` module (_All errors are from `src/mount/` files referring to the backend mount module_):

```rust
error[E0433]: failed to resolve: could not find `mount` in `backend`
  --> /root/.cargo/git/checkouts/rustix-8fafe793515fde97/ec96121/src/mount/fsopen.rs:3:21
   |
3  | use crate::backend::mount::types::{
   |                     ^^^^^ could not find `mount` in `backend`
   |
note: found an item that was configured out
  --> /root/.cargo/git/checkouts/rustix-8fafe793515fde97/ec96121/src/backend/linux_raw/mod.rs:49:16
   |
49 | pub(crate) mod mount;
   |                ^^^^^
note: the item is gated behind the `mount` feature
  --> /root/.cargo/git/checkouts/rustix-8fafe793515fde97/ec96121/src/backend/linux_raw/mod.rs:48:7
   |
48 | #[cfg(feature = "mount")]
   |       ^^^^^^^^^^^^^^^^^
```

The `fs` feature is implicitly enabled from `procfs` feature (_so `lib.rs` then enables the `mount` module_):

https://github.com/bytecodealliance/rustix/blob/ec96121bf66b065452bf2d784035400df558aabd/Cargo.toml#L161-L162

## Context

- Earlier PR: https://github.com/bytecodealliance/rustix/pull/763
- Recent Sep 2024 PR (_[removed for backends](https://github.com/bytecodealliance/rustix/pull/1154/files#diff-19da5b08da9ce4980b74a43f8ad55fe785de57fa3c9e0117db639962e59fb27a), but not `lib.rs`_): https://github.com/bytecodealliance/rustix/pull/1154

[...]

PR #1194, opened 2024-10-20
chore: Remove `fs` feature enabling `mod mount`

polarathene:
Seems to have been missed as part of: https://github.com/bytecodealliance/rustix/pull/1154
Fixes: https://github.com/bytecodealliance/rustix/issues/1193


sunfishcode:
Thanks!

Issue #1183, opened 2024-10-15
Feature request: support for `sched_setscheduler`


eglimi:
There is currently no support for getting or setting the scheduling policies with `sched_getscheduler` and `sched_setscheduler` respectively. I use this on Linux with PREEMPT_RT to set realtime priorities and FIFO scheduling, currently with `libc`.

Other related parts are also not in rustix, e.g. to set the stack size (pthread).

Would you consider this in scope of rustix?

sunfishcode:
I'm not very familiar with `pthread_setschedparam`, but at first glance, it looks like it's more than just a trivial wrapper around `sched_setscheduler`, and it does things that only libc can do. I am familiar with `pthread_attr_setstacksize`, and I know that it's doing things that only libc can do. So it seems the best thing for users of these functions would be to continue to use libc.

If a safe interface to these functions is desired, it'd ideally go in a separate library from rustix, as it would just wrap libc and wouldn't require access to rustix's internals.

eglimi:
As a user, the ergonomics of rustix is very nice. Since libc is listed as a backend, I thought these APIs could be considered. What you write make sense though.

I leave this issue open in case anybody else wants to comment and will close it after a couple of days.
Thank you for your input and great work!

PR #1195, opened 2024-10-23
Add `sched_getscheduler` & `sched_setscheduler`

ruihe774:
Solve #1183

### DIFF
--- a/src/backend/libc/process/syscalls.rs
+++ b/src/backend/libc/process/syscalls.rs
@@ -2,6 +2,8 @@
 
 #[cfg(any(freebsdlike, linux_kernel, target_os = "fuchsia"))]
 use super::types::RawCpuSet;
+#[cfg(any(freebsdlike, linux_kernel))]
+use super::types::{SchedParam, SchedPolicy};
 use crate::backend::c;
 #[cfg(not(any(target_os = "wasi", target_os = "fuchsia")))]
 use crate::backend::conv::borrowed_fd;
@@ -209,6 +211,29 @@ pub(crate) fn sched_setaffinity(pid: Option<Pid>, cpuset: &RawCpuSet) -> io::Res
             Pid::as_raw(pid) as _,
             core::mem::size_of::<RawCpuSet>(),
             cpuset,
+        ))
+    }
+}
+
+#[cfg(any(freebsdlike, linux_kernel))]
+#[inline]
+pub(crate) fn sched_getscheduler(pid: Option<Pid>) -> io::Result<SchedPolicy> {
+    unsafe { ret_c_int(c::sched_getscheduler(Pid::as_raw(pid) as _)) }
+        .map(|r| SchedPolicy::from_bits_retain(r as _))
+}
+
+#[cfg(any(freebsdlike, linux_kernel))]
+#[inline]
+pub(crate) fn sched_setscheduler(
+    pid: Option<Pid>,
+    policy: SchedPolicy,
+    param: &SchedParam,
+) -> io::Result<()> {
+    unsafe {
+        ret(c::sched_setscheduler(
+            Pid::as_raw(pid) as _,
+            policy.bits() as _,
+            param as *const SchedParam as _,
         ))
     }
 }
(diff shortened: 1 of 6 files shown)


sunfishcode (commented):
Could you say more about the use case that motivates this PR? In https://github.com/bytecodealliance/rustix/issues/1183, the reporter actually needed `pthread_setschedparam` etc., which has more consistent behavior across platforms, which isn't [entirely unimplmented on musl](https://git.musl-libc.org/cgit/musl/tree/src/sched/sched_setscheduler.c), and which cooperates with the platform libpthread rather than working behind its back, so it's better for most users.

Rustix may still want to add `sched_setscheduler`, but I'd like to understand how it would be used so that we can write a documentation comment pointing people to `pthread_setschedparam` in situations where that would be better.

ruihe774:
> Could you say more about the use case that motivates this PR? In #1183, the reporter actually needed `pthread_setschedparam` etc., which has more consistent behavior across platforms, which isn't [entirely unimplmented on musl](https://git.musl-libc.org/cgit/musl/tree/src/sched/sched_setscheduler.c), and which cooperates with the platform libpthread rather than working behind its back, so it's better for most users.
> 
> Rustix may still want to add `sched_setscheduler`, but I'd like to understand how it would be used so that we can write a documentation comment pointing people to `pthread_setschedparam` in situations where that would be better.

I don't think so. `sched_setscheduler` is also in POSIX standard. And `pthread_setschedparam` does not have more consistent behavior across platforms; its arguments `policy` and `sched_param` are the same as `sched_setscheduler` and still vary across platforms. I don't think there is much difference between these two APIs.

And using `pthread_setschedparam` means we cannot implement it in `linux_raw` and requires the libc.

In Linux, scheduling is not managed by libpthread; `pthread_setschedparam` is just a slim wrapper to `sched_setscheduler`. So there is no such thing that "cooperates with the platform libpthread rather than working behind its back". For other platforms, we use libc anyway.

sunfishcode:
> > Could you say more about the use case that motivates this PR? In #1183, the reporter actually needed `pthread_setschedparam` etc., which has more consistent behavior across platforms, which isn't [entirely unimplmented on musl](https://git.musl-libc.org/cgit/musl/tree/src/sched/sched_setscheduler.c), and which cooperates with the platform libpthread rather than working behind its back, so it's better for most users.
> > Rustix may still want to add `sched_setscheduler`, but I'd like to understand how it would be used so that we can write a documentation comment pointing people to `pthread_setschedparam` in situations where that would be better.
> 
> I don't think so. `sched_setscheduler` is also in POSIX standard. And `pthread_setschedparam` does not have more consistent behavior across platforms; its arguments `policy` and `sched_param` are the same as `sched_setscheduler` and still vary across platforms. I don't think there is much difference between these two APIs.

`sched_setscheduler` isn't portable between glibc and musl.

It appears that all of linux/glibc, linux/musl, FreeBSD, and POSIX all support `pthread_setschedparam`.

> And using `pthread_setschedparam` means we cannot implement it in `linux_raw` and requires the libc.

Could you say more about your use case?

> In Linux, scheduling is not managed by libpthread; `pthread_setschedparam` is just a slim wrapper to `sched_setscheduler`. So there is no such thing that "cooperates with the platform libpthread rather than working behind its back". For other platforms, we use libc anyway.

[glibc's wrapper](https://codebrowser.dev/glibc/glibc/nptl/pthread_setschedparam.c.html) does do some additional things, including update some metadata that would become stale if the user used `sched_setschedparam` directly. I don't know how hazardous that is, but it looks subtle enough that it'd be good to give users guidance.

ruihe774:
> `sched_setscheduler` isn't portable between glibc and musl.

We can implement it using raw syscall if musl does not provide it.

> Could you say more about your use case?

I consider not "requires the libc" an advantage.

> I don't know how hazardous that is, but it looks subtle enough that it'd be good to give users guidance.

Right.

I can implement both and point that in doc.

ruihe774:
> I can implement both

The problem is that `pthread_setschedparam` and `pthread_getschedparam` take `pthread_t`, which is not available in rustix.

sunfishcode:
Without knowing anything about your use case, I'm not able to evaluate this.

PR #1198, opened 2024-10-28
Add a `rustix::fs::STATX_ATTR_MOUNT_ROOT` constant.

JustSoup312:
Add a `STATX_ATTR_MOUNT_ROOT` constant for Linux with musl libc compatibility.

(Note: This is my first contribution to rustix, so I assume that I may have placed const declarations in the wrong place. I will gladly fix any issues with this MR. Thank you!)

### DIFF
--- a/src/backend/libc/c.rs
+++ b/src/backend/libc/c.rs
@@ -14,6 +14,12 @@ pub(crate) const PROC_SUPER_MAGIC: u32 = 0x0000_9fa0;
 #[cfg(all(linux_kernel, target_env = "musl"))]
 pub(crate) const NFS_SUPER_MAGIC: u32 = 0x0000_6969;
 
+/// `STATX_ATTR_MOUNT_ROOT`—Address of the root of a mount
+#[cfg(all(linux_kernel, target_env = "musl"))]
+pub(crate) const STATX_ATTR_MOUNT_ROOT: c_int = 0x2000;
+#[cfg(all(linux_kernel, target_env = "gnu"))]
+pub(crate) const STATX_ATTR_MOUNT_ROOT: c_int = libc::STATX_ATTR_MOUNT_ROOT;
+
 #[cfg(feature = "process")]
 #[cfg(not(any(target_os = "espidf", target_os = "wasi")))]
 pub(crate) const EXIT_SIGNALED_SIGABRT: c_int = 128 + SIGABRT as c_int;
--- a/src/backend/linux_raw/c.rs
+++ b/src/backend/linux_raw/c.rs
@@ -37,6 +37,12 @@ pub(crate) use linux_raw_sys::general::epoll_event;
         )
     )
 ))]
+#[cfg(any(target_env = "musl", target_env = "gnu"))]
+pub(crate) use linux_raw_sys::general::{
+    AT_FDCWD, NFS_SUPER_MAGIC, O_LARGEFILE, PROC_SUPER_MAGIC, STATX_ATTR_MOUNT_ROOT, UTIME_NOW,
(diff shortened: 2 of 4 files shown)


src/backend/libc/c.rs
```diff
@@ -14,6 +14,12 @@ pub(crate) const PROC_SUPER_MAGIC: u32 = 0x0000_9fa0;
 #[cfg(all(linux_kernel, target_env = "musl"))]
 pub(crate) const NFS_SUPER_MAGIC: u32 = 0x0000_6969;
 
+/// `STATX_ATTR_MOUNT_ROOT`—Address of the root of a mount
+#[cfg(all(linux_kernel, target_env = "musl"))]
+pub(crate) const STATX_ATTR_MOUNT_ROOT: c_int = 0x2000;
+#[cfg(all(linux_kernel, target_env = "gnu"))]
+pub(crate) const STATX_ATTR_MOUNT_ROOT: c_int = libc::STATX_ATTR_MOUNT_ROOT;
```
sunfishcode, line 21: This module does `use libc::*` at the top, so this line is unnecessary.

src/backend/linux_raw/c.rs
```diff
@@ -37,6 +37,12 @@ pub(crate) use linux_raw_sys::general::epoll_event;
         )
     )
 ))]
+#[cfg(any(target_env = "musl", target_env = "gnu"))]
+pub(crate) use linux_raw_sys::general::{
+    AT_FDCWD, NFS_SUPER_MAGIC, O_LARGEFILE, PROC_SUPER_MAGIC, STATX_ATTR_MOUNT_ROOT, UTIME_NOW,
+    UTIME_OMIT, XATTR_CREATE, XATTR_REPLACE,
+};
+#[cfg(not(any(target_env = "musl", target_env = "gnu")))]
```
sunfishcode, line 45: The src/backend/linux_raw directory doesn't use libc bindings for most things, so we should be able to just add `STATX_ATTR_MOUNT_ROOT` to the `use` below instead of adding a new `use` here.

src/fs/statx.rs
```diff
@@ -10,6 +10,11 @@ use backend::fs::syscalls::statx as _statx;
 #[cfg(not(feature = "linux_4_11"))]
 use compat::statx as _statx;
 
+/// `STATX_ATTR_MOUNT_ROOT`—Address of the root of a mount
+#[allow(dead_code)]
```
sunfishcode, line 14: Since this is a `pub const` and re-exported in mod.rs, it shouldn't need `allow(dead_code)`.

src/fs/mod.rs
```diff
@@ -116,6 +116,8 @@ pub use sendfile::sendfile;
 pub use special::*;
 #[cfg(linux_kernel)]
 pub use statx::statx;
+#[cfg(all(linux_kernel, any(target_env = "musl", target_env = "gnu")))]
+pub use statx::STATX_ATTR_MOUNT_ROOT;
```
sunfishcode, line 120: One thing we can do to make this code a little tidier is just change the `use statx::statx` above to be `use statx::*;`, instead of adding a new `use`, so that we don't have to repeat all the `cfg`s.

sunfishcode (commented):
Thanks! Overall it looks good; a few comments below:

rusty-snake:
This can be closed because of 7202da6e9, right?

sunfishcode:
Yes, https://github.com/bytecodealliance/rustix/pull/1282 intoduced a `StatxAttributes` flag type, which includes the `MOUNT_ROOT` flag. Thanks!

PR #1200, opened 2024-10-28
Don't store build check rmeta

### LINKED ISSUE #1199: rust metadata generated from build script creates determinism issues
The `build.rs` has a function [`can_compile`](https://github.com/bytecodealliance/rustix/blob/7d4a719f33495c45ba74a3a8fb7a16bf737f8036/build.rs#L205) that checks whether the provided `rustc` can compile some given code:

```rs
/// Test whether the rustc at `var("RUSTC")` can compile the given code.
fn can_compile<T: AsRef<str>>(test: T) -> bool {
    ...
    cmd.arg("--crate-type=rlib") // Don't require `main`.
        .arg("--emit=metadata") // Do as little as possible but still parse.
        .arg("--target")
        .arg(target)
        .arg("--out-dir")
        .arg(out_dir); // Put the output somewhere inconsequential.
    ...
    cmd.spawn().unwrap().wait().unwrap().success()
}
```

While `out_dir` might be inconsequential for most builds, it creates determinism issues (and cache misses) in some build systems like [Bazel](https://github.com/bazelbuild/rules_rust) where a changing `.rmeta` file might invalidate the cache and create spurious rebuilds.

nmattia:
Fixes #1199 

This ensures the `.rmeta` files generated by `can_compile` are not written to the build output. Those metadata files can create determinism and cache invalidation issues in some build systems.

Instead the output -- since it is never actually read -- is written to `/dev/null`.


nmattia:
Note: I'm not sure whether `/dev/null` will work on "exotic" platforms like Windows!

nmattia:
Note 2: as far as I can tell the metadata is not being read anymore since https://github.com/bytecodealliance/rustix/pull/730

sunfishcode:
Thanks for the PR! This looks reasonable to me, though I also don't know how this works on Windows. That said, the Windows CI appears to have passed. Is Windows is recognizing /dev/null for compatibility?

nmattia:
> Is Windows is recognizing /dev/null for compatibility?

If the CI runs under WSL then most likely yes. Also I see the build script uses various sorts of pipes for stdin, so I guess using `/dev/null` should be OK?

sunfishcode:
The Windows CI doesn't run in WSL. Windows has its own version of pipes and Rust's `std::process::Command` knows how to use it.

What if we instead used `.arg("-o").arg("-")` and `.stdout(Stdio::null())`? That seems like it should work and avoid any portability question.

nmattia:
> What if we instead used .arg("-o").arg("-") and .stdout(Stdio::null())? That seems like it should work and avoid any portability question.

That sounds good! I'll push a fix tomorrow when I have some time. Thanks for the quick feedback!

nmattia:
@sunfishcode PTAL!

sunfishcode:
Thanks!

sunfishcode:
This is now released in rustix 0.38.39.

PR #1203, opened 2024-10-30
Enable a few features on more platforms.

sunfishcode:
Define `PollFlags::RDHUP` on FreeBSD, `AT_EACCESS` on Emscripten, and `ptsname` on Illumos.

### DIFF
--- a/src/backend/libc/event/mod.rs
+++ b/src/backend/libc/event/mod.rs
@@ -5,5 +5,5 @@ pub(crate) mod types;
 #[cfg_attr(windows, path = "windows_syscalls.rs")]
 pub(crate) mod syscalls;
 
-#[cfg(any(linux_kernel, target_os = "redox"))]
+#[cfg(any(linux_kernel, solarish, target_os = "redox"))]
 pub mod epoll;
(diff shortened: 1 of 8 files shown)


src/backend/libc/event/mod.rs
```diff
@@ -5,5 +5,5 @@ pub(crate) mod types;
 #[cfg_attr(windows, path = "windows_syscalls.rs")]
 pub(crate) mod syscalls;
 
-#[cfg(any(linux_kernel, target_os = "redox"))]
+#[cfg(any(linux_kernel, solarish, target_os = "redox"))]
```
taiki-e, line 8: epoll is supported in illumos, but is it also supported in solaris? (I could not find the man page.)
https://illumos.org/man/7/epoll

Also, libc defines epoll-related stuff for solarish only in illumos.rs, so doesn't this actually cause a compile error in solaris?
https://github.com/search?q=repo%3Arust-lang%2Flibc+%2Fepoll_%2F+path%3Asrc%2Funix%2Fsolarish&type=code

taiki-e: Thanks for the link and PR. The comment in in src/unix/solarish/mod.rs removed in the libc PR says:
```rs
    // The epoll functions are actually only present on illumos.  However,
    // there are things using epoll on illumos (built using the
    // x86_64-pc-solaris target) which would break until the illumos target is
    // present in rustc.
```

It appears to be what was used before illumos was added as its own target.

PR #1205, opened 2024-11-04
fix: enable wasip2 feature for wasm32-wasip2 target

cdmurph32:
* Updating MSRV to 1.82 as that the first version with the wasip2 target
* wasip2 will require +nightly until https://github.com/rust-lang/rust/issues/130323 is resolved and/or `std::os::wasip2` is available in stable. (please correct me if I'm wrong about that)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2021"
 keywords = ["api", "file", "network", "safe", "syscall"]
 categories = ["os::unix-apis", "date-and-time", "filesystem", "network-programming"]
 include = ["src", "build.rs", "Cargo.toml", "COPYRIGHT", "LICENSE*", "/*.md", "benches"]
-rust-version = "1.63"
+rust-version = "1.82"
 
 [dependencies]
 bitflags = { version = "2.4.0", default-features = false }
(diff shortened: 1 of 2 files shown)


sunfishcode (commented):
Thanks! Just one comment:

Cargo.toml
```diff
@@ -13,7 +13,7 @@ edition = "2021"
 keywords = ["api", "file", "network", "safe", "syscall"]
 categories = ["os::unix-apis", "date-and-time", "filesystem", "network-programming"]
 include = ["src", "build.rs", "Cargo.toml", "COPYRIGHT", "LICENSE*", "/*.md", "benches"]
-rust-version = "1.63"
```
sunfishcode, line 16: Could you revert this `rust-version` change? It isn't practical for us to bump the MSRV for all rustix users. Since wasip2 is a new target, I think we can just leave rust-version at 1.63 and just expect people using wasip2 to use the lastest Rust version.

cdmurph32: Sure.

sunfishcode:
Thanks!

sunfishcode:
This is now released in rustix 0.38.39.

Issue #1204, opened 2024-11-04
Please document when `getppid` can return `None`


joshtriplett:
`getppid` returns `Option<Pid>`, but the `getppid` syscall can never fail. I'm *guessing* that `None` here occurs when `getppid` returns 0 because the current process is init and the parent process is the kernel's PID 0, and `Pid` can't represent PID 0?

PR #1209, opened 2024-11-08
Clarify `getppid` documentation: explain how `None` can occur

joshtriplett:
The `getppid` syscall can never fail, but rustix returns PID 0 as
`None`.

closes: #1204

### DIFF
--- a/src/process/id.rs
+++ b/src/process/id.rs
@@ -113,6 +113,9 @@ pub fn getpid() -> Pid {
 
 /// `getppid()`—Returns the parent process' ID.
 ///
+/// This will return `None` if the current process has no parent (or no parent accessible in the
+/// current PID namespace), such as if the current process is an init process (PID 1).
+///
 /// # References
 ///  - [POSIX]
 ///  - [Linux]


src/process/id.rs
```diff
@@ -113,6 +113,9 @@ pub fn getpid() -> Pid {
 
 /// `getppid()`—Returns the parent process' ID.
 ///
+/// This will return `None` if the parent process ID is 0; this will happen, for instance, for an
+/// init process (PID 1).
```
sunfishcode, line 117: According to [the docs](https://man7.org/linux/man-pages/man2/getpid.2.html#NOTES), `getppid` returns 0 if the parent is in a different pid namespace.

And as a minor nit, no process ever has ID 0; 0 is a sentinel value rather than an ID value.

What would you think of wording this as "This will return `None` if the the process is the init process (PID 1) or if the parent is in a different PID namespace"?

joshtriplett: My goal with saying "for instance" here was to avoid attempting to *exhaustively* enumerate the circumstances that could lead to this, not least of which because those circumstances are OS-specific. However, I can clarify this a bit further.

sunfishcode:
Looks great, thanks!

PR #1210, opened 2024-11-08
Migrate from wasm32-wasi (no longer exists on nightly) to wasm32-wasip1

joshtriplett:
Current CI is broken because wasm32-wasi no longer exists.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -79,9 +79,11 @@ jobs:
         sparcv9-sun-solaris
         aarch64-linux-android
         aarch64-apple-ios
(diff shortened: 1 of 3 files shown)


joshtriplett:
Looks like the remaining failure is coming from unrelated Solaris reasons.

sunfishcode:
The Solaris issues appear to be due to a recent libc 0.2.x release removing a bunch of types and constants on Solaris. I've now submitted https://github.com/bytecodealliance/rustix/pull/1211 to update rustix accordingly.

sunfishcode:
Thanks for fixing this!

PR #1212, opened 2024-11-09
Use/fix `ret_*` helpers for certain libc functions

pinotree:
- use `ret_usize()` instead of `ret_send_recv()` for `c::recvmsg()`, as it always returns `c::ssize_t`
- use `ret_usize()` instead of `ret_send_recv()` for `c::sendmsg()`, as it always returns `c::ssize_t`
- change the `ret_send_recv()` used on OSes different than Windows, Redox, and wasi to take `c::ssize_t` as parameter, as it is what `c::recv()`, `c::send()`, `c::recvfrom()`, and `c::sendto()` return on most/all of those OSes

### DIFF
--- a/src/backend/libc/conv.rs
+++ b/src/backend/libc/conv.rs
@@ -170,7 +170,7 @@ pub(super) fn send_recv_len(len: usize) -> i32 {
 /// Convert the return value of a `send` or `recv` call.
 #[cfg(not(any(windows, target_os = "redox", target_os = "wasi")))]
 #[inline]
-pub(super) fn ret_send_recv(len: isize) -> io::Result<usize> {
+pub(super) fn ret_send_recv(len: c::ssize_t) -> io::Result<usize> {
     ret_usize(len)
 }
 
--- a/src/backend/libc/net/syscalls.rs
+++ b/src/backend/libc/net/syscalls.rs
@@ -7,7 +7,9 @@ use super::msghdr::with_xdp_msghdr;
 #[cfg(target_os = "linux")]
 use super::write_sockaddr::encode_sockaddr_xdp;
 use crate::backend::c;
-use crate::backend::conv::{borrowed_fd, ret, ret_owned_fd, ret_send_recv, send_recv_len};
+use crate::backend::conv::{
+    borrowed_fd, ret, ret_owned_fd, ret_send_recv, ret_usize, send_recv_len,
+};
 use crate::fd::{BorrowedFd, OwnedFd};
 use crate::io;
 #[cfg(target_os = "linux")]
@@ -329,7 +331,7 @@ pub(crate) fn recvmsg(
 
     with_recv_msghdr(&mut storage, iov, control, |msghdr| {
         let result = unsafe {
-            ret_send_recv(c::recvmsg(
+            ret_usize(c::recvmsg(
                 borrowed_fd(sockfd),
                 msghdr,
                 bitflags_bits!(msg_flags),
@@ -364,7 +366,7 @@ pub(crate) fn sendmsg(
     msg_flags: SendFlags,
 ) -> io::Result<usize> {
     with_noaddr_msghdr(iov, control, |msghdr| unsafe {
-        ret_send_recv(c::sendmsg(
+        ret_usize(c::sendmsg(
             borrowed_fd(sockfd),
             &msghdr,
             bitflags_bits!(msg_flags),
@@ -387,7 +389,7 @@ pub(crate) fn sendmsg_v4(
     msg_flags: SendFlags,
 ) -> io::Result<usize> {
     with_v4_msghdr(addr, iov, control, |msghdr| unsafe {
-        ret_send_recv(c::sendmsg(
+        ret_usize(c::sendmsg(
             borrowed_fd(sockfd),
             &msghdr,
             bitflags_bits!(msg_flags),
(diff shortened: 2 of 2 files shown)


sunfishcode:
> * use `ret_usize()` instead of `ret_send_recv()` for `c::recvmsg()`, as it always returns `c::ssize_t`
> 
>     * use `ret_usize()` instead of `ret_send_recv()` for `c::sendmsg()`, as it always returns `c::ssize_t`

Windows doesn't have `sendmsg`/`recvmsg`, however `send`/`recv` return `i32` instead of `ssize_t`. `WSASendMsg` isn't directly analogous to POSIX `sendmsg`, but it returns its number of bytes as an `i32` as well. So it seems that `ret_send_recv` is the right thing to use for these.

>     * change the `ret_send_recv()` used on OSes different than Windows, Redox, and wasi to take `c::ssize_t` as parameter, as it is what `c::recv()`, `c::send()`, `c::recvfrom()`, and `c::sendto()` return on most/all of those OSes

This part sounds good.

sunfishcode:
> > Windows doesn't have `sendmsg`/`recvmsg`, however `send`/`recv` return `i32` instead of `ssize_t`. `WSASendMsg` isn't directly analogous to POSIX `sendmsg`, but it returns its number of bytes as an `i32` as well. So it seems that `ret_send_recv` is the right thing to use for these.
> 
> Right, there is the Windows version of `ret_send_recv()` already that does that:
> 
> https://github.com/bytecodealliance/rustix/blob/d51b19556eb637c8cf6a49effce439868e3aed16/src/backend/libc/conv.rs#L177-L182
> 
> Did I miss anything? 🤔

Because of that, `ret_send_recv` is already the right function to use for converting the return values of `sendmsg` and `recvmsg`.

pinotree:
> Because of that, `ret_send_recv` is already the right function to use for converting the return values of `sendmsg` and `recvmsg`.

Not really:
- `sendmsg()` and `recvmsg()` return `ssize_t` in their POSIX standard definitions [1][2]
- all the implementations of `sendmsg()` and `recvmsg()` in the libc crate return `ssize_t`
- in the libc create, the Hurd implementation does not have `ssize_t == isize` but `ssize_t == c_int`

While the different type of `libc::ssize_t` on Hurd might be an issue, I think that rustix ought to use the right type of `sendmsg()` and `recvmsg()` in libc which is `libc::ssize_t`, without relying on type equivalence.

[1] https://pubs.opengroup.org/onlinepubs/9699919799/functions/sendmsg.html
[2] https://pubs.opengroup.org/onlinepubs/9699919799/functions/recvmsg.html

sunfishcode:
[Windows doesn't follow POSIX](https://docs.rs/windows-sys/latest/windows_sys/Win32/Networking/WinSock/fn.send.html). It uses `i32`. `ret_send_recv` can take an `ssize_t` on POSIX platforms, and it takes an `i32` on Windows, so it's the right function to use.

pinotree:
> `ret_send_recv` can take an `ssize_t` on POSIX platforms,

As I said earlier, this only works when `ssize_t` is defined in the libc crate as `isize`, which happens on almost all the unix-ish platforms. On 32bit Hurd, `ssize_t` is defined as `i32`, which is not equivalent to `isize`, and thus building something that uses `rustix` fails:
```
error[E0308]: mismatched types
   --> /usr/share/cargo/registry/rustix-0.38.37/src/backend/libc/net/syscalls.rs:367:23
    |
367 |           ret_send_recv(c::sendmsg(
    |  _________-------------_^
    | |         |
    | |         arguments to this function are incorrect
368 | |             borrowed_fd(sockfd),
369 | |             &msghdr,
370 | |             bitflags_bits!(msg_flags),
371 | |         ))
    | |_________^ expected `isize`, found `i32`
```
Hence this PR:
- since `sendmsg()` and `recvmsg()` are not implemented on Windows, and `sendmsg()` and `recvmsg()` in the libc rustix backend are not implemented on Windows, their usage is unconditionally switched to `ret_usize()`, which (despite the name) takes `c::ssize_t`
- regarding the other send/recv functions: since `ret_send_recv()` has already a version for Windows (which is _not_ changed by this PR), the non-Windows version is switched to use `c::ssize_t` instead

So there should be no changes for Windows, and indeed the CI jobs for Windows pass.

sunfishcode:
Also, your branch doesn't build on hurd either:

```shell
    Checking rustix v0.38.35 (...)
error[E0308]: mismatched types
   --> src/backend/libc/net/syscalls.rs:369:19
    |
369 |           ret_usize(c::sendmsg(
    |  _________---------_^
    | |         |
    | |         arguments to this function are incorrect
370 | |             borrowed_fd(sockfd),
371 | |             &msghdr,
372 | |             bitflags_bits!(msg_flags),
373 | |         ))
    | |_________^ expected `isize`, found `i32`
    |
note: function defined here
   --> src/backend/libc/conv.rs:88:15
    |
88  | pub(super) fn ret_usize(raw: c::ssize_t) -> io::Result<usize> {
    |               ^^^^^^^^^ ---------------
help: you can convert an `i32` to an `isize` and panic if the converted value doesn't fit
    |
373 |         ).try_into().unwrap())
    |          ++++++++++++++++++++

...
```

sunfishcode:
It looks like https://github.com/rust-lang/libc/pull/4029 will change hurd's `ssize_t` to be an alias for `isize`.

sunfishcode:
So in summary here,
 - `sendmsg` and `recvmsg` should continue to use `ret_send_recv`.
 - `ret_send_recv` could take an `ssize_t` instead of an `isize`.

sunfishcode:
Closing, as the main change here isn't needed. If anyone is interested in support for platforms where `ssize_t` differs from `isize`, please submit new PRs.

PR #1215, opened 2024-11-13
Fix `port_getn` calling.

Berrysoft:
The `max` passed to `port_getn` should be the capacity of the buffer, not the length.

If `max` is zero, `nget` is the events avaliable in the port. Then the calling to `set_len` will be unsound.

Ref: https://illumos.org/man/3C/port_getn


sunfishcode:
Thanks!

Berrysoft:
By the way, would you publish a new release for this fix?

sunfishcode:
This is now fixed in rustix 0.38.41.

PR #1216, opened 2024-11-13
Add a `port_getn_query` function.

sunfishcode:
The solarish [`port_getn` function] special-cases a `max` argument value of 0 to be a query of the number of events available. #1215 added a special-case to protect the code from doing a resize in that case. And in case users actually do want to do a query, this PR adds a new `port_getn_query` function that passes a zero.

[`port_getn` function]: https://illumos.org/man/3C/port_getn


Berrysoft (approved):
I'm expecting a new publish for the fixes to `port_getn`.

PR #1226, opened 2024-11-28
fcntl_lock should be supported on Solaris

psumbera:
Even when Solaris libc doesn't define LOCK_SH, LOCK_EX, LOCK_NB and LOCK_UN (which are flock() related).

### DIFF
--- a/src/backend/libc/fs/types.rs
+++ b/src/backend/libc/fs/types.rs
@@ -949,27 +949,49 @@ bitflags! {
 ///
 /// [`flock`]: crate::fs::flock
 /// [`fcntl_lock`]: crate::fs::fcntl_lock
-#[cfg(not(any(
-    target_os = "espidf",
-    target_os = "solaris",
-    target_os = "vita",
-    target_os = "wasi"
-)))]
+// Solaris doesn't support `flock` and doesn't define `LOCK_SH` etc., but we
+// reuse this `FlockOperation` enum for `fcntl_lock`, so on Solaris we use
+// our own made-up integer values.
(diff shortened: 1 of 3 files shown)


src/backend/libc/fs/types.rs
```diff
@@ -949,27 +949,46 @@ bitflags! {
 ///
 /// [`flock`]: crate::fs::flock
 /// [`fcntl_lock`]: crate::fs::fcntl_lock
-#[cfg(not(any(
-    target_os = "espidf",
-    target_os = "solaris",
-    target_os = "vita",
-    target_os = "wasi"
-)))]
+#[cfg(not(any(target_os = "espidf", target_os = "vita", target_os = "wasi")))]
```
sunfishcode, line 952:
```suggestion
// Solaris doesn't support `flock` and doesn't define `LOCK_SH` etc., but we
// reuse this `FlockOperation` enum for `fcntl_lock`, so on Solaris we use
// our own made-up integer values.
#[cfg(not(any(target_os = "espidf", target_os = "vita", target_os = "wasi")))]
```

sunfishcode (commented):
Ah, this looks like my mistake. When libc removed `LOCK_SH` etc., I disabled `FlockOperation` on Solaris and then disabled `fcntl_lock` as well to make everything build, forgetting that we still need that on Solaris. The fix looks good; I just suggest adding something like the following comment to explain what's going on:

sunfishcode:
The CI failure is an unrelated bug now fixed on `main`.

sunfishcode:
Thanks!

sunfishcode:

PR #1229, opened 2024-12-01
bump windows-sys dependency to 0.59.0

vthib:
The windows-sys crate has been bumped to 0.59.0 4 months ago. In my projcets, most of the crates using it has been updated but rustix, hence this PR to align it.

I followed the same changes as were done in https://github.com/bytecodealliance/rustix/pull/935 , that is bumping the errno version at the same time to make it use this new windows-sys version as well.

I didn't know whether to push this to main or 0.38, i figured main is probably the right one and a backport could be done, but tell me if you need me to target another branch.

Thanks!


sunfishcode:
Could you change this to allow either 0.52 or 0.59, [as the errno crate does](https://github.com/lambda-fairy/rust-errno/blob/main/Cargo.toml#L19)? That way it won't introduce duplicate window-sys dependencies for existing users that are using 0.52.

And yes, main is good. I'll backport it to 0.38 after it's landed.

vthib:
Yes good idea, I updated the requirements

sunfishcode:
Thanks!

sunfishcode:

Issue #1222, opened 2024-11-20
seekdir() support 


nrath-js:
Is it in principle be possible to add seekdir() support, or is there something fundamental that would prevent that?

If it can't be added directly, is there any way I could implement my own seekdir() but continue to use Rustix's Dir structure, or would I have to roll my own version of opendir(), closedir() and readdir() as well?

sunfishcode:
In principle, it is probably possible. I'm not aware of anything fundamental blocking it.

If you only need Linux support, you could probably implement your own seekdir using `rustix::fs::seek`/`lseek`. If you want portable support, it would need to be added to `Dir`.

nrath-js:
Working on this now.

It seems that in theory, the `_seek` syscall can fail, but libc's `seekdir` call cannot.

What do you prefer?

- Have the libc backend return a `Result`, even though it cannot fail
- Have the raw backend panic on seek error (as far as I can tell, this is reasonable since the only way for seek() to fail is if *whence* is invalid or the provided fd is not seekabe, i.e. if there is a bug in Rustix)
- Have different return values for seekdir depending on which backend is used
-

sunfishcode:
What if, if we get an error from the seek, we set the `any_errors` field of the `Dir` to true? That way we won't continue trying to iterate in that case.

nrath-js:
Would anyone have time to review the pull request for this? (#1231 )

PR #1231, opened 2024-12-03
Add support for seekdir().

nrath-js:
This adds support for `Dir::seekdir`, and exposes `d_off` in `Dirent`.

Fixes: #1222.

### DIFF
--- a/src/backend/libc/fs/dir.rs
+++ b/src/backend/libc/fs/dir.rs
@@ -129,6 +129,14 @@ impl Dir {
         unsafe { c::rewinddir(self.libc_dir.as_ptr()) }
     }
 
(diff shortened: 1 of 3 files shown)


nrath-js:
Friendly ping... could anyone take a look at this?

sunfishcode:
Thanks for your patience here, and for the PR!

sunfishcode:
@nrath-js The `test_dir_seek` test is failing on main on arm on the libc backend:

https://github.com/bytecodealliance/rustix/actions/runs/12680283634/job/35341748535

It's possible this is related to a 64-bit offset being casted to a 32-bit `c_long` for the `seekdir` call. Do you know of any alternative we might use here?

Issue #1230, opened 2024-12-02
Why is `fs::Dir` not `Sync`?


nrath-js:
I was surprised to see that the `fs::Dir` type is not marked as `Sync`.

As far as I can tell, all methods that access the internal state require a `&mut self`, so I think it's fine to use this type from multiple threads without the need for another mutex (since at most one thread can hold a `&mut` reference).

Am I missing something?

sunfishcode:
I think you're right; it could be `Sync`.

notgull:
It is `Sync` on the linux_raw backend, but not the libc one.

PR #1232, opened 2024-12-03
Mark libc's ``fs::Dir` as `Sync`

nrath-js:
This is safe since all methods that mutate the internal state require a &mut self. It also makes behavior consistent with the linux_raw backend (whose `fs::Dir` is Sync+Send).

Fixes: #1230.

### DIFF
--- a/src/backend/libc/fs/dir.rs
+++ b/src/backend/libc/fs/dir.rs
@@ -227,11 +227,12 @@ impl Dir {
     }
 }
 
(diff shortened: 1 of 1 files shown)


nrath-js:
I think the (Windows) test failures are unrelated to this change. Is that correct?

sunfishcode:
Thanks!

sunfishcode:

PR #1239, opened 2024-12-15
RISC-V: fix vDSO getcpu

hack3ric:
The name should be `__vdso_getcpu`, not `__kernel_getcpu` according to both the [manpage] and real RISC-V hardware.

[manpage]: https://man7.org/linux/man-pages/man7/vdso.7.html


sunfishcode:
Thanks!

PR #1240, opened 2024-12-16
net/linux_kernel: add unnamed Unix-domain addresses

Kijewski:
I think it would be useful to have unnamed Unix-domain addressed in rustix. This PR adds the methods `SocketAddrUnix::new_unnamed()` and `SocketAddrUnix::is_unnamed()`.

In C it is possible to have an [unnamed Unix-domain] socket name, when you set `len` = 2 = `sizeof(c::socklen_t)`. Then the kernel will choose an abstract Unix-domain name for you when you bind the socket. The same feature present also in Python, when you call [`sock.bind("")`].

Invoking [`SocketAddrUnix::new_abstract_name(b"")`] gives you an empty abstract socket address, i.e. `SocketAddrUnix::len == 3`. The kernel will keep this empty abstract name on calling `bind()`.

[unnamed Unix-domain]: https://manpages.debian.org/bookworm/manpages/unix.7.en.html#unnamed
[`sock.bind("")`]: https://docs.python.org/3.13/library/socket.html#socket.socket.bind
[`SocketAddrUnix::new_abstract_name(b"")`]: https://docs.rs/rustix/0.38.42/rustix/net/struct.SocketAddrUnix.html#method.new_abstract_name

### DIFF
--- a/src/backend/libc/net/addr.rs
+++ b/src/backend/libc/net/addr.rs
@@ -74,6 +74,26 @@ impl SocketAddrUnix {
         })
     }
 
+    /// Construct a new unnamed address.
+    ///
+    /// The kernel will assign an abstract Unix-domain address to the socket when you call
+    /// [`bind_unix()`][crate::net::bind_unix]. You can inspect the assigned name with
+    /// [`getsockname`][crate::net::getsockname].
+    ///
+    /// # References
+    ///  - [Linux]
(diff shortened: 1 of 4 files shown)


Kijewski:
Accidentally based in non-HEAD branch. Replaced by #1242.

PR #1241, opened 2024-12-16
io: always `impl Error for Errno` in rustc 1.81+

### LINKED ISSUE #1234: `rustix::io::Errno` should impl `core::error::Error` with rust 1.81+, regardless of `cfg(std)`
Presently, `rustix::io::Errno` conditionally impl's `std::error::Error` when the `std` feature is enabled.  The `Error` trait is available in `core` since Rust 1.81 (per <https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html>); please make an impl of `core::error::Error` unconditionally available when sufficiently new rustc is in use.

Kijewski:
In rust 1.81 the feature [`error_in_core`] was stabilized. This makes the `Error` trait available in `core::error` instead of `std::error`, so it can be used in a no-std context, too.

Resolves #1234.

[`error_in_core`]: <https://github.com/rust-lang/rust/issues/103765>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -259,6 +259,7 @@ check-cfg = [
     'cfg(core_intrinsics)',
     'cfg(criterion)',
     'cfg(document_experimental_runtime_api)',
+    'cfg(error_in_core)',
     'cfg(fix_y2038)',
     'cfg(freebsdlike)',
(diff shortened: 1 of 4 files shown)


sunfishcode:
Thanks!

PR #1242, opened 2024-12-16
net/linux_kernel: add unnamed Unix-domain addresses

Kijewski:
I think it would be useful to have unnamed Unix-domain addressed in rustix. This PR adds the methods `SocketAddrUnix::new_unnamed()` and `SocketAddrUnix::is_unnamed()`.

In C it is possible to have an [unnamed Unix-domain] socket name, when you set `len` = 2 = `sizeof(c::socklen_t)`. Then the kernel will choose an abstract Unix-domain name for you when you bind the socket. The same feature present also in Python, when you call [`sock.bind("")`].

Invoking [`SocketAddrUnix::new_abstract_name(b"")`] gives you an empty abstract socket address, i.e. `SocketAddrUnix::len == 3`. The kernel will keep this empty abstract name on calling `bind()`.

[unnamed Unix-domain]: https://manpages.debian.org/bookworm/manpages/unix.7.en.html#unnamed
[`sock.bind("")`]: https://docs.python.org/3.13/library/socket.html#socket.socket.bind
[`SocketAddrUnix::new_abstract_name(b"")`]: https://docs.rs/rustix/0.38.42/rustix/net/struct.SocketAddrUnix.html#method.new_abstract_name

### DIFF
--- a/src/backend/libc/net/addr.rs
+++ b/src/backend/libc/net/addr.rs
@@ -121,19 +121,10 @@ impl SocketAddrUnix {
     /// For a filesystem path address, return the path.
     #[inline]
     pub fn path(&self) -> Option<&CStr> {
-        let len = self.len();
-        if len != 0 && self.unix.sun_path[0] != 0 {
-            let end = len as usize - offsetof_sun_path();
-            let bytes = &self.unix.sun_path[..end];
-            // SAFETY: `from_raw_parts` to convert from `&[c_char]` to `&[u8]`.
-            // And `from_bytes_with_nul_unchecked` since the string is
-            // NUL-terminated.
-            unsafe {
-                Some(CStr::from_bytes_with_nul_unchecked(slice::from_raw_parts(
-                    bytes.as_ptr().cast(),
-                    bytes.len(),
-                )))
-            }
+        let bytes = self.bytes()?;
+        if !bytes.is_empty() && bytes[0] != 0 {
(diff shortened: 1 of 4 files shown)


src/backend/libc/net/addr.rs
```diff
@@ -123,17 +143,23 @@ impl SocketAddrUnix {
     #[cfg(linux_kernel)]
     #[inline]
     pub fn abstract_name(&self) -> Option<&[u8]> {
-        let len = self.len();
-        if len != 0 && self.unix.sun_path[0] == 0 {
-            let end = len as usize - offsetof_sun_path();
+        let end = self.len().saturating_sub(offsetof_sun_path());
```
sunfishcode, line 146: Can you comment on why you changed this to use `saturating_sub`?

Kijewski: I wanted to remove one indentation level by removing the test if `len == 0`, followed by `len > offsetof_sun_path`. I don't know if there actually is a case where `len == 0`. If there isn't, then a normal subtraction would be good enough. Alternatively, because the method returns an `Option`, `checked_sub` would work, too.

Kijewski:
I refactored all three methods, `SocketAddrUnix::path()`, `…::abstract_name()` and `…::is_unnamed()`, to deduplicate code and make it (IMHO) easier to read. Please tell me, if I should revert the change.

sunfishcode:
Thanks!

PR #1244, opened 2024-12-17
Add setdomainname function

yuki0iq:


src/backend/libc/system/syscalls.rs
```diff
@@ -63,6 +63,30 @@ pub(crate) fn sethostname(name: &[u8]) -> io::Result<()> {
...
+    target_os = "vita",
+    target_os = "wasi"
+)))]
+pub(crate) fn setdomainname(name: &[u8]) -> io::Result<()> {
+    syscall! {
+        fn setdomainname(
+            name: *const c::c_char,
+            len: c::size_t
+        ) via SYS_setdomainname -> c::c_int
+    }
```
sunfishcode, line 80: At a quick glance, the libc crate does have declarations for `setdomainname`; could you comment on why tthis uses `syscall!` instead of just `libc::setdomainname`?

yuki0iq: Turns out `libc::setdomainname` isn't there for android (failed CI)

sunfishcode: It's unfortunate to make non-Linux users use a weak symbol though. Could you use `libc::setdomainname` on all platforms except Android using `cfg`s?

I also filed https://github.com/rust-lang/libc/pull/4212 to add the declarations to libc for Android.

yuki0iq: Fixed

sunfishcode:
Thanks!

PR #1250, opened 2025-01-04
Implement `raise` and `tgkill`

joshtriplett:
This implements the `tgkill` system call, and the `raise` function (which sends a signal to the current thread).

I've implemented these in the `runtime` module for consistency, though I'd ideally like to have these available somewhere less "beware of dragons".


sunfishcode:
Yes, `raise` and `tgkill` can go in `rustix::thread`, and be implemented in the libc backend as well. `tgkill` is `cfg(linux_kernel)`, and on musl it looks like it'll need to go through `syscall`.

Issue #1253, opened 2025-01-07
Segmentation fault on Apple M2


steffenvongrabau:
I've stumbled upon this issue while trying to run [Niri](https://github.com/YaLTeR/niri) and it instantly segfaulted. 
Using GDB I generated the following backtrace, which shows: Segmentation fault.
init_from_sysinfo_ehdr () at src/backend/linux_raw/vdso.rs:187
[gdb.txt](https://github.com/user-attachments/files/18331561/gdb.txt) 

I cloned this git repository and ran `cargo test --features=all-apis --no-fail-fast` which showed multiple tests failing. 
[cargo_test.txt](https://github.com/user-attachments/files/18332801/cargo_test.txt)

tests/time/main.rs segfaults with no further error message, so I manually ran target/debug/deps/time-ebc378475d0c4529 with rust-gdb to generate this backtrace:
[gdb_time.txt](https://github.com/user-attachments/files/18332811/gdb_time.txt)
Like the Niri backtrace it also shows: Segmentation fault.
rustix::backend::vdso::init_from_sysinfo_ehdr () at src/backend/linux_raw/vdso.rs:187

### System Information
* Gentoo aarch64
    * KernelPageSize: 16KB
    * MMUPageSize: 16KB 
* CPU: Apple M2
* rustix version: 0.38.35

steffenvongrabau:
Thanks a lot for the quick response! I've tested the patch in #1254 and tests/time/main.rs passed but test backend::vdso::test:vdso fails now.
[cargo_test_patch1254.txt](https://github.com/user-attachments/files/18335735/cargo_test_patch1254.txt)

vdso_test_gettimeofday outputs `Could not find __kernel_gettimeofday`
Please let me know if there's anything else I can do.

sunfishcode:
This may be caused by https://github.com/torvalds/linux/commit/48f6430505c0b0498ee9020ce3cf9558b1caaaeb, in which case it's not a bug. It appears the changes to Linux's reference parser haven't landed yet, but it looks like https://lore.kernel.org/lkml/b3b02cc6-69c0-4cde-b459-cb7c7fab9f30@kernel.org/T/#r3166a373d53e9beb19a33abcbd3eb131dcdcc107 is the needed change.

sunfishcode:
[...]

PR #1259, opened 2025-01-08
Implement the `DT_GNU_HASH` mechanism for vDSO parsing.

sunfishcode:
Linux recently [removed] the `DT_HASH` section from the aarch64 vDSO. To continue to be able to read vDSOs, implement the `DT_GNU_HASH` section, following the logic in [this patch].

[removed]: https://github.com/torvalds/linux/commit/48f6430505c0b0498ee9020ce3cf9558b1caaaeb
[this patch]: https://lkml.org/lkml/2024/12/6/828

### DIFF
--- a/src/backend/linux_raw/process/syscalls.rs
+++ b/src/backend/linux_raw/process/syscalls.rs
@@ -41,7 +41,8 @@ use {crate::backend::conv::slice_just_addr_mut, crate::process::Gid};
     target_arch = "x86_64",
     target_arch = "x86",
     target_arch = "riscv64",
-    target_arch = "powerpc64"
+    target_arch = "powerpc64",
+    target_arch = "s390x"
 ))]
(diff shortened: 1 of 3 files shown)


ids1024:
I wonder if `vdso.rs` could use less unsafe code. If the vDSO is just a region of memory that isn't mutated, it could be cast as slice (or portions of it could be; I guess ELF headers don't have a field that simply represents the size of the whole elf binary?). Something like https://github.com/bytecodealliance/rustix/issues/1253 should at least not be a segfault.

Though I guess if it mostly just matches the reference implementation in C, this would make it a bit less close.

sunfishcode:
@ids1024 My sense is that keeping the code close to the reference implementation is more valuable in this situation. The ELF data structures are pretty subtle, and it's easy to accidentally use them in ways that work but aren't guaranteed by the ABI. By sticking super close to the reference code, we can be more confident that we're only depending on things that matter.

PR #1285, opened 2025-01-23
Change `poll`'s timeout from `c_int` to `Option<&Timespec>`.

sunfishcode:
This harmonizes the timeout with the rest of rustix, which uses `Timespec` for all time values. And, it eliminates the awkwardness of using `-1` as a sentinel value.

On platforms with `ppoll`, the `Timespec` can be passed straight to the OS. On platforms without `ppoll`, we have to do a fallible conversion into `c_int`.

### DIFF
--- a/src/backend/libc/event/syscalls.rs
+++ b/src/backend/libc/event/syscalls.rs
@@ -164,8 +164,31 @@ pub(crate) fn poll(fds: &mut [PollFd<'_>], timeout: Option<&Timespec>) -> io::Re
             }
         };
 
-        ret_c_int(unsafe { c::ppoll(fds.as_mut_ptr().cast(), nfds, timeout, null()) })
-            .map(|nready| nready as usize)
+        #[cfg(not(target_os = "netbsd"))]
+        {
+            ret_c_int(unsafe { c::ppoll(fds.as_mut_ptr().cast(), nfds, timeout, null()) })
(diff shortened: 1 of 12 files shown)


SUPERCILEX:
Will take a look sometime this week.

src/backend/libc/event/syscalls.rs
```diff
@@ -119,14 +127,96 @@ pub(crate) unsafe fn kevent(
...
+            None => -1,
+            Some(timeout) => {
+                // Convert from `Timespec` to `c_int` milliseconds.
+                let secs = timeout.tv_sec;
+                if secs < 0 {
+                    return Err(io::Errno::INVAL);
+                }
+                secs.checked_mul(1000)
+                    .and_then(|millis| {
+                        millis.checked_add((i64::from(timeout.tv_nsec) + 999_999) / 1_000_000)
```
SUPERCILEX, line 211: Can we add a comment that says the purpose of this is to round up to the nearest millisecond?

SUPERCILEX (approved):

src/backend/libc/event/syscalls.rs
sunfishcode: Done.

SUPERCILEX: Sweet, this looks amazing!

PR #1286, opened 2025-01-23
Make Ioctl::opcode a method instead of a constant

coolreader18:
This lets the `Ioctl` trait be compatible with ioctls that take a runtime length value, e.g. these evdev ioctls from `input.h`:
```h
#define EVIOCGNAME(len)		_IOC(_IOC_READ, 'E', 0x06, len)		/* get device name */
#define EVIOCGPHYS(len)		_IOC(_IOC_READ, 'E', 0x07, len)		/* get physical location */
#define EVIOCGUNIQ(len)		_IOC(_IOC_READ, 'E', 0x08, len)		/* get unique identifier */
#define EVIOCGPROP(len)		_IOC(_IOC_READ, 'E', 0x09, len)		/* get device properties */
```
These generally specify the length of the passed buffer, so that the driver can copy data into userspace (à la `read()`). `rg 'len\)\s*_IO' /usr/include` gives 26 results on my system, so they're not rare, and it'd be great to be able to use the `Ioctl` trait to wrap them.

Prior art: [`nix::ioctl_read_buf!()`](https://docs.rs/nix/latest/nix/macro.ioctl_read_buf.html)

### DIFF
--- a/src/fs/ioctl.rs
+++ b/src/fs/ioctl.rs
@@ -79,7 +79,10 @@ unsafe impl ioctl::Ioctl for Ficlone<'_> {
     type Output = ();
 
     const IS_MUTATING: bool = false;
-    const OPCODE: ioctl::Opcode = ioctl::Opcode::old(c::FICLONE as ioctl::RawOpcode);
+
+    fn opcode(&self) -> ioctl::Opcode {
+        ioctl::Opcode::old(c::FICLONE as ioctl::RawOpcode)
+    }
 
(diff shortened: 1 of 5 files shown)


sunfishcode:
This looks reasonable to me. @notgull do you have any thoughts on `Ioctl`'s `OPCODE` changing to `opcode()`?

notgull:
This should be fine. I mistakenly believed that the opcode was always a constant and that const evaluation could fill in what gaps there were.

sunfishcode:
Thanks!

PR #1291, opened 2025-01-27
Add setter and getter for `multicast_ifv4` and `multicast_ifv6`

al8n:
Hi, this is the PR for #1299.

### DIFF
--- a/src/net/sockopt.rs
+++ b/src/net/sockopt.rs
@@ -647,6 +647,50 @@ pub fn ipv6_v6only<Fd: AsFd>(fd: Fd) -> io::Result<bool> {
     backend::net::sockopt::ipv6_v6only(fd.as_fd())
 }
 
+/// `setsockopt(fd, IPPROTO_IP, IP_MULTICAST_IF, value)`
+///
+/// See the [module-level documentation] for more.
+///
+/// [module-level documentation]: self#references-for-get_ip_-and-set_ip_-functions
+#[inline]
+#[doc(alias = "IP_MULTICAST_IF")]
+pub fn set_ip_multicast_ifv4<Fd: AsFd>(fd: Fd, value: &Ipv4Addr) -> io::Result<()> {
+    backend::net::sockopt::set_ip_multicast_ifv4(fd.as_fd(), value)
+}
+
+/// `getsockopt(fd, IPPROTO_IP, IP_MULTICAST_IF)`
+///
+/// See the [module-level documentation] for more.
+///
+/// [module-level documentation]: self#references-for-get_ip_-and-set_ip_-functions
+#[inline]
+#[doc(alias = "IP_MULTICAST_IF")]
+pub fn ip_multicast_ifv4<Fd: AsFd>(fd: Fd) -> io::Result<Ipv4Addr> {
+    backend::net::sockopt::ip_multicast_ifv4(fd.as_fd())
+}
+
+/// `setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_IF, value)`
+///
+/// See the [module-level documentation] for more.
+///
+/// [module-level documentation]: self#references-for-get_ip_-and-set_ip_-functions
+#[inline]
+#[doc(alias = "IPV6_MULTICAST_IF")]
+pub fn set_ip_multicast_ifv6<Fd: AsFd>(fd: Fd, value: u32) -> io::Result<()> {
+    backend::net::sockopt::set_ip_multicast_ifv6(fd.as_fd(), value)
+}
+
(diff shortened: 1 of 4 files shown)


src/net/sockopt.rs
```diff
@@ -647,6 +647,50 @@ pub fn ipv6_v6only<Fd: AsFd>(fd: Fd) -> io::Result<bool> {
...
+///
+/// See the [module-level documentation] for more.
+///
+/// [module-level documentation]: self#references-for-get_ip_-and-set_ip_-functions
+#[inline]
+#[doc(alias = "IP_MULTICAST_IF")]
+pub fn set_ip_multicast_ifv4<Fd: AsFd>(fd: Fd, value: &Ipv4Addr) -> io::Result<()> {
+    backend::net::sockopt::set_ip_multicast_ifv4(fd.as_fd(), value)
+}
+
```
sunfishcode, line 660: The [Linux docs](https://man7.org/linux/man-pages/man7/ip.7.html) for `IP_MULTICAST_IF` say "The argument for [setsockopt(2)](https://man7.org/linux/man-pages/man2/setsockopt.2.html) is an ip_mreqn or (since Linux 3.5) ip_mreq structure similar to IP_ADD_MEMBERSHIP, or an
 in_addr structure.". Would it make sense here to also add `set_ip_multicast_ifv4_with_ifindex`, similar to the existing `set_ip_add_membership_with_ifindex`, which uses the `ip_mreqn` form?

src/net/sockopt.rs
```diff
@@ -647,6 +647,50 @@ pub fn ipv6_v6only<Fd: AsFd>(fd: Fd) -> io::Result<bool> {
...
 }
 
+/// `setsockopt(fd, IPPROTO_IP, IP_MULTICAST_IF, value)`
+///
+/// See the [module-level documentation] for more.
+///
+/// [module-level documentation]: self#references-for-get_ip_-and-set_ip_-functions
+#[inline]
+#[doc(alias = "IP_MULTICAST_IF")]
+pub fn set_ip_multicast_ifv4<Fd: AsFd>(fd: Fd, value: &Ipv4Addr) -> io::Result<()> {
```
sunfishcode, line 657: For naming, the `*_ip_*` names are all specific to IPv4, so we shouldn't need the `v4` suffix here and in similar functions.

src/net/sockopt.rs
```diff
@@ -647,6 +647,50 @@ pub fn ipv6_v6only<Fd: AsFd>(fd: Fd) -> io::Result<bool> {
...
+}
+
+/// `setsockopt(fd, IPPROTO_IPV6, IPV6_MULTICAST_IF, value)`
+///
+/// See the [module-level documentation] for more.
+///
+/// [module-level documentation]: self#references-for-get_ip_-and-set_ip_-functions
+#[inline]
+#[doc(alias = "IPV6_MULTICAST_IF")]
+pub fn set_ip_multicast_ifv6<Fd: AsFd>(fd: Fd, value: u32) -> io::Result<()> {
```
sunfishcode, line 679: IPv6 sockopt functions should use `_ipv6_` instead of `_ip_`, and then they don't need a `v6` suffix.

sunfishcode, line 679: Other sockopts that have an interface index argument use `i32`, rather than `u32`. Is there a reason to prefer `u32`?

sunfishcode (commented):
Thanks for working on these!

src/net/sockopt.rs
al8n: Hi, `u32` is just for the compatible with [`socket2::set_multicast_if_v6`](https://docs.rs/socket2/latest/socket2/struct.Socket.html#method.set_multicast_if_v6).

sunfishcode: Ah. Well, rustix is preparing for a major semver bump soon, so we can change that if we want. Or we can change `set_ip_add_membership_with_ifindex` and `set_ip_drop_membership_with_ifindex` to use `u32`. Which do you think is better?

al8n: IMO, changing it to `u32` is better:
1. compatible with `socket2`'s `set_ip_add_membership_with_ifindex ` and `set_ip_drop_membership_with_ifindex `
2. A negative interface index makes no sense and it is an invalid value, so `u32` can avoid misusing the API.

Should I change it in this PR?

sunfishcode: Yes, changing it in this PR would be good. Thanks!

src/net/sockopt.rs
al8n: Yes, add it.

sunfishcode:
Thanks!

PR #1292, opened 2025-01-27
Support `SIGRTMIN+n` signals.

### LINKED ISSUE #1249: `Signal` should support arbitrary signals
The `Signal` enum is missing the real-time signals, from `SIGRTMIN` to `SIGRTMAX`.

Handling these will be interesting, because glibc reserves the first two of them, and does so by changing its exposed value of `SIGRTMIN`.

I would suggest exposing the same `SIGRTMIN` that glibc does (34 rather than 32), and additionally exposing enum values for the two reserved signal numbers, with names that clearly indicate that they're reserved by libc.

That way, when people port from C to Rust, they won't get surprised by the difference.

sunfishcode:
Change `Signal` from an `enum` to a `struct` wrapper around an `i32` so that it can accomodate values it doesn't know about statically, and add APIs for constructing and working with `SIGRTMIN+n` signal values.

This requires making `Signal::from_raw` depend on a new "use-libc-sigrt" feature, as it depends on being able to call into libc. This also adds `Signal::from_raw_unchecked` which does not call into libc and is unsafe.

Fixes #1249.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,7 +85,7 @@ name = "mod"
 harness = false
 
 [package.metadata.docs.rs]
-features = ["all-apis"]
+features = ["all-apis", "use-libc-sigrt"]
 targets = [
     "x86_64-unknown-linux-gnu",
     "i686-unknown-linux-gnu",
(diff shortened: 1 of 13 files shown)


joshtriplett:
This looks great, thank you.

PR #1300, opened 2025-01-30
Add a way to get the raw fields out of a `WaitStatus`.

sunfishcode:


sunfishcode:
This hides the linux-raw-sys type in the public API, and will be used by c-scape to replace [this code](https://github.com/sunfishcode/c-ward/blob/80e48875698c31deabcadb7dc15515a9058fd6a7/c-scape/src/process/wait.rs#L100).

PR #1308, opened 2025-02-02
Derive Ord for Timespec

yuki0iq:
Meaningless if `Timespec`s were from different kind (duration and timestamp).

Maybe add conversion to/from `core::time::Duration` and some arithmetic?


sunfishcode:
Makes sense to me. An yes, I'd accept PRs adding conversions to/from `core::time::Duration` and some arithmetic.

PR #1309, opened 2025-02-04
Add an `epoll::wait_uninit`.

sunfishcode:
This is like `epoll::wait`, except that it can read into an uninitialized buffer, and it returns a pair of slices, similar to `read_uninit`. This also means it doesn't require "alloc".

This is a possible alternative direction to #1290. I'm on the fence still. I like some things about the `Buffer` trait magic, including how it can encapsulate the unsafety of the `set_len` to resize a `Vec` to the number of initialized bytes. However, I don't like how it seems to feel more magical than it is. In `read(fd, extend(&mut v))`, you still have to think about end-of-stream, `EINTR`, short reads, and spare capacity.


sunfishcode:
Closing in favor of https://github.com/bytecodealliance/rustix/pull/1290, which I'm feeling more confident in now.

PR #1310, opened 2025-02-05
Add `fcntl_getlk` for fetching information of a lock held by another process

### LINKED ISSUE #1076: Feature request: Add `fcntl_getlk` support

metent:
Fixes #1076.

I added this inside the `process` module as `Pid` was inaccessible in the `fs` module, and it makes more sense to include it inside `process` as we can use it to access the pid, and lock data which is held by another process. I read the code for `rustix::fs::fcntl_lock`, and added this functionality in a similar manner, but let me know if the APIs/code organization need work.

I still need to add a test for this though, but I'm not sure what would be the best way to do that. As a process will always be able to acquire it's own lock, we'll probably need to spawn another process and read the advisory lock set by the parent process over there. We may use `std::os::unix::process::CommandExt::before_exec` for forking but it might not be the best way, given that it's unsafe and we'll have to handle race conditions, so I need some guidance here.

EDIT: Used `std::os::unix::process::CommandExt::pre_exec` for spawning child processes in tests for now.

EDIT 2: After running the tests, I found that MacOS and FreeBSD are the only platform which don't return `EINVAL` when `l_type` is set to `F_UNLCK`. For ensuring consistent behavior, I've added a check [here](https://github.com/bytecodealliance/rustix/pull/1310/files#diff-3f3e90e7733855d35135272bb9f04f01aef49e024cbcb90afe5da63a86ee5b53R669-R671). Let me know if I should respect platform-specific behavior or keep the status quo.

### DIFF
--- a/src/backend/libc/process/syscalls.rs
+++ b/src/backend/libc/process/syscalls.rs
@@ -666,12 +666,6 @@ pub(crate) fn getgroups(buf: &mut [Gid]) -> io::Result<usize> {
 )))]
 #[inline]
 pub(crate) fn fcntl_getlk(fd: BorrowedFd<'_>, lock: &Flock) -> io::Result<Option<Flock>> {
-    // In order to guarantee consistent behavior across all POSIX platforms, return `EINVAL` when
-    // `flock.l_type` is `F_UNLCK`
-    if lock.typ == crate::process::FlockType::Unlocked {
-        return Err(io::Errno::INVAL);
-    }
-
     let mut curr_lock: c::flock = lock.as_raw();
     unsafe { ret(c::fcntl(borrowed_fd(fd), c::F_GETLK, &mut curr_lock))? };
 
--- a/src/backend/libc/process/syscalls.rs
+++ b/src/backend/libc/process/syscalls.rs
@@ -29,6 +29,15 @@ use crate::ffi::CStr;
 #[cfg(feature = "fs")]
 use crate::fs::Mode;
 use crate::io;
+#[cfg(not(any(
+    target_os = "emscripten",
+    target_os = "espidf",
+    target_os = "fuchsia",
+    target_os = "redox",
+    target_os = "vita",
+    target_os = "wasi"
+)))]
+use crate::process::Flock;
 #[cfg(all(feature = "alloc", not(target_os = "wasi")))]
 use crate::process::Gid;
 #[cfg(not(target_os = "wasi"))]
@@ -43,7 +52,7 @@ use crate::process::Signal;
 )))]
 use crate::process::Uid;
 #[cfg(not(any(target_os = "espidf", target_os = "vita", target_os = "wasi")))]
-use crate::process::{Flock, RawPid, WaitOptions, WaitStatus};
+use crate::process::{RawPid, WaitOptions, WaitStatus};
 #[cfg(not(any(
     target_os = "espidf",
     target_os = "fuchsia",
@@ -647,6 +656,14 @@ pub(crate) fn getgroups(buf: &mut [Gid]) -> io::Result<usize> {
     unsafe { ret_usize(c::getgroups(len, buf.as_mut_ptr().cast()) as isize) }
 }
 
+#[cfg(not(any(
+    target_os = "emscripten",
+    target_os = "espidf",
+    target_os = "fuchsia",
+    target_os = "redox",
+    target_os = "vita",
+    target_os = "wasi"
+)))]
 #[inline]
 pub(crate) fn fcntl_getlk(fd: BorrowedFd<'_>, lock: &Flock) -> io::Result<Option<Flock>> {
     let mut curr_lock: c::flock = lock.as_raw();
--- a/src/backend/libc/process/syscalls.rs
+++ b/src/backend/libc/process/syscalls.rs
@@ -666,6 +666,8 @@ pub(crate) fn getgroups(buf: &mut [Gid]) -> io::Result<usize> {
 )))]
 #[inline]
 pub(crate) fn fcntl_getlk(fd: BorrowedFd<'_>, lock: &Flock) -> io::Result<Option<Flock>> {
+    // In order to guarantee consistent behavior across all POSIX platforms, return `EINVAL` when
+    // `flock.l_type` is `F_UNLCK`
     if lock.typ == crate::process::FlockType::Unlocked {
         return Err(io::Errno::INVAL);
     }
--- a/src/backend/libc/process/syscalls.rs
+++ b/src/backend/libc/process/syscalls.rs
@@ -673,7 +673,8 @@ pub(crate) fn fcntl_getlk(fd: BorrowedFd<'_>, lock: &Flock) -> io::Result<Option
     let mut curr_lock: c::flock = lock.as_raw();
     unsafe { ret(c::fcntl(borrowed_fd(fd), c::F_GETLK, &mut curr_lock))? };
 
-    if curr_lock.l_pid == Pid::as_raw(lock.pid) {
+    // If no blocking lock is found, `fcntl(GETLK, ..)` sets `l_type` to `F_UNLCK`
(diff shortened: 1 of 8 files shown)


metent:
Apparently someone else added a PR for this before(https://github.com/bytecodealliance/rustix/pull/1077). I couldn't find such issue or PR the last time I searched.

src/process/types.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// File lock data structure used in [`fcntl_getlk`].
+///
+/// [`fcntl_getlk`]: crate::fs::fcntl_getlk
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Flock {
+    /// Starting offset for lock
+    pub start: isize,
+    /// Number of bytes to lock
+    pub length: isize,
```
sunfishcode, line 15: Rustix exposes file sizes and offsets as 64-bit whenever possible, and follows [`std::io::SeekFrom`](https://doc.rust-lang.org/stable/std/io/enum.SeekFrom.html) in using `u64` for absolute offsets. Could you use `u64` for the start and length?

src/backend/libc/process/syscalls.rs
```diff
@@ -646,3 +655,27 @@ pub(crate) fn getgroups(buf: &mut [Gid]) -> io::Result<usize> {
...
+    target_os = "fuchsia",
+    target_os = "redox",
+    target_os = "vita",
+    target_os = "wasi"
+)))]
+#[inline]
+pub(crate) fn fcntl_getlk(fd: BorrowedFd<'_>, lock: &Flock) -> io::Result<Option<Flock>> {
+    if lock.typ == crate::process::FlockType::Unlocked {
+        return Err(io::Errno::INVAL);
+    }
```
sunfishcode, line 671: Is it necessary to do this check here, or would it be ok to just expose the raw behavior of the `fcntl` here? If we have a check like this, we should document it.

src/backend/libc/process/syscalls.rs
```diff
@@ -646,3 +655,27 @@ pub(crate) fn getgroups(buf: &mut [Gid]) -> io::Result<usize> {
...
+
+    let mut curr_lock: c::flock = lock.as_raw();
+    unsafe { ret(c::fcntl(borrowed_fd(fd), c::F_GETLK, &mut curr_lock))? };
+
+    if curr_lock.l_pid == Pid::as_raw(lock.pid) {
```
sunfishcode, line 676: Could you comment on what this code is doing? I'm looking at the documentation for `F_GETLK` and it's not clear what this condition corresponds to.

src/backend/linux_raw/c.rs
```diff
@@ -184,12 +184,21 @@ pub(crate) const CLONE_CHILD_SETTID: c_int = linux_raw_sys::general::CLONE_CHILD
...
+        CLD_CONTINUED, CLD_DUMPED, CLD_EXITED, CLD_KILLED, CLD_STOPPED, CLD_TRAPPED, F_RDLCK,
+        F_UNLCK, F_WRLCK, O_NONBLOCK as PIDFD_NONBLOCK, P_ALL, P_PGID, P_PID, P_PIDFD, SEEK_CUR,
+        SEEK_END, SEEK_SET,
     },
     ioctl::TIOCSCTTY,
 };
 
+#[cfg(feature = "process")]
+#[cfg(target_pointer_width = "32")]
+pub(crate) use linux_raw_sys::general::{flock64 as flock, F_GETLK64 as F_GETLK};
```
sunfishcode, line 196: Rustix does often alias `*64` things to non`*64` names, but in the case of `F_GETLK64`, it's used with `__NR_fcntl64` which is not currently aliased. Could you either remove this `F_GETLK64` alias and just use `F_GETLK64` in the code, or, if you prefer, add an `__NR_fcntl` alias for `__NR_fcntl64` alias, and use that?

sunfishcode (commented):
Thanks!

I think putting this in `rustix::process` makes sense, for this and for furture fcntl-style lock functions which expose the full fcntl-style process-associated lock API. The current `rustix::fs::fcntl_getlk` doesn't expose the full fcntl-style lock functionality, and is mainly intended as an alternative for `flock` on platforms which lack `flock`. So we can evolve toward `rustix::process::fcntl_*` eventually being the "full" fcntl-style lock API.

src/backend/libc/process/syscalls.rs
metent: I think we should keep the check. Here's my reasoning:

The POSIX specification of `F_GETLK` states that the third argument `arg` is the **lock** description and it explictly mentions the tested creation of the lock:

> Get any lock which blocks the process-owned file **lock description pointed to by the third argument, arg**, taken as a pointer to type struct flock, defined in <fcntl.h>. The information retrieved shall overwrite the information passed to fcntl() in the structure flock. If no lock is found that would prevent **this lock from being created**, then the structure shall be left unchanged except for the lock type in l_type which shall be set to F_UNLCK.

The 'ERRORS' sections states that if the `cmd` argument is `F_GETLK` and the data pointed by it is invalid, then the error is set to `EINVAL`.

> ERRORS
> The fcntl() function shall fail if:
> 
> [EINVAL]
> The cmd argument is invalid; or the cmd argument is F_DUPFD, F_DUPFD_CLOEXEC, or F_DUPFD_CLOFORK and arg is negative or is greater than or equal to {OPEN_MAX}; or the cmd argument is F_SETOWN_EX and the type member of the f_owner_ex structure pointed to by arg is invalid, or the pid member is negative and the type member is F_OWNER_PID or F_OWNER_PGRP; **or the cmd argument is F_GETLK**, F_SETLK, F_SETLKW, F_OFD_GETLK, F_OFD_SETLK, or F_OFD_SETLKW **and the data pointed to by arg is not valid**, or fildes refers to a file that does not support locking.

So, according to POSIX `arg` should be a valid lock description which tests the creation of the lock, and if `l_type` is set to `F_UNLCK` it's a request to unlock, so it should be an invalid argument.

Furthermore, I tried running `fcntl_getlk` with `l_type` set to `F_UNLCK` in MacOS CI, and the result was a no-op (No mutation of passed flock and no error returned). This may also be the case for `FreeBSD` and other OS's, and I think it's best to keep this check for every target, so that we don't have to investigate platform-specific behavior.

I noticed that `FreeBSD` doesn't follow POSIX specification for `fadvise` either, and in `rustix::fd::fadvise`, there is some additional logic for ensuring POSIX compatibility. I'll document the code in a similar way for now, but let me know if something else needs to be done.

metent:
I believe I've addressed everything, let me know if any other change is needed.

src/backend/libc/process/syscalls.rs
sunfishcode: The `fadvise` case is technically about bridging the gap from POSIX's signed `off_t` to Rust's convention of using unsigned `u64`, which rustix mentions in src/lib.rs. That said, looking at it now, I agree that we should clarify that in the documentation. I'll submit a PR.

Looking a `F_GETLK` more closely, calling `fcntl_getlk` with `l_type` set to `F_UNLCK` looks like it would be a bug; application code shouldn't ever want to do that. There are a lot of minor portability hazards and POSIX nonconformances that rustix doesn't hide, so my sense is that we shouldn't make an exception and hide this one. It certainly makes sense to document it though,

sunfishcode (commented):
This looks great! I just have the one comment above. While I understand the desire to add code to hide platform differences, rustix in general doesn't do that, especially for error conditions, an I'm reluctant to start doing it here.

metent:
> There are a lot of minor portability hazards and POSIX nonconformances that rustix doesn't hide

Makes sense if rustix is aiming for a relatively low-level API. I've removed the check and updated the docs and tests accordingly.

sunfishcode:
Thanks!

sunfishcode:

PR #1311, opened 2025-02-05
Add addition, subtraction and conversion to/from `core::time::Duration` for Timespec

yuki0iq:
Follow-up for #1308


sunfishcode:
Thanks!

PR #1313, opened 2025-02-06
Document experimental runtime API on docs.rs

### LINKED ISSUE #1180: Feature request: Document rustix::runtime
rustix does not currently expose the `_exit` system call as far as I can tell.  This would be useful in at least two contexts:

- On the child side of a `fork` (which is also missing right now but I trust it's already on your list of things that need to be added)
- In `#[no_std]` crates that nonetheless have the kernel available (e.g. the various projects trying to make it possible to write Rust programs that do not use a C library).

yuki0iq:
One can also use `cargo docs-rs` locally for viewing docs.

Closes #1180

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,6 +99,7 @@ targets = [
     "wasm32-unknown-emscripten",
     "wasm32-wasip1",
 ]
+rustdoc-args = ["--cfg", "document_experimental_runtime_api"]
 
 [features]
 


sunfishcode:
I don't believe `rustix::runtime` is suitable to be stabilized and publicly documented on docs.rs in its current form. The safety conditions do not document what is required to use these APIs.

If you would like to see specific functions in `rustix::runtime` become publicly documented and stabilized, please bring them up for discussion.

sunfishcode:
Thinking about what rustix could do to help in this area, I've now posted these two items:

https://github.com/bytecodealliance/rustix/pull/1315 adds documentation for these functions, so that at least if someone is searching for them in the docs, they'll find more information.

I've now also filed https://github.com/bytecodealliance/rustix/issues/1314 to encourage people to post about functions in rustix::runtime that would be useful to publicly expose.

Issue #1252, opened 2025-01-06
Netlink socket protocols are largely actually socket options, which are also missing


DianaNites:
https://docs.rs/rustix/0.38.42/rustix/net/sockopt/index.html

Many of the supposed netlink protocols are in reality netlink specific socket options, which are also completely missing in the sockopt module

https://docs.rs/rustix/0.38.42/rustix/net/netlink/constant.ADD_MEMBERSHIP.html
https://docs.rs/rustix/0.38.42/rustix/net/netlink/constant.DROP_MEMBERSHIP.html
https://docs.rs/rustix/0.38.42/rustix/net/netlink/constant.LIST_MEMBERSHIPS.html
https://docs.rs/rustix/0.38.42/rustix/net/netlink/constant.BROADCAST_ERROR.html
https://docs.rs/rustix/0.38.42/rustix/net/netlink/constant.NO_ENOBUFS.html
https://docs.rs/rustix/0.38.42/rustix/net/netlink/constant.CAP_ACK.html
https://docs.rs/rustix/0.38.42/rustix/net/netlink/constant.EXT_ACK.html
https://docs.rs/rustix/0.38.42/rustix/net/netlink/constant.GET_STRICT_CHK.html
https://docs.rs/rustix/0.38.42/rustix/net/netlink/constant.LISTEN_ALL_NSID.html
https://docs.rs/rustix/0.38.42/rustix/net/netlink/constant.PKTINFO.html

all of these are socket options described in [`netlink.7`](https://man.archlinux.org/man/netlink.7.en) or [the netlink handbook](https://docs.kernel.org/userspace-api/netlink/intro.html#strict-checking).

----

Additionally, when I read rustix was low-level and "aims to provide safe and idiomatic Rust interfaces to low-level syscalls", I was surprised to find no direct interface to many syscalls, but instead high level wrappers exclusively exposing specific subsets of them that rustix knows about, and nothing for those it doesn't.

I was expecting a lower level API to syscalls like this, with all the benefits of type, memory, and I/O safety rustix provides over the raw syscalls and bare integers, but that doesn't seem to be a design goal for rustix.

sunfishcode:
I don't know how it could be possible to expose a lower-level API while still having any of the benefits of type, memory, or I/O safety. To provide a safe interface, rustix has to know which arguments are pointers, which are file descriptors, what memory is to be mutated, and so on.

[...]

PR #1336, opened 2025-02-16
Remove netlink protocol constants that are actually socket options

kevinmehall:
From https://github.com/bytecodealliance/rustix/issues/1252: these are not protocols but rather socket options. Most of these are listed under `Socket Options` in the [netlink man page](https://man7.org/linux/man-pages/man7/netlink.7.html). The remaining two (`RX_RING` and `TX_RING`) are in the [same block of `#defines` in the kernel source](https://github.com/torvalds/linux/blob/496659003dac5d08ea292c44fce9dfb36fa34691/tools/include/uapi/linux/netlink.h#L145-L158) and all the code I can find using these passes them to `setsockopt`.

As discussed in https://github.com/bytecodealliance/rustix/issues/1252 these socket options could be exposed in various ways, but for the sake of 1.0, remove the `Protocol` constants that don't belong.

### DIFF
--- a/src/net/types.rs
+++ b/src/net/types.rs
@@ -1110,46 +1110,6 @@ pub mod netlink {
     /// `NETLINK_INET_DIAG`
     #[cfg(linux_kernel)]
     pub const INET_DIAG: Protocol = Protocol(new_raw_protocol(c::NETLINK_INET_DIAG as _));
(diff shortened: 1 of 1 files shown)


sunfishcode:
Thanks!

PR #1339, opened 2025-02-20
Expose socket option SO_SNDBUFFORCE

edmonds:
This is the equivalent of #1174, for allowing privileged users to override the socket send buffer size limit.


sunfishcode:
Thanks!

sunfishcode:

PR #1342, opened 2025-02-21
Rename some runtime functions to `kernel_*`.

sunfishcode:
To help avoid confusion with functions that have the same name as libc functions but have behavior that differs from similarly-named libc functions, rename `runtime::fork` and similar to `runtime::kernel_fork` and similar. Also rename types such as `Sigaction` and similar to `KernelSigaction` and similar. And rename `SIGRTMIN`/`SIGRTMAX` to
`KERNEL_SIGRTMIN`/`KERNEL_SIGRTMAX`.

These make it easier to keep these are the kernel items straight from the similarly-named libc structures.

### DIFF
--- a/src/backend/linux_raw/runtime/syscalls.rs
+++ b/src/backend/linux_raw/runtime/syscalls.rs
@@ -21,7 +21,7 @@ use crate::ffi::CStr;
 use crate::fs::AtFlags;
 use crate::io;
 use crate::pid::{Pid, RawPid};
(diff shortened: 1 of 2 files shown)


sunfishcode:
For example, `rustix::runtime::fork` differs from `libc::fork` in that it doesn't invoke the `pthread_atfork` handlers. That's a significant behavioral difference, so renaming it to `rustix::runtime::kernel_fork` makes it more clear that it isn't the same thing.

Issue #1022, opened 2024-02-23
Misleading name of `ioctl::BadOpcode`


oxalica:
This struct is to make `RawOpcode` in const level to satisfy `CompileTimeOpcode` bound. It is pretty common (and already [used by `rustix` itself](https://github.com/bytecodealliance/rustix/blob/394dd7d3bcc6d076dde411c92905733f8d63f704/src/fs/ioctl.rs#L28)) to plug a generated constants, eg. from `linux_raw_sys`, into the `ioctl::{Getter,Setter,Updater}` interface. This practice uses less magic numbers in Rust code and should not be called "bad".

This issue was also mentioned in https://github.com/bytecodealliance/rustix/pull/861

> The name `Bad` made it seem like something I shouldn't be using, but taking the value by name from `linux_raw_sys` seems less error prone than looking up these components in a C header file and hard-coding them.

I'm not sure about the alternative names here, but here are some candidates:
1. `LiteralOpcode`: but it can accept any const expression, not necessarily literals.
2. `RawCompileTimeOpcode`: really long to write.
3.  `ConstOpcode`: seems like synonym to the trait `CompileTimeOpcode`.
4. `RawOpcode`: name collision with type alias `RawOpcode`.

notgull:
The name was taken from `nix`, which I assumed took it from some Linux kernel documentation. I have a memory of finding this document but I can't seem to find it now when I try to look for it.

sunfishcode:
Rustix has a semver break coming up. Does it make sense to rename `BadOpcode` now?

sunfishcode:
I found the place in Linux that calls them "bad':

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/asm-generic/ioctl.h?id=6d89ead19946181df1e41d38917fddc951dbd95b#n89

sunfishcode:
I've now added [documentation](https://docs.rs/rustix/1.0.0-prerelease.0/rustix/ioctl/struct.BadOpcode.html) about `BadOpcode` which explains that they're not bad to use. Does that address the concern here?

kevinmehall:
[...]

PR #1348, opened 2025-02-23
Simplify the `ioctl` API.

sunfishcode:
As suggested [here], simplify the ioctl API by renaming `RawOpcode` to `Opcode` and using const generics.

[here]: https://github.com/bytecodealliance/rustix/issues/1022#issuecomment-2661284241

### DIFF
--- a/src/ioctl/mod.rs
+++ b/src/ioctl/mod.rs
@@ -185,6 +185,14 @@ pub unsafe trait Ioctl {
 }
 
 /// Const functions for computing opcode values.
+///
+/// Linux's headers define macros such as `_IO`, `_IOR`, `_IOW`, and `_IORW`
+/// for defining ioctl values in a structured way that encode whether they
+/// are reading and/or writing, and other information about the ioctl. The
+/// functions in this module correspond to those macros.
+///
+/// If you're writing a driver and defining your own ioctl numbers, it's
+/// recommended to use these functions to compute them.
 #[cfg(any(linux_kernel, bsd))]
 pub mod opcode {
     use super::*;
@@ -192,6 +200,7 @@ pub mod opcode {
     /// Create a new opcode from a direction, group, number, and size.
     ///
     /// This corresponds to the C macro `_IOC(direction, group, number, size)`
+    #[doc(alias = "_IOC")]
     #[inline]
     pub const fn from_components(
         direction: Direction,
@@ -212,6 +221,7 @@ pub mod opcode {
     /// Create a new opcode from a group, a number, that uses no data.
     ///
     /// This corresponds to the C macro `_IO(group, number)`.
+    #[doc(alias = "_IO")]
     #[inline]
     pub const fn none(group: u8, number: u8) -> Opcode {
         from_components(Direction::None, group, number, 0)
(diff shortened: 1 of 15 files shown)


src/ioctl/mod.rs
```diff
@@ -188,93 +184,66 @@ pub unsafe trait Ioctl {
...
     /// Create a new non-mutating opcode from a group, a number, and the type
     /// of data.
     ///
     /// This corresponds to the C macro `_IO(group, number)` when `T` is zero
     /// sized.
-    #[cfg(any(linux_kernel, bsd))]
     #[inline]
-    pub const fn none<T>(group: u8, number: u8) -> Self {
-        Self::from_components(Direction::None, group, number, mem::size_of::<T>())
+    pub const fn none<T>(group: u8, number: u8) -> Opcode {
```
kevinmehall, line 218: Given that the corresponding Linux `_IO` macro doesn't take an `argtype` like the others, I wonder if `T` could be removed here. I think the size could be hard-coded to 0.

kevinmehall (approved):

src/ioctl/mod.rs
sunfishcode: Makes sense, I've now added a patch for that.

sunfishcode:
I thought some about whether we should do more to discourage "bad" opcodes and promote "new" opcodes.

Right now, rustix's own tree doesn't contain any examples of using `ReadOpcode`, `WriteOpcode`, `ReadWriteOpcode`, or `NoneOpcode`. It's all `BadOpcode`, even when using "new"-stytle opcode numbers. For exmaple, `EXT4_IOC_GETVERSION` in Linux is defined using the "new"-style macros:

```c
#define EXT4_IOC_GETVERSION             _IOR('f', 3, long)
```

but it's [defined] is linux-raw-sys as

pub const EXT4_IOC_GETVERSION: u32 = 2148034051;

[defined]: https://docs.rs/linux-raw-sys/latest/linux_raw_sys/ioctl/constant.EXT4_IOC_GETVERSION.html

The `_IOR` is already baked in, and there's no further need to use `ReadOpcode` or `opcode::read`.

Should we change that? I'm inclined to say no, because I don't think we want to encourage most users to have magic numbers like the `'f` and `3` here spread around their codebase. And Linux itself defines macros like `EXT4_IOC_GETVERSION` to hide these details from most users.

I imagine most users will be like this, using the `BadOpcode` form at the callsites at least. So it's not a bad thing that that becomes the most ergonomic path after the PR here. I've now added more comments advertising the structured opcodes, which I hope suffices.

src/ioctl/mod.rs
```diff
@@ -301,19 +282,19 @@ pub enum Direction {
 pub type IoctlOutput = c::c_int;
 
 /// The type used by the `ioctl` to signify the command.
-pub type RawOpcode = _RawOpcode;
+pub type Opcode = _Opcode;
```
oxalica, line 285: Sorry for being late, but would a newtype struct be better than a type alias? That will be more informative in IDE and can avoids accidental mixing integers in.
Not sure if it interferes with const eval though.

sunfishcode: Newtypes are incompatible with const generics.

PR #1371, opened 2025-03-03
Handle non-NUL-terminated strings in `SocketAddrUnix`.

### LINKED ISSUE #1316: Panic on Unix-domain sockets with length 108
On Linux, Unix-domain sockets with a path of length 108 don't get a terminator within `sun_path`:

https://man7.org/linux/man-pages/man7/unix.7.html#BUGS

Rustix currently panics on such paths:

https://github.com/bytecodealliance/rustix/pull/1004#discussion_r1947915868

sunfishcode:
Unix-domain socket address can be long enough that the NUL terminator does not fit. Handle this case by making `path()` return a `Cow<CStr>` and adding a NUL terminator as needed.

Also add a `path_bytes()` function for returning the raw bytes.

Fixes #1316.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,7 +55,7 @@ jobs:
         cargo update --package=regex --precise=1.9.0
         cargo update --package=half --precise=2.2.1
         cargo update --package=flate2 --precise=1.0.35
-        cargo update --package=textwrap --precise=v0.16.1
+        cargo update --package=textwrap --precise=0.16.1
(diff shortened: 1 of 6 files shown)


Kijewski:
Would it maybe be possible to make the array one element longer instead?

sunfishcode:
We define `SocketAddrUnix` to have the same layout as `sockaddr_un`, so we can't easily change the size of the `sun_path` field.

PR #1372, opened 2025-03-03
Allocated `SocketAddrUnix::path()` with correct length

Kijewski:
Using `.to_owned()` + `.push()` will cause a reallocation, because the initially allocated array with be one byte too short.

We can use `CString::from_vec_with_nul_unchecked()` because it is a known invariant that the input does not contain any `NUL`s, not even the terminating `NUL`.

---

@sunfishcode, I hope it's okay that I post a PR to a PR. It was easier to implement the change in a commit than to add multiple suggestions in <https://github.com/bytecodealliance/rustix/pull/1371/>.

### DIFF
--- a/src/backend/libc/net/addr.rs
+++ b/src/backend/libc/net/addr.rs
@@ -14,10 +14,7 @@ use {
     core::slice,
 };
 #[cfg(all(unix, feature = "alloc"))]
-use {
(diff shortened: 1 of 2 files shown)


sunfishcode:
Cool, thanks!

PR #1388, opened 2025-03-06
Add convenient is_ functions to FileType

rusty-snake:


src/fs/constants.rs
```diff
@@ -5,6 +5,50 @@ use crate::backend;
...
+
+    /// Returns `true` if this `FileType` is a block device.
+    pub fn is_block_device(self) -> bool {
+        self == FileType::BlockDevice
+    }
+
+    /// Returns `true` if this `FileType` is unknown.
+    pub fn is_unknown(self) -> bool {
+        self == FileType::Unknown
+    }
```
sunfishcode, line 49: Is `is_unknown` useful? Rust's `FileTypeExt` [lacks it](https://doc.rust-lang.org/stable/std/os/unix/fs/trait.FileTypeExt.html#tymethod.is_char_device).

rusty-snake: Good point. I carry these functions as an extension trait for one or two years now and added the `is_unkown` only for "completeness" in this PR.

I've no problem without it.

sunfishcode: Makes sense; unless there's a specific need for it, my sense here is it's best to follow std.

PR #1389, opened 2025-03-06
Disable use-libc-auxv by default, and document cargo features

### LINKED ISSUE #1387: Document `linux_*` features
The `linux_*` features are not documented nor do they have a "don't use" comment in Cargo.toml.

sunfishcode:
Disable use-libc-auxv by default, as we can now use `PR_GET_AUXV` on most systems today, and fall back to "/proc/self/auxv".

And add documentation for the main user-facing Cargo features to README.md.

Fixes #1387.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,12 +106,8 @@ targets = [
 
 [features]
 
-# By default, use `std` and use libc for aux values.
-#
-# It turns out to be bizarrely awkward to obtain the aux values reliably and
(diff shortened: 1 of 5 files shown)


joshtriplett:
@sunfishcode This makes me so happy to see! I'm glad that `PR_GET_AUXV` has managed to work out this well.

sunfishcode:
@joshtriplett Thanks for `PR_GET_AUXV`; it has indeed worked out great!

PR #1397, opened 2025-03-07
Add `index_to_name_inlined`

### LINKED ISSUE #1393: `rustix::net::netdevice::index_to_name` is returning `String`
Hi, on the Linux and BSD platform, the max interface name length is `16`, which is small enough to be inlined instead of a heap-allocated `String`.

If we change the returning type to an inlined array or somehow a wrapper over `[u8; 16]`, we can remove the `alloc` requirement.

Image

al8n:
Close #1393.


sunfishcode:
Thanks!

PR #1406, opened 2025-03-09
Restore `Signal::from_raw` as `Signal::from_named_raw`.

sunfishcode:
Rust 1.0 moved `Signal::from_raw` into a [separate crate], since properly supporting `SIGRTMIN+n` signals requires querying libc. But it's still useful to have a version of this function in rustix which is just documented to not support `SIGRTMIN+n` signals, so re-introduce it, with a different name and with appropriatet documentation.

[separate crate]: https://crates.io/crates/rustix-libc-wrappers


sunfishcode:
This is restoring the functionality that previously was provided by [`Signal::from_raw`](https://docs.rs/rustix/0.38.44/rustix/process/enum.Signal.html#method.from_raw), so I've now added a CHANGES.md entry for it.

PR #1409, opened 2025-03-11
Add support for SO_TXTIME

colinmarc:
This is an undocumented linux feature that lets you time outgoing packets.

This depends on https://github.com/sunfishcode/linux-raw-sys/pull/153.

### DIFF
--- a/src/clockid.rs
+++ b/src/clockid.rs
@@ -1,5 +1,6 @@
 use crate::backend::c;
 use crate::fd::BorrowedFd;
+use crate::io;
 
 /// `CLOCK_*` constants for use with [`clock_gettime`].
 ///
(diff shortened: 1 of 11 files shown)


src/net/sockopt.rs
```diff
@@ -1522,6 +1526,20 @@ pub fn socket_peercred<Fd: AsFd>(fd: Fd) -> io::Result<super::UCred> {
     backend::net::sockopt::socket_peercred(fd.as_fd())
 }
 
+/// `getsockopt(fd, SOL_SOCKET, SO_TXTIME)` — Get transmit timing settings.
+#[cfg(all(target_os = "linux", feature = "time"))]
```
colinmarc, line 1530: should this be `#[cfg(linux_kernel)]`? I don't get the distinction.

sunfishcode: `linux_kernel` is equivalent to `any(target_os = "linux", target_os = "android"))`. If this feature compiles on Android, then we can enable it there too.

colinmarc:
I just realized this needs the `SCM_TXTIME` part, to be useful. I'll add that in the next few days. Marking as a draft for now.

src/clockid.rs
```diff
@@ -94,6 +95,51 @@ pub enum ClockId {
     BoottimeAlarm = bitcast!(c::CLOCK_BOOTTIME_ALARM),
 }
 
+#[cfg(not(any(apple, target_os = "wasi")))]
```
colinmarc, line 98: I thought this would be straightforward, but it's really not. libc has `clockid_t`, while linux-raw-sys has `__kernel_clockid_t`. Some platforms have neither. What's the best way to proceed with this? I need it for `get_txtime`, since the kernel returns the configured clock.

colinmarc:
Added `SCM_TXTIME` and a e2e test.

sunfishcode:
linux-raw-sys 0.9.3 is now released, with https://github.com/sunfishcode/linux-raw-sys/pull/153.

colinmarc:
> linux-raw-sys 0.9.3 is now released, with [sunfishcode/linux-raw-sys#153](https://github.com/sunfishcode/linux-raw-sys/pull/153).

Great, thanks :) I rebased now and removed the hardcoded declarations.

sunfishcode:
Thanks!

PR #1415, opened 2025-03-12
Experiment with changing `Buffer::parts_mut` to `Buffer::as_mut_ptr`.

sunfishcode:
Based on an idea suggested [here], experiment with changing from a `parts_mut` that returns `(*mut T, usize)` to a `as_mut_ptr` which returns a `*mut [MaybeUninit<T>]`.

[here]: https://hachyderm.io/deck/@wffl@im-in.space/114145162248844972

### DIFF
--- a/src/backend/libc/event/syscalls.rs
+++ b/src/backend/libc/event/syscalls.rs
@@ -4,8 +4,6 @@ use crate::backend::c;
 #[cfg(any(linux_kernel, solarish, target_os = "redox"))]
 use crate::backend::conv::ret;
 use crate::backend::conv::ret_c_int;
-#[cfg(any(linux_kernel, target_os = "illumos", target_os = "redox"))]
-use crate::backend::conv::ret_u32;
 #[cfg(solarish)]
 use crate::event::port::Event;
 #[cfg(any(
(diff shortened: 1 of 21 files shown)


sunfishcode:
It appears `pointer::len` requires Rust 1.79.0. So we may not make this change in rustix for now, but we should still think about it for `Buffer` in general.

joshtriplett:
In theory, `[MaybeUninit<T>]` encapsulates all the "might be uninitialized", so could this use `&mut` rather than *mut`?

sunfishcode:
> In theory, `[MaybeUninit<T>]` encapsulates all the "might be uninitialized", so could this use `&mut` rather than *mut`?

The subtlety is that if the buffer is a `&mut [T]`, a `&mut MaybeUninit<T>` view of it could let users could store `MaybeUninit::uninit()` into it.

joshtriplett:
> > In theory, `[MaybeUninit<T>]` encapsulates all the "might be uninitialized", so could this use `&mut` rather than *mut`?
> 
> The subtlety is that if the buffer is a `&mut [T]`, a `&mut MaybeUninit<T>` view of it could let users could store `MaybeUninit::uninit()` into it.

Ah, right, I *did* remember that part but forgot about the case of passing in initialized memory (`&mut [T]`) in the first place. :)

sunfishcode:
I like this idea in theory, but in practice, it ends up making the code more verbose and repetitive. I'm now prototyping `Buffer` type changes [in a standalone repo](https://docs.rs/buffer-trait/latest/buffer_trait/trait.Buffer.html), where it has simplified the code to have separate getters for the pointer and length, because it's sometimes useful to be able to get the length without having a `&mut self`. So I'll close this for now.

Another idea here is that the subtle problem above with `&mut [MaybeUninit<T>]` could be avoided if we could instead use `&[MaybeUninitCell<T>]`, where `MaybeUninitCell` would be a wrapper around a `MaybeUninit` which has an interor-mutability way to store initialized values. That way, users with a shared reference could initialize but not deinitialize the elements. That would fix the subtlety above. However, in practice, that wouldn't perform the bookkeeping to track how many elements have been initialized, so the resulting API still wouldn't be very ergonomic.

PR #1426, opened 2025-03-26
Implement `socket_bind_device` and `set_socket_bind_device`

al8n:
Hi, this PR adds `socket_bind_device` and `set_socket_bind_device` methods.

Besides, as the interface name can only be `[u8; 16]`, I think adding an `InterfaceBuffer` is better, so that we can remove the `alloc` feature, which has been mentioned in #1393 and #1397 previously.


sunfishcode:
It looks like this has a build error in the tests on non-Linux platforms:

```
error[E0433]: failed to resolve: could not find `netdevice` in `net`
   --> tests/net/sockopt.rs:642:29
    |
642 |     let name = rustix::net::netdevice::index_to_name(&fd, loopback_index).unwrap();
    |                             ^^^^^^^^^ could not find `netdevice` in `net`
    |
```
Can we enclose those tests in `#[cfg(linux_kernel)]`?

Sawangg:
I'm interested in this feature. @al8n are you still working on it?

al8n:
> I'm interested in this feature. @al8n are you still working on it?

I almost forget this PR. But I will update the code this weekend.

PR #1435, opened 2025-04-05
Fix rustix 0.38.42 build on riscv32

kraj:
RISCV32 libc does not have these defines so use them from linux_raw_sys crate

ioctl() signare does not match between musl and glibc ioctl is called the glibc way with ioctl(int, unsigned long int, ...), musl/POSIX has ioctl(int, int, ...)

so address this as well

### DIFF
--- a/src/fs/ioctl.rs
+++ b/src/fs/ioctl.rs
@@ -11,6 +11,14 @@ use {
 
 use bitflags::bitflags;
 
+/// The datatype used for the ioctl number
+#[cfg(any(target_os = "android", target_env = "musl"))]
+#[doc(hidden)]
+pub type IoctlNumType = c::c_int;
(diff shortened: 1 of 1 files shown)


sunfishcode:
Thanks for the PR! However, I think in this case it's better to just fix the upstream libc crate to define these constants on riscv32. I've now submitted https://github.com/rust-lang/libc/pull/4382 to do this. When I build rustix with a libc with that patch on riscv32gc-unknown-linux-gnu, it builds without any other modifications.

sunfishcode:
https://github.com/rust-lang/libc/pull/4382 is now landed, backported, and released in libc 0.2.172. I believe this fixes the underlying issue here; please file new issues if there are any further issues!

PR #1436, opened 2025-04-06
Rework Capability types

### LINKED ISSUE #1434: Capability vs. CapabilityFlags or How to raise all inheritable capabilities in the ambient set.
Capabilities (as well as namespaces, we discussed elsewhere) are bitflags in C/Linux. Some system calls do take/return an actual set while others take/return only a single bit. rustix uses bitflags types for the first and enum types for the later (increased typesafety).
For capabilities in rustix there is a `Capability` enum with opinionated names used for bounding and ambient sets where the underlying system calls take only one capability and a `CapabilityFlags` bitflags with original names used for inheritable/effective/permitted sets where the underlying system calls operate on sets.
I understand that rustix does not want to abstract these one/set semantic away and stick closer to the actual system calls. However the types should be unified.
What I mean with this is that 1. the names should be the same (e.g. `CHOWN` vs. `ChangeOwnership`) and 2. there must be easy convert functions between thoses two:
- [ ] `Capability` -> `CapabilityFlags`, easy you just need a new `CapabilityFlags` with just one bit set.
- [ ] `CapabilityFlags` -> `Capability`, likely as an iterator?

rusty-snake:
- Rename CapabilityFlags to CapabilitySet
- Deprecated Capability, use CapabilitySet instead.
- Add sealed CompatCapability trait.

Closes #1434

### DIFF
--- a/src/thread/libcap.rs
+++ b/src/thread/libcap.rs
@@ -8,18 +8,22 @@ use crate::{backend, io};
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct CapabilitySets {
     /// `__user_cap_data_struct.effective`
-    pub effective: CapabilityFlags,
+    pub effective: CapabilitySet,
     /// `__user_cap_data_struct.permitted`
-    pub permitted: CapabilityFlags,
+    pub permitted: CapabilitySet,
     /// `__user_cap_data_struct.inheritable`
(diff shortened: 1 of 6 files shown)


rusty-snake:
This one is ready for review. There should be a follow-up to improve cap_last_cap stuff in a nice way.

sunfishcode:
Thanks!

PR #1442, opened 2025-04-09
Always cast `pread`, `lseek` etc. offsets to `off_t`.

### LINKED ISSUE #1432: Suggesting additional use of `c::off_t` inside `seek`/`fallocate`/`pread`/`pwrite`
## Summary

I think certain `rustix` functions under `backend/` should use `c::off_t` instead of `i64` when calling certain `libc` functions.

## Problem

While building a custom target that has a 32-bit `off_t` (and a 32-bit `long`), I noticed compilation failures including this one:

```
error[E0308]: mismatched types
    --> vendor/rustix/src/backend/libc/fs/syscalls.rs:1305:63
     |
1305 |     let offset = unsafe { ret_off_t(c::lseek(borrowed_fd(fd), offset, whence))? };
     |                                     --------                  ^^^^^^ expected `i32`, found `i64`
     |                                     |
     |                                     arguments to this function are incorrect
     |
note: function defined here
    --> /home/kulp/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.171/src/unix/mod.rs:1000:12
     |
1000 |     pub fn lseek(fd: c_int, offset: off_t, whence: c_int) -> off_t;
     |            ^^^^^
help: you can convert an `i64` to an `i32` and panic if the converted value doesn't fit
     |
1305 |     let offset = unsafe { ret_off_t(c::lseek(borrowed_fd(fd), offset.try_into().unwrap(), whence))? };
     |                                                                     ++++++++++++++++++++
```

## Details

This issue concerns at least four functions:
- `seek`
- `fallocate`
- `pread`
- `pwrite`

Take `seek` as an example, which delegates to `lseek`. [`lseek` per POSIX.1](https://pubs.opengroup.org/onlinepubs/9799919799/functions/lseek.html) has the prototype
```c
off_t lseek(int fd, off_t offset, int whence);
```
and [the `libc` crate uses `off_t` for `offset` as expected](https://github.com/rust-lang/libc/blob/main/src/unix/mod.rs#L996), whereas the `seek` in `rustix` calls it with either `i64`:

https://github.com/bytecodealliance/rustix/blob/cb01fbe4660844b67fdd4eee2a5f769518f6a655/src/backend/libc/fs/syscalls.rs#L1385-L1390

or else `i32`, if a target-specific workaround is in play:

https://github.com/bytecodealliance/rustix/blob/cb01fbe4660844b67fdd4eee2a5f769518f6a655/src/backend/libc/fs/syscalls.rs#L1407-L1409

which was [introduced](https://github.com/bytecodealliance/rustix/commit/23af8a8c4ca37f6fdf602a26a67d3e32a1f64e80#diff-4ff7b5821ddc768048f4e5ede4eee035e568c6f0dac1934a693fc00ff5c147edL1172-L1187) in 23af8a8c4ca37f6fdf602a26a67d3e32a1f64e80. Before that commit, `c::off_t` was named in a type ascription:

https://github.com/bytecodealliance/rustix/blob/609ec9b393a735e77fc3c0932c6291dcd20bd327/src/backend/libc/fs/syscalls.rs#L1171-L1177

although I wonder if that might have compiled only when `c::off_t` actually matched `i64`.

# Workaround

I was able to work around this by converting `offset` universally to `c::off_t` (albeit against a vendored version of v0.38.44), like this:
```diff
diff --git a/rustix/src/backend/libc/fs/syscalls.rs b/rustix/src/backend/libc/fs/syscalls.rs
index 2d2f4c6..0fdb3ad 100644
--- a/rustix/src/backend/libc/fs/syscalls.rs
+++ b/rustix/src/backend/libc/fs/syscalls.rs
@@ -1298,6 +1298,8 @@ pub(crate) fn seek(fd: BorrowedFd<'_>, pos: SeekFrom) -> io::Result<u64> {
         SeekFrom::Hole(offset) => (c::SEEK_HOLE, offset),
     };
 
+    let offset: c::off_t = offset.try_into().map_err(|_| io::Errno::OVERFLOW)?;
+
     // ESP-IDF and Vita don't support 64-bit offsets.
     #[cfg(any(target_os = "espidf", target_os = "vita"))]
     let offset: i32 = offset.try_into().map_err(|_| io::Errno::OVERFLOW)?;
```

It also seems possible that my workaround might be able to **replace** the workaround for ESP-IDF and Vita.

## Questions

1. Can someone correct my understanding?
2. Would a PR be welcome to add the conversion as I expressed it, or is there a better way?

sunfishcode:
When calling libc functions that take `off_t` arguments, always cast to `off_t`. On platforms where this cast is a no-op, these casts should be optimized away. This eliminates the need for explicit `cfg`s for platforms which need the casts.

Fixes #1432.

### DIFF
--- a/src/backend/libc/fs/syscalls.rs
+++ b/src/backend/libc/fs/syscalls.rs
@@ -1404,9 +1404,8 @@ pub(crate) fn seek(fd: BorrowedFd<'_>, pos: SeekFrom) -> io::Result<u64> {
         }
     };
 
-    // ESP-IDF and Vita don't support 64-bit offsets.
-    #[cfg(any(target_os = "espidf", target_os = "vita"))]
-    let offset: i32 = offset.try_into().map_err(|_| io::Errno::OVERFLOW)?;
+    // ESP-IDF and Vita don't support 64-bit offsets, for example.
+    let offset = offset.try_into().map_err(|_| io::Errno::OVERFLOW)?;
 
     let offset = unsafe { ret_off_t(c::lseek(borrowed_fd(fd), offset, whence))? };
     Ok(offset as u64)
@@ -1734,9 +1733,12 @@ pub(crate) fn fallocate(
     offset: u64,
     len: u64,
 ) -> io::Result<()> {
-    // Silently cast; we'll get `EINVAL` if the value is negative.
+    // Silently cast to `i64`; we'll get `EINVAL` if the value is negative.
     let offset = offset as i64;
     let len = len as i64;
+
+    // ESP-IDF and Vita don't support 64-bit offsets, for example.
+    let offset = offset.try_into().map_err(|_| io::Errno::OVERFLOW)?;
 
     #[cfg(any(linux_kernel, target_os = "fuchsia"))]
     unsafe {
--- a/src/backend/libc/fs/syscalls.rs
+++ b/src/backend/libc/fs/syscalls.rs
@@ -1739,6 +1739,7 @@ pub(crate) fn fallocate(
 
     // ESP-IDF and Vita don't support 64-bit offsets, for example.
     let offset = offset.try_into().map_err(|_| io::Errno::OVERFLOW)?;
+    let len = len.try_into().map_err(|_| io::Errno::OVERFLOW)?;
 
     #[cfg(any(linux_kernel, target_os = "fuchsia"))]
     unsafe {
--- a/src/backend/libc/io/syscalls.rs
+++ b/src/backend/libc/io/syscalls.rs
@@ -52,9 +52,8 @@ pub(crate) unsafe fn pread(
     // Silently cast; we'll get `EINVAL` if the value is negative.
     let offset = offset as i64;
 
-    // ESP-IDF and Vita don't support 64-bit offsets.
-    #[cfg(any(target_os = "espidf", target_os = "vita"))]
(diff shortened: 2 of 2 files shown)


src/backend/libc/fs/syscalls.rs
```diff
@@ -1734,10 +1733,13 @@ pub(crate) fn fallocate(
...
     len: u64,
 ) -> io::Result<()> {
-    // Silently cast; we'll get `EINVAL` if the value is negative.
+    // Silently cast to `i64`; we'll get `EINVAL` if the value is negative.
     let offset = offset as i64;
     let len = len as i64;
 
+    // ESP-IDF and Vita don't support 64-bit offsets, for example.
+    let offset = offset.try_into().map_err(|_| io::Errno::OVERFLOW)?;
+
```
kulp, line 1742: A cast is needed for `len` here too.

kulp: (I failed to point this out in #1432; it is the only non-`offset` case that I personally have run into, but there may be others I missed.)

kulp:
Due to unrelated changes between v0.38.44 and v1.0.5, I cannot right now directly test the branch in the current MR, but I did successfully compile my project after cherry-picking 5dbcf91000825800d6aefd81c07dac143a46dfb0 plus an extra cast for `len` in `fallocate`. That is good enough for me personally.

Thanks!

Issue #1433, opened 2025-04-05
UB if noreturn syscall actually returns


rusty-snake:
https://github.com/bytecodealliance/rustix/blob/cb01fbe4660844b67fdd4eee2a5f769518f6a655/src/backend/linux_raw/arch/x86_64.rs#L55-L63

https://github.com/bytecodealliance/rustix/blob/cb01fbe4660844b67fdd4eee2a5f769518f6a655/src/backend/linux_raw/runtime/syscalls.rs#L143-L146

***Every syscall may return** with any value* because of seccomp-bpf.

This includes pure syscalls like `getuid` that may return the current uid as 4294967295_u32 (`-EPERM`) or `exit`/`exit_group` returning. While such seccomp-bpf filters will have huge compatibility problems and are unlikely in real-world, they must be handled in a safe way.

In the case of `exit`/`exit_group` there are two safe possibilities that maintain `-> !` to the caller:

1. Infinity loop 
   ```rust
   // Note that this is a pseudo example. If the `sys_exit` function is `-> !`,
   // Rust is free to remove the loop because returning from `-> !` is UB.
   // The loop must be implemented in assembly with a backward jump
   // if the `asm!` is marked `noreturn` as return from the `noreturn`-`asm!`
   // is already UB.
   loop {
       sys_exit(code)
   }
   ```
2. Segfault, i.e. `ud2` (x86-64) after the `syscall`.

sunfishcode:
rustix also trusts that when `read` returns 4, that exactly 4 bytes have been written into the buffer. If someone rigs a seccomp-bpf script to make `read` return a greater number, `rustix` will execute UB when it forms a `&[u8]` containing uninitialized bytes.

rustix also trusts that if `openat` returns a non-negative number, it's a new unique fd that upholds the assumptions of I/O safety. rustix also trusts that `mmap` returns the pointer is what it says it is rather than a pointer to something else (yes, `mmap` is unsafe, but the safety comment trusts that this isn't something users need to worry about).

And these issues affect the libc backend as well as the linux_raw backend. They even affect typical Rust idioms using eg. `libc::read` or even `std::io::Read::read` and not using rustix at all.

It doesn't seem practical to defend against arbitrary theoretical seccomp_bpf scripts, in rustix, or in the Rust ecosystem as a whole.

rusty-snake:
Yeah, I see your point.

[...]

PR #1443, opened 2025-04-10
Document that negative pids aren't UB, but may cause unexpected behavior.

sunfishcode:
As discussed in #1433, allow `Pid` values to be negative without invoking UB, because they can occur on systems configured with seccomp-bpf to make syscalls like `getpid` fail. However, also document that while this doesn't invoke UB, it isn't useful.

### DIFF
--- a/src/pid.rs
+++ b/src/pid.rs
@@ -27,7 +27,7 @@ impl Pid {
 
     /// Converts a `RawPid` into a `Pid`.
     ///
-    /// Returns `Some` for positive `RawPid`s. Otherwise, returns `None`.
+    /// Returns `Some` for positive values, and `None` for zero values.
     ///
     /// This is safe because a `Pid` is a number without any guarantees for the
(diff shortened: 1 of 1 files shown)


src/pid.rs
```diff
@@ -27,17 +27,21 @@ impl Pid {
...
     /// non-child processes, please consider other mechanisms like [pidfd] on
     /// Linux.
     ///
+    /// Passing a negative number doesn't invoke undefined behavior, but it
+    /// may cause unexpected behavior.
+    ///
     /// [pidfd]: https://man7.org/linux/man-pages/man2/pidfd_open.2.html
     #[inline]
     pub const fn from_raw(raw: RawPid) -> Option<Self> {
+        debug_assert!(raw > 0);
```
rusty-snake, line 44: This makes `from_raw` panic in debug builds with a zero value but return `None` is release builds.

PR #1449, opened 2025-04-11
Implement fmt::Display for Pid, Uid, Gid

### LINKED ISSUE #1439: impl core::fmt::Display for Pid, Uid, Gid?
What do you think about implementing `core::fmt::Display` for `Pid`, `Uid`, `Gid`? 

Display would be guaranteed to always print the integer representation.

Do you know other types in rustix where this would make sense.

rusty-snake:
Closes #1439

### DIFF
--- a/src/pid.rs
+++ b/src/pid.rs
@@ -2,7 +2,7 @@
 
 #![allow(unsafe_code)]
 
-use core::num::NonZeroI32;
+use core::{fmt, num::NonZeroI32};
 
 /// A process identifier as a raw integer.
 pub type RawPid = i32;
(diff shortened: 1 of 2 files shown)


evilpie:
It might be nice to implement more formatters like UpperHex. e.g. https://doc.rust-lang.org/src/core/num/nonzero.rs.html#136-153

src/pid.rs
```diff
@@ -87,6 +87,12 @@ impl Pid {
     }
 }
 
+impl fmt::Display for Pid {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
```
evilpie, line 92:
```suggestion
        self.0.fmt(f)
```

This makes some of the custom alignments etc. work.

src/ugid.rs
```diff
@@ -86,6 +88,78 @@ impl Gid {
...
 }
 
+impl fmt::Display for Uid {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+impl fmt::Binary for Uid {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
```
sunfishcode, line 98: Which `fmt` does this call? Does it call `Binary::fmt` or does it forward to `Display::fmt`?

And similar for all the other `fmt::*` impls.

rusty-snake: Always the right one

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0748876dbb30e5dcc211d8d65ce1870b

sunfishcode:
Thanks!

PR #1466, opened 2025-05-22
Add `ip_mtu_discover` and `ipv6_mtu_discover` socket options

pekkarr:
Add socket options to get and set the Path MTU Discovery setting for a socket.


sunfishcode:
Thanks!

PR #1471, opened 2025-05-30
libc: add MS_NOSYMFOLLOW support

cyphar:
Linux platforms that are forced to use the libc backend cannot access
MS_NOSYMFOLLOW. The upstream libc crate doesn't expose MS_NOSYMFOLLOW,
but on Linux we can just remap linux-raw-sys.

FreeBSD also has nosymfollow support in the form of MNT_NOSYMFOLLOW, but
rustix::mount is not provided for FreeBSD.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>


sunfishcode:
Thanks!

PR #1472, opened 2025-06-04
Be

WorksButNotTested:
* Configure `rustix` to use the `linux_raw` backend on `armeb-unknown-linux-gnueabi` and `aarch64_be-unknown-linux-gnu` targets. (Both of these are tier 3 rust platforms and hence no public binaries are provided).
* Since `rustix` uses `tempfile` for its unit test, which in turn uses `rustix` as a dependency we have a cyclic dependency. Change the `Cargo.toml` so that the version of `rustix` used to satisfy this `tempfile` dependency is the working copy.


WorksButNotTested:
@sunfishcode Grateful for your thoughts on this.

Cargo.toml
```diff
@@ -64,8 +64,14 @@ version = "0.3.10"
 package = "errno"
 default-features = false
 
+# Rustix uses tempfile for its tests, which in turn uses rustix! We want to use
+# the same consistent version. We therefore need to use a version of tempfile
+# for which its rustix dependency is a semver match for our current version.
+[patch.crates-io]
+rustix = { path = "." }
```
bjorn3, line 71: This will make it impossible to do a semver incompatible release of rustix as there is no tempfile version that has a semver compatible dependency on rustix yet.

src/backend/linux_raw/c.rs
```diff
@@ -286,11 +286,11 @@ pub(crate) const SIGSYS: c_int = linux_raw_sys::general::SIGSYS as _;
 ))]
 pub(crate) const SIGEMT: c_int = linux_raw_sys::general::SIGEMT as _;
 
-#[cfg(feature = "stdio")]
+#[cfg(any(test, feature = "stdio"))]
```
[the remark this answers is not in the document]
WorksButNotTested: These constants are used in the tests [here](https://github.com/bytecodealliance/rustix/blob/0a6cb07f8270aab17e534853fb25de1f7a0c9b33/src/fs/special.rs#L60-L61) and [here](https://github.com/bytecodealliance/rustix/blob/0a6cb07f8270aab17e534853fb25de1f7a0c9b33/src/io_uring/mod.rs#L2203). Not related to the purpose of my change, but trying to get the CI into life to check for any regressions.

WorksButNotTested:
@bjorn3 Thanks for the feedback. Just let me know how best to proceed and I'm happy to make any changes.

Cargo.toml
sunfishcode: tempfile's use of rustix is not part of rustix's testsuite. We don't need to make tempfile's version of rustix match the version of rustix using tempfile for tests. They can differ and it's fine.

src/backend/linux_raw/c.rs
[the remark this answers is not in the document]
sunfishcode: Instead of adding `cfg(test)` here, please either put the test code that uses `STDIN_FILENO` under `cfg(feature = "stdio")`, or define `STDIN_FILENO` here also when `cfg(feature = "fs")`.

Cargo.toml
sunfishcode: So to be clear, I think it's best to omit this part of the patch.

sunfishcode:
How much testing have you done with the linux_raw backend on these targets? There are some places in the Linux syscall ABI which are sensitive to endianness, such as when 32-bit values are split into two halves, so it'd be good to have some assurance that this code is working before enabling it on these targets.

WorksButNotTested:
Ideally I'd like to add CI for these targets, the only thing is that it requires:
* An `armbe-linux-gnueabi-` and `aarch64_be-linux-gnu-` cross compiler (since we would need a linker) and neither the standard linux distros, nor ARM themselves provide a pre-built one ready to use.
* Either:
  * A build of `qemu-system` supporting the relevant targets with accompanying `initrd` and `kernel`.
  *  A build of `qemu-user` supporting the relevant targets.

I have already made most of these for some other projects, but wondered what if anything would be the best approach.

* There are containers with cross-compilers as part of `frida`'s `docker-images` repo for [armbe](https://github.com/frida/docker-images/blob/main/x-tools/Dockerfile.linux-armbe8) and [aarch64_be](https://github.com/frida/docker-images/blob/main/x-tools/Dockerfile.linux-arm64be). These could either be used as is, or duplicated into the repo.
* There is also a container with `qemu-system`, `kernel` and `initrd` [here](https://github.com/frida/docker-images/blob/main/x-tools/Dockerfile.linux-be-target). Again this could be either used as is, or duplicated. The workflow for running this sort of setup is a little clunky and would look a bit like [this](https://github.com/frida/frida-gum/blob/90124fb78e707d95209e3d81c229f9f2c6dc2440/.github/workflows/ci.yml#L172).
* Building `qemu` from source to support `qemu-armbe` and `qemu-aarch64_be` isn't too hard, so again a container could be made for that, or it could be built and installed as part of a CI job (though likely some form of caching would be useful to speed up this job).

The consideration is in deciding whether to use resources from other projects or whether to replicate the same functionality to give more control. Then of course there is the additional cost in terms of resources and time for the CI.

I'm happy to go away and start working on a solution if you think it would be beneficial, but would be grateful if you have some advice on which direction to take.

sunfishcode:
I don't want to enable linux_raw on architectures that it hasn't been tested on, and I'm not eager to maintain new complex testing configurations.

Would it make sense to enable linux_raw for these architectures under the `RUSTFLAGS=--cfg=rustix_use_experimental_asm` flag, similar to what powerpc, s390x, and mips use? That way, end users can explicitly opt into it if they wish to. See the `rustix_use_experimental_asm` code in build.rs for details.

WorksButNotTested:
Sure. That makes sense. They aren't the most common of architectures to be fair. I'll introduce a new flag, something like "use_linux_raw_on_experimental_arch".

WorksButNotTested:
It seemed the reasons for having the flags differed although the result was the same (defaulting to the libc bindings). In the case of PPC, it was because rust hasn't stabilised inline asm. In the case of big endian arm it was because the overhead of adding sufficient CI was to great considering its not a very popular architecture. Happy to change if you'd rather?

PR #1474, opened 2025-06-11
Add `socket_peerpidfd`

ids1024:
Supported on Linux since 6.5.

I see `socket_peercred` is only provided on Linux, while OpenBSD (https://man.openbsd.org/getsockopt.2) has some version of `SO_PEERCRED`, and FreeBSD (https://man.freebsd.org/cgi/man.cgi?unix(4)) has a `LOCAL_PEERCRED` (libwayland for instance has code to use those on their respective platforms). These APIs do seem to be slightly different though. In any case, `SO_PEERPIDFD` is definitely Linux-specific for the time being.

### DIFF
--- a/tests/event/select.rs
+++ b/tests/event/select.rs
@@ -88,6 +88,7 @@ fn test_select_with_pipes() {
 #[serial] // for `setrlimit` usage
 fn test_select_with_great_fds() {
     use core::cmp::max;
+    use rustix::fd::{FromRawFd as _, OwnedFd};
     use rustix::io::{read, write};
     use rustix::pipe::pipe;
     use rustix::process::{getrlimit, setrlimit, Resource};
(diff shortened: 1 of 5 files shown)


tests/event/select.rs
```diff
@@ -277,6 +278,7 @@ fn test_select_with_sockets() {
 #[test]
 #[serial] // for `setrlimit` usage, and `crate::init`
 fn test_select_with_maxfd_sockets() {
+    use rustix::fd::{FromRawFd as _, OwnedFd};
```
sunfishcode, line 281: I'm curious why these lines were needed in the tests, that weren't otherwise changed.

sunfishcode (commented):
This looks good to me, with one quick question:

sunfishcode:
Those imports are unfortunately evoking warnings, which are currently configured to be errors in CI. Could you add conditionals to the imports so that they're not imported if they're not needed?

zeenix:
@ids1024 Thanks so much for doing this. Do you intend to finish this PR?

ids1024:
Or well, one of the errors was related to this change. Fixed that, and added a commit for another CI error. So now everything is passing.

ids1024:
Rebased again to get CI fixes from `master`.

Ping @sunfishcode

PR #1478, opened 2025-06-20
[RFC][wip] rustix: on Linux, support a build without linux-raw-sys

gurchetansingh:
Rustix support was added to Mesa3D recently:

https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/35210

but it only works with the linux-raw-sys backend.  On Linux, even the libc backend, there is a dependency on linux-raw-sys for certain ABI definitions.

When updating rustix in Android to support this, I realized only the libc backend is supported in Android.  This is because Android as a platform wants to discourage raw syscalls, and to be able to globally update libc.

Features like fdtrack would not work with the linux-raw-sys backend:

https://android.googlesource.com/platform/bionic/+/HEAD/docs/fdtrack.md

Given Android wants to incentivize use of libc only, this MR introduces the concept of "linux-raw-dep".  This means that linux-raw-sys is present as a dependency for the user.  This will be on by default even for the libc backend, but will be turned off for Android.  This allows the libc backend to compile on Linux, without linux-raw-sys.

This will also help others who don't need all the features rustix offers on Linux, so they have one less dependency to maintain.

Tested via:

cargo build --features=use-libc,net,std,alloc,event,fs,mm,param,pipe

These are the same set of features needed by Mesa3D.


gurchetansingh:
I don't think we care about the freebsd-15 runner?  It passes on freebsd-14, and the other changes were merged with checks not passing on freebsd-16:

https://github.com/bytecodealliance/rustix/pull/1460/checks

Either way, changes are ready for review.  Thank you!

sunfishcode:
I appologize for being slow here. Thanks for putting this together! I think this makes sense. And the FreeBSD CI failure is likely unrelated.

PR #1437, opened 2025-04-06
Rework namespace functions and types

rusty-snake:
Successor of #1402

- Add set_namespaces function.
- Deprecate move_into_link_name_space and move_into_thread_name_spaces functions.
- Add NamespaceType.
- Deprecate ThreadNameSpaceType and LinkNameSpaceType.
- Add nsfs ioctls.
- Add NoArgGetter ioctl pattern.

PR #1482, opened 2025-06-28
Deprecate `unshare` and add `unshare_unsafe`.

### LINKED ISSUE #1479: `rustix::thread::unshare(UnshareFlags::FILES)` can violate `io_safety`
Pretty simple:
 - create a thread
 - unshare the file descriptor table using this API that's not marked `unsafe`
 - open a file, create a pipe, whatever you like, get an `OwnedFd`
 - `OwnedFd` is `Send` so you can move it to another thread, but the fd won't exist there
 - even worse, if another unrelated fd with the same number gets created in the receiving thread and then you drop the moved fd from the thread then it'll close the other fd, leaving a dangling reference, potentially in far-away and unrelated code
 - also nice, but not technically UB: by moving the `OwnedFd` out of the first thread, you'll have effectively leaked the original fd

`unshare` is a pretty powerful API.  I wouldn't be surprised if some of the other variants caused issues (although we are protected by the kernel requiring single-threaded processes for many of them).

Here's a reproducer:

```rust
use std::thread::spawn;

use rustix::{fs::{OFlags, Mode, open, fstat}, thread::{unshare, UnshareFlags}};

fn main() {
    let broken_fd = spawn(|| {
        unshare(UnshareFlags::FILES).unwrap();
        open("/", OFlags::PATH, Mode::empty()).unwrap()
    }).join().unwrap();

    // fstat() is always successful on a valid fd, right?
    fstat(broken_fd).unwrap();
}
```

```toml
[package]
name = "fd-unsafe"
version = "0.1.0"
edition = "2024"

[dependencies]
rustix = { version = "1.0.7", features = ["fs", "thread"] }
```

when you run it:

```
$ cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/fd-unsafe`
fatal runtime error: IO Safety violation: owned file descriptor already closed
Aborted (core dumped)
```

sunfishcode:
As pointed out in #1479, `unshare` can lead to threads observing dangling file descriptors. Deprecate it, and add a new `unshare_unsafe` that's unsafe, to reflect this.

Fixes #1479.

### DIFF
--- a/src/backend/libc/thread/syscalls.rs
+++ b/src/backend/libc/thread/syscalls.rs
@@ -343,8 +343,8 @@ pub(crate) fn setns(fd: BorrowedFd<'_>, nstype: c::c_int) -> io::Result<c::c_int
 
 #[cfg(linux_kernel)]
 #[inline]
-pub(crate) fn unshare(flags: crate::thread::UnshareFlags) -> io::Result<()> {
-    unsafe { ret(c::unshare(flags.bits() as i32)) }
+pub(crate) unsafe fn unshare(flags: crate::thread::UnshareFlags) -> io::Result<()> {
+    ret(c::unshare(flags.bits() as i32))
 }
 
 #[cfg(linux_kernel)]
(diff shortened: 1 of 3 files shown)


sunfishcode:
Yes, there are safe subsets of the function, but safety has to applies to the whole function. [This comment](https://github.com/bytecodealliance/rustix/pull/1437#issuecomment-3015995449) mentions the idea of adding new more specialized functions which can be safe, and that sounds like a good approach.

PR #1483, opened 2025-06-28
Remove the compiler-builtins optional dependency.

### LINKED ISSUE #1475: Drop the `compiler_builtins` dependency and feature
Since https://github.com/rust-lang/rust/pull/141993, crates that are deps of `std` don't need to specify `compiler_builtins`. This crate isn't currently a dependency of `std` so there is no urgency, I'm just documenting that it can be cleaned up :)

sunfishcode:
As advised in #1475, the compiler-builtins dependency is no longer needed. It was only used in rustc-dep-of-std mode.

Fixes #1475.


tgross35:
Thanks!

PR #1484, opened 2025-06-28
Don't let a mismatch AT_SYSINFO_EHDR image preclude use of AUX values.

### LINKED ISSUE #1465: Panic in page_size if platform of docker base image does not match
Hello `rustix` Team,

I create this issue as a result of hunting down a panic in wasmtime https://github.com/bytecodealliance/wasmtime/issues/10802. Now that I have a minimal reproducing example, I wonder if this even should be supported, yet I leave that judgement to you.

This is the panic we see:

```
thread 'main' panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustix-1.0.7/src/backend/linux_raw/param/auxv.rs:302:68:
called `Option::unwrap()` on a `None` value
```

As suggested by @alexcrichton the reproducing code is simple:

```rust
use rustix;

fn main() {
    let page_size = rustix::param::page_size();
    eprintln!("{page_size}");
}
```

The environment to reproduce it, uses this container file:

```containerfile
FROM --platform=linux/amd64 ubuntu:24.04

RUN apt-get update && \
    apt-get upgrade -y && \
    apt-get install -y pkg-config libssl-dev curl build-essential && \
    rm -rf /var/lib/apt/lists/*

RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs -o rustup-init && \
    sh rustup-init -y --profile minimal

ENV PATH="/root/.cargo/bin:${PATH}"

WORKDIR /usr/local/src

ENV CARGO_INCREMENTAL=0

WORKDIR /build

# Build application
COPY . .
RUN cargo build --release
CMD ["cargo", "run", "--release"]
```

On my ARM Mac Book Pro M3 executing:

```
podman build . --tag mre --platform linux/arm64
podman run mre
```

Reproduces the above error.

---

The actual production code hit by this, did not actually specify a platform in the containerfile though. It simply used a base image build on an x86_64 platform

Not sure if this is an issue with rustix, or Rossetta or if this is supposed to work at all. Yet I thought you folks may be interessted.

Best, Markus

sunfishcode:
If the ELF image pointed to by the `AT_SYSINFO_EHDR` AUX record doesn't match the architecture rustix is compiled for, don't use it, but do continue to use the rest of the AUX fields.

Fixes #1465.

### DIFF
--- a/src/backend/linux_raw/param/auxv.rs
+++ b/src/backend/linux_raw/param/auxv.rs
@@ -22,8 +22,7 @@ use core::sync::atomic::Ordering::Relaxed;
 use core::sync::atomic::{AtomicPtr, AtomicUsize};
 use linux_raw_sys::elf::*;
 use linux_raw_sys::general::{
-    AT_BASE, AT_CLKTCK, AT_EXECFN, AT_HWCAP, AT_HWCAP2, AT_MINSIGSTKSZ, AT_NULL, AT_PAGESZ,
-    AT_SYSINFO_EHDR,
+    AT_CLKTCK, AT_EXECFN, AT_HWCAP, AT_HWCAP2, AT_MINSIGSTKSZ, AT_NULL, AT_PAGESZ, AT_SYSINFO_EHDR,
 };
 #[cfg(feature = "runtime")]
 use linux_raw_sys::general::{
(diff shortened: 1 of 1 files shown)


sunfishcode:

PR #1486, opened 2025-06-28
Pin CI to Rust nightly-2025-06-23.

sunfishcode:
This works around the issue filed upstream in rust-lang/rust#143164.


sunfishcode:

PR #1493, opened 2025-08-05
Enable fstatvfs for Haiku

pulkomandy:
I'm not sure why this was disabled. It is available in Haiku since 2005.

Also skip a test that fails on Haiku.


sunfishcode:
Thanks!

PR #1497, opened 2025-08-12
build: Use temporary file for can_compile

keiichiw:
The `can_compile` function in build.rs previously invoked `rustc` with `-o -`, which causes `rustc` to attempt to create a temporary metadata directory in the current working directory.

When building this in another build system such as Portage, this happens in a sandbox where the current working directory is not writeable, resulting in a sandbox access violation.

This change modifies the `can_compile` function to use a temporary directory for the output file. This forces `rustc` to create the temporary file in a writeable location, avoiding the sandbox issue.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -258,7 +258,7 @@ fn can_compile<T: AsRef<str>>(test: T) -> bool {
         .arg("--target")
         .arg(target)
         .arg("-o")
-        .arg("-")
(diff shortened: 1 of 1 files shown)


sunfishcode:
This is now released in rustix 1.1.1.

keiichiw:
Thanks for the report.  I'll fix it and send a separate PR next week.

PR #1499, opened 2025-08-17
Make the flock tests more robust

asomers:
Instead of attempting to lock Cargo.toml, use a temporary file.  This fixes test failures that result from the source file system not supporting flock, like some NFS file systems.


sunfishcode:
This is now released in rustix 1.1.1.

PR #1505, opened 2025-09-01
Update CI to Rust nightly 2025-08-31.

sunfishcode:


sunfishcode:
rust-lang/rust#143164 will not be fixed, so we'll need to change the test before we can unpin nightly.

PR #1506, opened 2025-09-03
Enable `statvfs` on Redox

willnode:
`statvfs` on Redox is working.

I've checked this code compiles on both Linux and Redox, and [the test on Redox's relibc](https://gitlab.redox-os.org/redox-os/relibc/-/blob/master/tests/sys_statvfs/statvfs.c) is also working on Redox.


sunfishcode:
Thanks!

PR #1508, opened 2025-09-09
Update to linux-raw-sys 0.11.0.

sunfishcode:


ystreet:
This caused two versions of linux-raw-sys to become dependencies as one version was missed:

https://github.com/bytecodealliance/rustix/blob/main/Cargo.toml#L53

PR #1515, opened 2025-09-09
Enable statvfs for Haiku.

pulkomandy:
A previous commit enabled fstatvfs, but missed statvfs, which exists on Haiku as well.

Sorry for not noticing earlier.


sunfishcode:
Thanks!

PR #1517, opened 2025-09-10
Add support for name_to_handle_at(2), open_by_handle_at(2) on Linux

bertschingert:
This adds support for `{name_to,open_by}_handle_at(2)` on Linux.

I have a Rust application that wants to use these syscalls and it would be really nice to use a standard library rather than roll my own wrappers.

I would consider this an RFC at this point because I do have a few open questions about the API:

1. I don't believe it's valid for user applications to interpret or modify the `handle_bytes` and `handle_type` fields, even though they are exposed in the C struct definition. (The `handle_bytes` field does need to be accessed to size the handle correctly, but that is handled by this library, so user apps need not worry about that.) So this API makes `struct FileHandle` entirely opaque. If I'm wrong and there is a valid reason for a user app to access those fields, I think getter/setters could be added in the future without breaking the API...
2. It could be useful to have some kind of `BorrowedFileHandle` type that contains `&[u8]` instead of `Box<[u8]>` -- I don't immediately have a use case for it, so I didn't implement it, but putting it out there in case the idea does seem useful
3. `name_to_handle_at(2)` is interesting in that its signature is different depending on the flag `AT_HANDLE_MNT_ID_UNIQUE` -- the `mount_id` parameter is either an `int *`, or `u64 *`. I thought of two ways to handle this in rust: create a separate `name_to_handle_at_unique()` function for the `u64` variant, or create an Enum so that one function can return both kinds of mount_id. I went with the latter, but I could see an argument for the former being better. Thoughts?

I want to be able to store a filehandle and/or send it over the network, and then use it again when I get it back, so that is the reason for adding public `from_raw()` and `into_raw()` methods onto `struct FileHandle`.

Another question is how to test this -- since open_by_handle_at(2) requires CAP_DAC_READ_SEARCH privilege, programs that use it typically run as root. Is there a precedent for how to write unit/integration tests that require root in Rustix? (I have an external test program I ran with sudo to test this, I can share it if that would be helpful.)

### DIFF
--- a/src/backend/libc/fs/syscalls.rs
+++ b/src/backend/libc/fs/syscalls.rs
@@ -1929,6 +1929,29 @@ pub(crate) fn name_to_handle_at(
     }
 }
 
+#[cfg(all(target_os = "linux", feature = "alloc"))]
+pub(crate) fn open_by_handle_at(
+    mount_fd: BorrowedFd<'_>,
+    handle: *const core::ffi::c_void,
+    flags: OFlags,
+) -> io::Result<OwnedFd> {
+    syscall! {
+        fn open_by_handle_at(
+            mount_fd: c::c_int,
+            handle: *const ffi::c_void,
+            flags: u32
+        ) via SYS_open_by_handle_at -> c::c_int
+    }
+
+    unsafe {
+        ret_owned_fd(open_by_handle_at(
+            borrowed_fd(mount_fd),
+            handle,
+            flags.bits(),
+        ))
+    }
+}
+
 #[cfg(target_os = "linux")]
 pub(crate) fn sendfile(
     out_fd: BorrowedFd<'_>,
--- a/src/backend/libc/fs/syscalls.rs
+++ b/src/backend/libc/fs/syscalls.rs
@@ -31,6 +31,8 @@ use crate::fs::FallocateFlags;
     target_os = "wasi"
 )))]
 use crate::fs::FlockOperation;
+#[cfg(all(target_os = "linux", feature = "alloc"))]
+use crate::fs::HandleFlags;
 #[cfg(any(linux_kernel, target_os = "freebsd"))]
 use crate::fs::MemfdFlags;
 #[cfg(any(linux_kernel, apple, target_os = "redox"))]
@@ -1897,6 +1899,35 @@ pub(crate) fn openat2(
 const SYS_OPENAT2: i32 = 437;
 #[cfg(all(linux_kernel, target_pointer_width = "64"))]
 const SYS_OPENAT2: i64 = 437;
+
+#[cfg(all(target_os = "linux", feature = "alloc"))]
+pub(crate) fn name_to_handle_at(
+    dirfd: BorrowedFd<'_>,
+    path: &CStr,
+    handle: *mut ffi::c_void,
+    mount_id: *mut ffi::c_void,
+    flags: HandleFlags,
+) -> io::Result<()> {
+    syscall! {
+        fn name_to_handle_at(
+            dir_fd: c::c_int,
+            path: *const ffi::c_char,
+            handle: *mut ffi::c_void,
+            mount_id: *mut ffi::c_void,
+            flags: u32
+        ) via SYS_name_to_handle_at -> c::c_int
+    }
+
+    unsafe {
+        ret(name_to_handle_at(
+            borrowed_fd(dirfd),
+            c_str(path),
+            handle,
+            mount_id,
+            flags.bits(),
+        ))
+    }
+}
 
 #[cfg(target_os = "linux")]
 pub(crate) fn sendfile(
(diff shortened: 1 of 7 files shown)


src/fs/filehandle.rs
```diff
@@ -0,0 +1,194 @@
...
+/// An identifier for a mount that is returned by [`name_to_handle_at`].
+///
+/// [`name_to_handle_at`]: crate::fs::name_to_handle_at
+#[derive(Debug)]
+pub enum MountId {
+    /// By default a MountId is a C int.
+    Regular(ffi::c_int),
+    /// When `AT_HANDLE_MNT_ID_UNIQUE` is passed in `HandleFlags`, MountId is a u64.
+    Unique(u64),
+}
```
cyphar, line 101: I would honestly just suggest using a `u64` in the API for both and not bothering with this split, it's just more annoying for callers and there's no real benefit to getting less bits for classic mount IDs (IMHO).

This is also what `Statx::stx_mnt_id` does (`statx` even has the same flag-based behaviour -- in fact, I picked the `AT_HANDLE_MNT_ID_UNIQUE` to match the name of the `STATX_` flag). (Though `Statx` does that because it matches the kernel struct definition.)

src/fs/filehandle.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// excessively high when the filehandle will usually be much smaller than MAX_HANDLE_SIZE.
+    /// This function "trims" the filehandle so that the slice is only as large as it needs to be.
+    fn trim(&mut self) {
+        let len = self.get_handle_bytes() + HANDLE_STRUCT_SIZE;
+
+        self.raw = Box::from(&self.raw[0..len]);
+    }
+
+    /// Set the `handle_bytes` field (first 4 bytes of the struct) to the given length.
+    fn set_handle_bytes(&mut self, size: usize) {
```
cyphar, line 70: nit: This *is* the name of the field but this just looks confusing in Rust code -- you're getting and setting the length of the byte slice, so `set_handle_len` seems more appropriate in Rust.

src/fs/filehandle.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        assert!(handle.raw.len() >= handle.get_handle_bytes() + HANDLE_STRUCT_SIZE);
+
+        handle
+    }
+
+    /// Get the raw bytes of a file handle.
+    pub fn into_raw(self) -> Box<[u8]> {
+        self.raw
+    }
```
cyphar, line 57: It would be nice to have an `as_raw` too imho.

cyphar (changes requested):
Seems quite reasonable, my only strong comment is about `MountId`.

src/fs/filehandle.rs
```diff
@@ -0,0 +1,194 @@
+use core::mem::size_of;
+
+use crate::{backend, ffi, io, path};
+use backend::fd::{AsFd, OwnedFd};
+use backend::fs::types::{HandleFlags, OFlags};
+
+/// This maximum is more of a "guideline"; the man page for name_to_handle_at(2) indicates it could
+/// increase in the future.
+const MAX_HANDLE_SIZE: usize = 128;
```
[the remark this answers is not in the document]
cyphar: Indeed it does.

emilazy:
Re: reasons to look at `handle_bytes`, see https://blog.sebastianwick.net/posts/three-little-rust-crates/ for a use case that is necessary to interoperate with systemd, and these relevant kernel commits: https://github.com/torvalds/linux/commit/b95361481b1e5bd3627835b7e4b921d5a09e68a4, https://github.com/torvalds/linux/commit/e83f0b5d10dcf62833008327cb661c7d118bca85, https://github.com/torvalds/linux/commit/e83f0b5d10dcf62833008327cb661c7d118bca85, https://github.com/torvalds/linux/commit/04173501a69ed0387cceb190181ef8f7f1c54f0a.

Re: testing, see https://github.com/torvalds/linux/commit/620c266f394932e5decc4b34683a75dfc59dc2f4 – `CAP_DAC_READ_SEARCH` is not always required on Linux ≥ 6.11.

bertschingert:
Thanks for the comments / reviews so far.

> I would honestly just suggest using a u64 in the API for both

I think that is a nicer API too. This is what it does now.

> Re: reasons to look at handle_bytes, see https://blog.sebastianwick.net/posts/three-little-rust-crates/ for a use case

Added `Filehandle::get_handle_contents(&self) -> &[u8]` which should allow the user access to interpret the handle contents when necessary.

src/backend/libc/fs/syscalls.rs
```diff
@@ -31,6 +31,8 @@ use crate::fs::FallocateFlags;
     target_os = "wasi"
 )))]
 use crate::fs::FlockOperation;
+#[cfg(all(target_os = "linux", feature = "alloc"))]
```
bertschingert, line 34: If a `BorrowedFileHandle` type is later added, the `feature = "alloc"` check can be made specifically for just `struct FileHandle` and removed from all of these places...

PR #1522, opened 2025-09-21
Mark xer as clobbered in powerpc/powerpc64 syscalls

taiki-e:
powerpc/powerpc64 syscalls can clobber ctr and xer.

Since ctr cannot currently be marked as clobbered (will be fixed in https://github.com/rust-lang/rust/pull/146831), this PR only marks xer as clobbered for now.

Refs:
- musl: https://git.musl-libc.org/cgit/musl/commit/?id=f6944eb3c4ce1c97dc39dc36d32390dc9f70b67b
- glibc: https://github.com/bminor/glibc/blob/8aa99c592bf8f5894215fd62fef8248fac19dee1/sysdeps/unix/sysv/linux/powerpc/sysdep.h#L117


sunfishcode:
Thanks!

PR #1525, opened 2025-09-24
Make it possible to use Path and friends on wasip2 without the so named feature

Stebalien:
(draft because this is a bit of a hack and needs some discussion first)

At the moment, it's not possible to use `&Path`, `&OsStr`, etc. with rustix on WASIp2 without the nightly compiler (requires the `wasip2` feature) as `OsStrExt` and `OsStringExt` require the `wasip2` feature on WASIp2 so we can't use `as_bytes()` on `OsStr` and friends.

However, the recommendation on the [upstream issue](https://github.com/rust-lang/rust/issues/130323#issuecomment-2350430693) is to just use `to_str` as WASI paths are Unicode. This PR does exactly that.

@alexcrichton am I interpreting your comment correctly?

Alternatives:

- Wait. I'd prefer not to do this as I don't see a quick path to stabilization here.
- Work around this downstream. I'm submitting this to try to get [tempfile working](https://github.com/Stebalien/tempfile/pull/382) on the wasip2 platform on rust stable. I'm also happy to work around this downstream (in tempfile itself).
- Use [`OsStr::as_encoded_bytes`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.as_encoded_bytes) on wasi instead of `OsStrExt::as_bytes`. It will technically work but it also technically relies on an unstable representation (one that will likely never change on WASI, but still...).

---

This PR has two parts:

1. The first commit DRYs up the code a bit to avoid having to implement too many platform-specific conditionals. I tried to keep this change minimal but I'm also happy to go ahead and DRY up the rest of the code (there's a fair amount of duplicate code in this file).
2. The second commit special-cases conversions for `OsString` and `OsStr`, converting to utf8 strings where necessary.

### DIFF
--- a/src/path/arg.rs
+++ b/src/path/arg.rs
@@ -229,7 +229,6 @@ impl Arg for String {
 }
 
 #[cfg(feature = "std")]
-#[cfg(any(not(target_os = "wasi"), not(target_env = "p2"), wasip2))]
 impl Arg for &OsStr {
     #[inline]
     fn as_str(&self) -> io::Result<&str> {
@@ -251,6 +250,9 @@ impl Arg for &OsStr {
     where
         Self: 'b,
     {
+        #[cfg(all(target_os = "wasi", target_env = "p2", not(wasip2)))]
+        return self.to_str().ok_or(io::Errno::INVAL)?.into_c_str();
+        #[cfg(any(wasip2, not(all(target_os = "wasi", target_env = "p2"))))]
         return self.as_bytes().into_c_str();
     }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
I would agree that my thoughts there are being interpreted correctly. I can't comment much as to this PR specifically, however, as I am not familiar with this part of rustix (so would leave that part of review to someone else)

sunfishcode:
For WASI for all time, the question of whether there will ever be a WASI API that we'd want to be reachable through Rust's `std::fs` that would accept filesystem paths that aren't USV strings does not yet have a final answer.

But I think we can answer this question with sufficient confidence for WASIp2, specifically. WASIp2 filesystem paths are always USV strings. So for WASIp2, I think it's fine for Rustix to do the `to_str()` thing, as this PR does.

Issue #1519, opened 2025-09-13
More compilation errors on Android (version 1.1.2)


chenxiaolong:
I updated one of my Android projects from version rustix 1.0.8 to 1.1.2 and started encountering many compilation failures. They all seem to be similar to what was fixed in #1513/#1514 and appear to be introduced by fda7cbe6b8c6ebcfd18652ca1ca3e8177f9e0d6a.

```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `linux_raw_sys`
   --> /media/scratch/rustup/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustix-1.1.2/src/backend/libc/thread/syscalls.rs:670:9
    |
670 |     use linux_raw_sys::general::__kernel_clockid_t as clockid_t;
    |         ^^^^^^^^^^^^^ use of unresolved module or unlinked crate `linux_raw_sys`
    |
    = help: if you wanted to use a crate named `linux_raw_sys`, use `cargo add linux_raw_sys` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `linux_raw_sys`
 --> /media/scratch/rustup/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustix-1.1.2/src/thread/setns.rs:9:5
  |
9 | use linux_raw_sys::general::{
  |     ^^^^^^^^^^^^^ use of unresolved module or unlinked crate `linux_raw_sys`
  |
  = help: if you wanted to use a crate named `linux_raw_sys`, use `cargo add linux_raw_sys` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `linux_raw_sys`
  --> /media/scratch/rustup/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustix-1.1.2/src/backend/libc/thread/futex.rs:32:25
   |
32 |         const SIZE_U8 = linux_raw_sys::general::FUTEX2_SIZE_U8;
   |                         ^^^^^^^^^^^^^ use of unresolved module or unlinked crate `linux_raw_sys`
   |
   = help: if you wanted to use a crate named `linux_raw_sys`, use `cargo add linux_raw_sys` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `linux_raw_sys`
[...]

Issue #1095, opened 2024-07-24
Should Android use linux_raw by default?


sunfishcode:
https://github.com/bytecodealliance/rustix/pull/1090 proposes to switch the default backend on Android from libc to linux_raw. I myself don't have enough context to determine whether this is a good idea. I'm opening this issue in case anyone who sees this has any insight to add.

maurer:
As a data point, in the platform, we [use the libc backend explicitly](https://cs.android.com/android/platform/superproject/main/+/main:external/rust/crates/rustix/Android.bp;l=20). Changing the default won't cause us any trouble, but we chose to do it this way because:

1. It makes it possible to automatically use newer syscalls on newer platforms. For example, `open` will [use `openat`](https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/bionic/open.cpp;l=66;bpv=1), but on older systems, it would use `open` directly. This means that using `libc` allows you to get new kernel features as they're available, rather than being stuck with the lowest common denominator of system call options of all the devices you want to target.
2. There are various sanitizer and tracking tools that plug into bionic. Continuing with the `open` example, [`fdtrack`](https://android.googlesource.com/platform/bionic/+/HEAD/docs/fdtrack.md) is implemented inside `bionic` - by using the system calls directly, code using Rustix will be hidden from these. Opting out of them is fine, but I would expect most folks to want interaction to work by default.
3. That's the ABI we actually require to work. We don't *technically* require that OEMs ship Linux, that's just by far the easiest way to meet the requirements. `libc` on the other hand has to work, and has a defined stable ABI. This one is likely of less interest to you, since in practice, nearly every Android device runs Linux.

maurer:
> For 1: rustix is in some ways a lower-level interface than the platform libc or the standard library, and many users of rustix want to directly use specific syscalls and know which ones they're using. I think people already understand that by using rustix they may create dependencies on newer kernels for the functionality they use, and may have to write their own fallbacks if they want to support older kernels.

[...]

PR #1528, opened 2025-09-30
Fix Android build: enable `linux-raw-sys` and exclude Android-unsupported Linux userspace features

HemantKArya:
Enable access to Linux kernel types on Android targets while excluding a small set of Linux userspace features that are not available on Android (XDP, io_uring, certain `STATX` constants), restoring Android builds without changing Linux behavior.

---

## Motivation / Background

Recently the crate started failing to build for [Android targets](https://github.com/bytecodealliance/rustix/issues/1519) with errors like:

```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `linux_raw_sys`
error[E0659]: `STATX_ATTR_NODUMP` is ambiguous
```

Root cause: Android had been explicitly excluded from the `linux_raw_dep` / `linux-raw-sys` path in `build.rs` / `Cargo.toml`, which prevented Android builds from getting Linux kernel constant / syscall/type definitions that are required even when using the libc backend. Bionic (Android libc) does not expose all kernel-level constants and some syscalls/constants are only available via `linux-raw-sys`. At the same time, Android’s userspace is missing a few advanced Linux-only features and constants (e.g. `SOL_XDP`, `RWF_*`) — those should remain disabled on Android rather than enabled.

This PR restores Android builds by enabling access to `linux-raw-sys` on Android targets, and then selectively disables features that truly require Linux userspace support not present in Bionic.

---

## What this PR changes

**High level**

* Enable `linux-raw-sys` kernel types for Android targets so Android can access kernel constants/types it needs.
* Keep Linux functionality unchanged.
* Add targeted `#[cfg(...)]` guards so features not supported by Bionic (XDP, io_uring, certain STATX usages) remain disabled on Android.
* Update a small test to exclude Android-only-incompatible assertions.

## Rationale / Technical explanation

* Android uses the Linux kernel, so kernel-level types and syscall constants are *valid* on Android and MUST be available to the crate (via `linux-raw-sys`) so code paths that depend on kernel definitions compile.
* However, Android’s userspace (Bionic) does not expose certain Linux userspace constants and features:

  * **XDP / AF_XDP / SOL_XDP**: `SOL_XDP` and related constants are not present in Bionic. XDP is a Linux networking acceleration feature that is not applicable to Android userspace; those code paths are guarded so they are only enabled for desktop/server Linux.
  * **io_uring**: `RWF_*` (e.g., `RWF_NOWAIT`, `RWF_DSYNC`) flags and other io_uring-related userspace support may be absent in Bionic. io_uring code is therefore excluded on Android targets.
  * **STATX constants**: Some STATX constants can be defined both in libc and `linux-raw-sys`, causing ambiguity; adding Android to the exclusion list resolves the ambiguity and keeps the definitions consistent.

**Net effect**: Android gets what it needs from kernel types; Linux retains full feature set. No behavior changes for Linux targets.

---

## Testing performed

All CI workflow targets tested locally/in-workflow:

**Android** (all passing)

* `aarch64-linux-android` (default & `use-libc` features)
* `armv7-linux-androideabi` (`use-libc`)
* `x86_64-linux-android` (`use-libc`)
* `i686-linux-android` (`use-libc`)

**Linux** (all passing)

* Default build
* `all-apis` feature builds
* `use-libc,all-apis` builds
* `linux_latest,all-apis` builds
* All library tests pass (51/51)

(These are the results from my test runs; CI logs are attached in the workflow that exercised all Android and Linux variants.)

---

* Changes are intentionally minimal and scoped:

  * Only enables kernel-type dependency for Android.
  * Only disables a **small, well-defined** subset of features on Android that are known not to be available in Bionic (XDP, io_uring, RWF flags, certain STATX exposures).
* Linux behavior is unchanged — `#[cfg(all(linux_raw_dep, target_os = "linux"))]` keeps full Linux feature set intact.
* Tests and CI flows covering both Android and Linux have been exercised and pass.

---

## References (for reviewer verification)

* Android + Linux kernel architecture: [https://source.android.com/docs/core/architecture/kernel](https://source.android.com/docs/core/architecture/kernel)
* XDP (AF_XDP) docs: [https://www.kernel.org/doc/html/latest/networking/af_xdp.html](https://www.kernel.org/doc/html/latest/networking/af_xdp.html)
* io_uring paper / reference: [https://kernel.dk/io_uring.pdf](https://kernel.dk/io_uring.pdf)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,14 +25,14 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        build: [stable, nightly-2025-06-23, 1.63]
+        build: [stable, nightly, 1.63]
         include:
           - build: stable
             os: ubuntu-latest
             rust: stable
-          - build: nightly-2025-06-23
+          - build: nightly
             os: ubuntu-latest
-            rust: nightly-2025-06-23
+            rust: nightly
           - build: 1.63
             os: ubuntu-latest
             rust: 1.63
@@ -55,6 +55,7 @@ jobs:
         cargo update --package=regex --precise=1.9.0
         cargo update --package=half --precise=2.2.1
         cargo update --package=flate2 --precise=1.0.35
+        cargo update --package=syn --precise=2.0.106
         cargo update --package=textwrap --precise=0.16.1
         cargo update --package=once_cell --precise=1.20.3
         cargo update --package=parking_lot --precise=0.12.3
@@ -62,7 +63,8 @@ jobs:
         cargo update --package=lock_api --precise=0.4.12
         cargo update --package=rayon --precise=1.10.0
         cargo update --package=rayon-core --precise=1.12.1
-        cargo update --package=windows-sys@0.61.0 --precise=0.60.2
+        cargo update --package=windows-sys@0.61.2 --precise=0.60.2
+        cargo update --package=quote --precise=1.0.41
 
     - run: >
         rustup target add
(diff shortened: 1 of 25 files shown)


yujincheng08:
https://github.com/bytecodealliance/rustix/issues/1095 🤔

PR #1532, opened 2025-10-10
libc: use upstreamed SIGEMT support for mips and sparc

chenx97:
SIGEMT support for mips and sparc linux in libc has been merged upstream and is now available in 0.2.177.


sunfishcode:
Thanks!

PR #1534, opened 2025-10-25
Fix sockopt tests

pekkarr:
This makes the txtime sockopt tests to be run only when the `time` feature is enabled, otherwise `cargo test --features net` would fail with a compile error.

Also change the `socket_passcred` tests to use a Unix domain socket, because on Linux 6.16 and newer the `SO_PASSCRED` socket option is restricted to only Unix, netlink and Bluetooth sockets.

See the individual commit messages for more details.


sunfishcode:
Thanks!

PR #1536, opened 2025-10-30
Enable msghdr functions on Redox

jackpot51:
Recent work on unix sockets has made these functional.


jackpot51:
@sunfishcode can a new crates.io release be done with this change?

PR #1537, opened 2025-11-04
ci: Fix dependency versions for Rust 1.63 test

ids1024:
This seems to be working locally; hopefully CI passes...


sunfishcode:

PR #1541, opened 2025-11-09
Rudimentary `sanitize_memory` support.

### LINKED ISSUE #1531: msan finding in fstat implementation
There appears to be a problem with https://github.com/bytecodealliance/rustix/blob/main/src/backend/linux_raw/fs/syscalls.rs#L509 producing uninitialized memory.

I got the below WARNING when running a C++ test that exercises https://github.com/jj-vcs/jj under memory sanitizer.

e.g.
==8298==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x558e78ecd3f9 in <jj_lib::lock::unix::FileLock>::lock [third_party/rust/jj_lib/v0_34/src/lock/unix.rs:50]:24
Uninitialized value was stored to memory at
    #0 0x558e45994dea in __msan_memcpy [third_party/llvm/llvm-project/compiler-rt/lib/msan/msan_interceptors.cpp:1755]:3
    #1 0x558e7a60daa2 in rustix::backend::fs::syscalls::fstat [third_party/rust/rustix/v1/src/backend/linux_raw/fs/syscalls.rs:512]:9

The uninitialized usage occurs at:
https://github.com/jj-vcs/jj/blob/main/lib/src/lock/unix.rs#L50

It seems surprising that fstat wouldn't initialize the st_nlink field, perhaps there's some kind of msan annotation missing?

sunfishcode:
This adds a simple mechanism for informing msan of memory initializations that happen within `asm` blocks, and uses it in `fstat`, `read`, `pread`, and `getrandom`. More functions can be added in the future.

Fixes #1531.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -262,6 +262,7 @@ check-cfg = [
     'cfg(linux_raw)',
     'cfg(linux_raw_dep)',
     'cfg(lower_upper_exp_for_non_zero)',
+    'cfg(sanitize_memory)',
     'cfg(netbsdlike)',
     'cfg(rustc_attrs)',
     'cfg(rustc_diagnostics)',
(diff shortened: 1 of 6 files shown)


sunfishcode:

PR #1549, opened 2025-11-18
Implement dirfd equivalent functions

### LINKED ISSUE #1455: Consider adding a Dir::into_fd() method
Currently, once an `OwnedFd` has been passed to `rustix::fs::Dir`, there is no way to reuse it.

My usecase is to call `rustix::fs::openat()` and `rustix::fs::statat()` on it after I’m done with the iteration (or even during, I don’t think there is any issue with that since `getdents64(2)` is stateless).

My current solution is to call `rustix::io::dup()` on the fd before creating the `Dir`, but that shouldn’t be needed and wastes one fd.  Another solution would be to `rustix::fs::open()` it again once the iteration is done, but that’s wasteful too.

LGFae:
Implement functions that allow us to retrieve the directory stream file descriptor, equivalent to libc's `dirfd` function.

On linux, we can simply return the file descriptor itself (I verified this is what musl does).

---

This fixes #1455. I've tried following `rustix`'s style in general. Overall, it is a fairly simple change. I don't know if you want to add some tests for it or not (I may need a little bit of guidance in that case).

### DIFF
--- a/src/backend/libc/fs/dir.rs
+++ b/src/backend/libc/fs/dir.rs
@@ -75,6 +75,27 @@ impl Dir {
         }
     }
 
+    /// Returns the file descriptor associated with the directory stream.
+    ///
+    /// The file descriptor is used internally by the directory stream. As a result, it is useful
+    /// only for functions which do not depend or alter the file position.
+    ///
+    /// # References
+    ///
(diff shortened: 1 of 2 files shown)


sunfishcode:

PR #1555, opened 2025-12-16
Enable ioctl and makedev on Redox 

jackpot51:
This is needed for porting smithay


sunfishcode:
I took the liberty of modifying this to avoid adding the `as Dev` cast for non-Redox platforms. I added your name with a Co-authored-by in the commit message. Thanks for the patch!

jackpot51:
For those systems, I'll adjust the type to be the same as on 64-bit systems (and Android 32-bit)

PR #1557, opened 2025-12-21
clippy: fix if_then_some_else_none lint

xtqqczze:
https://rust-lang.github.io/rust-clippy/master/index.html#if_then_some_else_none


sunfishcode:
I have acquired a preference for avoiding `.then` in these kinds of functions in rustix. It makes the control flow a little harder to follow, and doesn't make the code that much easier to read.

PR #1558, opened 2025-12-21
Add AF_VSOCK based SocketAddr

notgull:
This is necessary in order to connect to AF_VSOCK sockets, or listen
on the AF_VSOCK socket type.

Dependent on https://github.com/sunfishcode/linux-raw-sys/pull/177


notgull:
No, I don't know why this `* 8` is necessary. "addr_len" keeps returning 16 when the real address length is 128. Maybe some kind of weird endianness issue.

PR #1569, opened 2026-01-03
hurd: Fix l_type and l_whence types

sthibaul:
on hurd x86, they are set to int types:

https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/i386/bits/types/struct_flock.h


sthibaul:
the build failure on st_atime/etc. showing up in selected Tier 3 musl, and in riscv64 are unrelated

sunfishcode:
This is now released in rustix 1.1.4.

PR #1570, opened 2026-01-07
Clobber vector registers and do not use preserves_flags in riscv64 syscalls

taiki-e:
From kernel docs: https://docs.kernel.org/arch/riscv/vector.html#vector-register-state-across-system-calls

> ## 3. Vector Register State Across System Calls
>
> As indicated by version 1.0 of the V extension [1], vector registers are clobbered by system calls.
>
> 1: https://github.com/riscv/riscv-v-spec/blob/master/calling-convention.adoc

And the linked docs says:

> Executing a system call causes all caller-saved vector registers (`v0-v31`, `vl`, `vtype`) and `vstart` to become unspecified.

See also this glibc bugfix: https://patchwork.sourceware.org/project/glibc/patch/18a15d9d-e3d0-4625-a49e-03e3ecddf011@tenstorrent.com/
(Probably musl ([riscv64](https://github.com/bminor/musl/blob/1b76ff0767d01df72f692806ee5adee13c67ef88/arch/riscv64/syscall_arch.h), [riscv32](https://github.com/bminor/musl/blob/1b76ff0767d01df72f692806ee5adee13c67ef88/arch/riscv32/syscall_arch.h)) also has the same bug. cc @Amanieu @kraj as [target maintainers of corresponding Rust targets](https://doc.rust-lang.org/nightly/rustc/platform-support/riscv64gc-unknown-linux-musl.html))

Unlike the above glibc patch which clobbers them depending on whether the vector ISA is enabled, this patch always clobbers them. (`cfg(target_feature = “vector”)` does not work except on nightly because the `vector` target feature is unstable. And, in Rust, always marking them as clobbered works fine (as [clobber_abi does](https://github.com/rust-lang/rust/blob/84c84421cc5cd8a416f58c77face28e79f6fac96/tests/codegen-llvm/asm/riscv-clobbers.rs)).


sunfishcode:
This is now released in rustix 1.1.4.

PR #1571, opened 2026-01-13
Cargo.toml: Add support for big-endian aarch64

neuschaefer:
I tested this change with "cargo test" using Rust 1.91 on an unofficial build of Alpine Linux for aarch64_be (aarch64_be-unknown-linux-musl). All tests pass, including doctests.

---

This PR could be considered a simplified version of https://github.com/bytecodealliance/rustix/pull/1472. Most notably, I did not include the `rustix_use_untested_linux_raw` flag, because I think it adds unnecessary friction.


folkertdev:
@neuschaefer could you re-run CI (maybe just rebase and force push) here? The logs have expired, locally `cargo +1.63.0-x86_64-unknown-linux-gnu check` works fine, so probably CI would just pass? 

Merging this would be useful for testing `aarch64_be` (in particular as part of `rust-lang/stdarch` development)

folkertdev:
Or @sunfishcode can maybe close/reopen the PR to the same effect?

PR #1575, opened 2026-01-29
Add `getcpu` system call support for getting a NUMA node of the current thread

Eugene-Usachev:
Not much time ago I needed to get a NUMA node of the current thread. Linux provides the `getcpu` system call for it. But I did not find any Rust library that provides this function. I read this repository code, and I like what you are doing, so I want to  help you to extend the functionality of this library. 

I implemented the `getcpu` syscall for Linux only. Also, I develop with WSL, which adds \r\n at the end of lines, so I patched tests for not reading the end of lines.

### DIFF
--- a/src/backend/linux_raw/vdso_wrappers.rs
+++ b/src/backend/linux_raw/vdso_wrappers.rs
@@ -127,7 +127,39 @@ pub(crate) fn clock_gettime_dynamic(id: DynamicClockId<'_>) -> io::Result<Timesp
     target_arch = "s390x",
 ))]
 #[inline]
+pub(crate) fn getcpu() -> (usize, usize) {
+    // SAFETY: `GETCPU` contains either null or the address of a function with
+    // an ABI like libc `getcpu`, and calling it has the side effect of writing
+    // to the result buffers, and no others.
+    unsafe {
+        let mut cpu = MaybeUninit::<u32>::uninit();
+        let mut numa_node = MaybeUninit::<u32>::uninit();
+        let callee = match transmute(GETCPU.load(Relaxed)) {
+            Some(callee) => callee,
+            None => init_getcpu(),
+        };
+        let r0 = callee(cpu.as_mut_ptr(), numa_node.as_mut_ptr(), null_mut());
+
+        debug_assert_eq!(r0, 0);
+
+        (cpu.assume_init() as usize, numa_node.assume_init() as usize)
+    }
+}
+
+#[cfg(feature = "thread")]
+#[cfg(any(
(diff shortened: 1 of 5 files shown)


src/backend/linux_raw/vdso_wrappers.rs
```diff
@@ -308,6 +340,7 @@ fn init_clock_gettime() -> ClockGettimeType {
     target_arch = "s390x",
 ))]
 #[cold]
+#[inline(never)]
```
sunfishcode: Is there a reason for `inline(never)` here? It's already marked `#[cold]`, which should have the desired effect. It a compiler decides it really wants to inline this, even given what we've told it, that seems fine.

src/thread/sched.rs
```diff
@@ -159,3 +159,25 @@ pub fn sched_getaffinity(pid: Option<Pid>) -> io::Result<CpuSet> {
...
+/// println!("The current thread was on the {core} core and {numa_node} numa node.");
+/// ```
+///
+/// # References
+///  - [Linux]
+///
+/// [Linux]: https://man7.org/linux/man-pages/man2/getcpu.2.html
+#[cfg(linux_kernel)]
+#[inline]
+pub fn getcpu() -> (usize, usize) {
```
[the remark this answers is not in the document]
Eugene-Usachev: I tried to do the same as `sched_getcpu`. To be honest, I am not sure what users prefer more here. I use `numa_node: usize` in my code, but I can't say everyone does it. If you want me to change it I can do it but I think it is not important.

src/backend/linux_raw/vdso_wrappers.rs
Eugene-Usachev: I am still sure that non-inlining provides a better performance, but I agree with both points: the difference is tiny and the compiler can do it as it wants. But I want to explicitly tell the compiler to generate exactly one call method for the calling this function that should be called only once. If you don't like this, I can't refuse this change. After all, the main goal of the PR is adding `getcpu`.

sunfishcode: I'm not so sure, myself. In a compiler with hot/cold basic block partitioning, inlining and then moving the cold blocks away from the hot path achieves a similar result to not-inlining, except that the compiler can more easily use an effectively custom calling convention. I don't know how much it matters in the code in question here, but in general, I don't like prohibiting compilers from doing things unless I have specific reasons.

Eugene-Usachev: I have removed this line.

PR #1591, opened 2026-03-05
Define flags `AT_STATX_*` for Linux musl 1.2.3

heiher:
libc flags: https://github.com/rust-lang/libc/blob/e879ee90b6cd8f79b352d4d4d1f8ca05f94f2f53/src/unix/linux_like/mod.rs#L1618-L1621


heiher:
Apparently, not all targets are using musl v1.2.3 by default right now. It looks like we need to define this explicitly for the specific targets that are actually using musl v1.2.3.

Issue #1602, opened 2026-03-31
ci: feature `asm_experimental_arch` is declared but not used


xtqqczze:
```log
error: feature `asm_experimental_arch` is declared but not used
   --> src/lib.rs:117:44
    |
117 | #![cfg_attr(asm_experimental_arch, feature(asm_experimental_arch))]
    |                                            ^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D unused-features` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unused_features)]`

error: could not compile `rustix` (lib test) due to 1 previous error
```

xtqqczze:
https://github.com/rust-lang/rust/issues/44232

sunfishcode:
According to https://github.com/rust-lang/rust/issues/93335, inline asm for powerpc and s390x is now stablized, so the feature is no longer needed.

sunfishcode:
This is now fixed in https://github.com/bytecodealliance/rustix/pull/1613.

PR #1594, opened 2026-03-14
Increase the MSRV from 1.63 to 1.65

xtqqczze:
Closes https://github.com/bytecodealliance/rustix/issues/1607 (`libc` bumped MSRV to 1.65)

* Replace some runtime panics with compile-time assertions
* Disable targets that now fail these assertions
* Consolidate duplicated FFI re-exports

PR #1606, opened 2026-04-09
fix: pin libc crate to <0.2.184

### LINKED ISSUE #1607: ci: build broken due to libc MSRV bump
```log
error: package `libc v0.2.184` cannot be built because it requires rustc 1.65 or newer, while the currently active rustc version is 1.63.0
```

xtqqczze:
Closes https://github.com/bytecodealliance/rustix/issues/1607


xtqqczze:
[CI / Test rustix_use_experimental_asm (s390x-linux) (pull_request)](https://github.com/bytecodealliance/rustix/actions/runs/24203318445/job/70652474770?pr=1606) failure tracked by https://github.com/bytecodealliance/rustix/issues/1602.

LingMan:
The pinning should only apply to CI and not to every user. See https://github.com/bytecodealliance/rustix/blob/9f588bc922f98d19fdb1f44cdb17f190d239b0eb/.github/workflows/main.yml#L51 where other dependencies are already pinned.

xtqqczze:
The better approach is to bump MSRV to 1.65: https://github.com/bytecodealliance/rustix/pull/1594

PR #1613, opened 2026-04-17
Avoid `feature(asm_experimental_arch)` on arches that don't need it.

sunfishcode:
Per rust-lang/rust#93335, inline asm is stablized on powerpc and s390x, so avoid using `feature(asm_experimental_arch)` on those architectures.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,7 +114,15 @@
     any(feature = "rustc-dep-of-std", core_intrinsics),
     feature(core_intrinsics)
 )]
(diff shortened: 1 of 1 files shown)


Gelbpunkt:
Is it intentional that this did not modify the Cargo.toml [here](https://github.com/bytecodealliance/rustix/blob/main/Cargo.toml#L35) to no longer require the cfg to use linux-raw on s390x, powerpc and powerpc64?

sunfishcode:

Issue #1620, opened 2026-05-04
error: attributes starting with `rustc` are reserved for use by the `rustc` compiler


xtqqczze:
CI is currently broken due to the following error:

```text
error: attributes starting with `rustc` are reserved for use by the `rustc` compiler
  --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustix-1.1.4/src/backend/linux_raw/io/errno.rs:49:25
   |
49 | #[cfg_attr(rustc_attrs, rustc_layout_scalar_valid_range_start(0xf001))]
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

https://github.com/bytecodealliance/rustix/actions/runs/25315134393/job/74210836411?pr=1594

kevinmehall:
It was removed and replaced with pattern types: https://github.com/rust-lang/rust/pull/155433

sunfishcode:
This is now fixed on `main`.

smoelius:
It looks like this is not yet fixed in the latest version on crates.io(?).

https://crates.io/crates/rustix/1.1.4/code/src/backend/linux_raw/io/errno.rs#L49

Would it be possible to get a release with the fix? 🙏

smoelius:
I see that version 1.1.5 is now published on crates.io and includes the fix:

https://crates.io/crates/rustix/1.1.5/code/src/backend/linux_raw/io/errno.rs#L49

Thank you! 🙏

PR #1625, opened 2026-05-16
Expose mknodat/mkfifoat on Apple targets

mattsu2020:
## Summary

- Expose `rustix::fs::{mknodat, mkfifoat}` on Apple targets, with support starting in macOS 13.0, iOS 16.0, tvOS 16.0, and watchOS 9.0.
- Weak-link `mknodat` on Apple and return `NOSYS` when unavailable on older systems. `mkfifoat` delegates to the same implementation. Other targets retain direct libc calls.
- Document the supported Apple OS versions and behavior on older systems for both public APIs.

The documentation build succeeded on macOS with 10 unresolved-link warnings in unchanged documentation; none were in the updated API documentation.


xtqqczze:
I think a fallback is needed for macOS ≤ 13 since [Apple macOS targets](https://doc.rust-lang.org/nightly/rustc/platform-support/apple-darwin.html) are supported as far back as macOS 10.12, and there are also additional Apple platforms to consider such as tvOS and watchOS.

It would also be good to add references for the newly exposed functions on Apple targets. For example, https://github.com/apple-oss-distributions/xnu/blob/main/bsd/man/man2/mkfifoat.2

oech3:
Or add non-*at too for macOS?

xtqqczze:
Right, the POSIX `mknod` and `mkfifo` are currently absent from all platforms.

mattsu2020:
I'll revise this PR to restore weak-linking/runtime detection. When the symbol is unavailable, I'll only fall back to `mknod`/`mkfifo` for the `AT_FDCWD` case, and otherwise return an unsupported/error result. I'll also add Apple man page references for these functions.

xtqqczze:
On second thoughts, we probably don't need to implement a fallback for the libc backend. We can just return the appropriate Err and let callers handle it.

src/backend/libc/fs/syscalls.rs
```diff
@@ -1210,13 +1208,30 @@ pub(crate) fn mknodat(
...
+                *const ffi::c_char,
+                c::mode_t,
+                c::dev_t
+            ) -> c::c_int
+        }
+        // If we have `mknodat`, use it.
+        if let Some(libc_mknodat) = mknodat.get() {
+            return ret(libc_mknodat(borrowed_fd(dirfd), c_str(path), mode, dev));
+        }
+        Err(io::Errno::NOSYS)
```
xtqqczze:
```suggestion
        let libc_mknodat = mknodat.get().ok_or(io::Errno::NOSYS)?;
        ret(libc_mknodat(borrowed_fd(dirfd), c_str(path), mode, dev))
```

src/backend/libc/fs/syscalls.rs
```diff
@@ -1210,13 +1208,27 @@ pub(crate) fn mknodat(
     mode: Mode,
     dev: Dev,
 ) -> io::Result<()> {
+    let mode = (mode.bits() | file_type.as_raw_mode()) as c::mode_t;
+    let dev = dev.try_into().map_err(|_e| io::Errno::PERM)?;
+
+    // Older Apple platforms lack `mknodat`.
```
xtqqczze:
```suggestion
    // macOS < 13.0 lacks `mknodat`.
```

xtqqczze (approved):

xtqqczze:
[CI / Test (ubuntu) (pull_request)](https://github.com/bytecodealliance/rustix/actions/runs/28479033741/job/84410789485?pr=1625) failures tracked by https://github.com/bytecodealliance/rustix/issues/1620.

xtqqczze (approved):

oech3:
@sunfishcode Would you review this?

src/backend/libc/fs/syscalls.rs
```diff
@@ -1210,13 +1208,27 @@ pub(crate) fn mknodat(
     mode: Mode,
     dev: Dev,
 ) -> io::Result<()> {
+    let mode = (mode.bits() | file_type.as_raw_mode()) as c::mode_t;
```
xtqqczze, line 1213:
```suggestion
    let dirfd = borrowed_fd(dirfd);
    let pathname = c_str(path);
    let mode = (mode.bits() | file_type.as_raw_mode()) as c::mode_t;
```

We could remove the duplication and take these safe functions out of the unsafe block.

badeend:
@mattsu2020 The fix for the CI failure #1620 has been merged. Could you update this PR?

mattsu2020:
> @mattsu2020 The fix for the CI failure #1620 has been merged. Could you update this PR?

I’ve updated it.

badeend:
@sunfishcode Could you take a look at this? The code change seems relatively straightforward, and it passes CI.

oech3:
Resolve conflicts

mattsu2020:
> Resolve conflicts

fix it

xtqqczze:
@mattsu2020 please squash your commits and write a new commit description in your own words

mattsu2020:
> @mattsu2020 please squash your commits and write a new commit description in your own words

I performed a squash merge and added a comment. 
Do you think we can merge it?

xtqqczze (changes requested):
Can we clarify the platform support for `mknodat` and `mkfifoat` in `src/fs/at.rs`?

The macro-expanded declaration in `sys/stat.h` shows that it was introduced in macOS 13.0, iOS 16.0, tvOS 16.0, and watchOS 9.0.

<details>
<summary>Macro-expanded declaration</summary>

```c
int mkfifoat(int, const char *, mode_t)
__attribute__((availability(macos,introduced=13.0)))
__attribute__((availability(ios,introduced=16.0)))
__attribute__((availability(tvos,introduced=16.0)))
__attribute__((availability(watchos,introduced=9.0)));
```

</details>

Also, please amend the commit message to summarize the changes rather than describing every implementation detail. For example:

> Expose mknodat and mkfifoat on Apple targets.
>
> Weak-link `mknodat` on Apple and return `NOSYS` when unavailable on older systems. Keep direct libc calls on other targets.

This will make the PR easier to review when the maintainer gets to it.

mattsu2020:
Thank you for your feedback. I have incorporated your comments accordingly.

xtqqczze:
@mattsu2020 Please don't include PR number in commit.

Also, please update the PR description to remove unnecessary details, was that generated by an LLM?

I opened an issue for the rustdoc warnings: https://github.com/bytecodealliance/rustix/issues/1685

mattsu2020:
I added it because I thought it was necessary, but I'll remove it.

mattsu2020:
Should the rustdoc warnings be addressed in this PR?

I'm also fine with handling them in a separate PR.

xtqqczze:
The rustdoc warnings appear unrelated to this PR, so I don't think we should address them here.

Also, could you make the description a bit less verbose? For example, the "Validation" section can be removed. It reads like LLM-generated content, and mentioning that `git diff --check` passed doesn't add much value when we already run `rustfmt` in CI.

mattsu2020:
I revised the text.

PR #1628, opened 2026-05-29
build: Remove OUT_DIR probe artifact

zertosh:
Remove the `can_compile()` probe artifact from `OUT_DIR` at the end of the build script, so `OUT_DIR` ends up empty after the script runs.

Where the probe output goes has gone back and forth:

- #1200 (2024): moved the output out of `OUT_DIR` (`-o -`) because the emitted `.rmeta` is not byte-reproducible and destabilizes cache keys in build systems that capture `OUT_DIR` as a build output (e.g. Bazel).
- #1497 (2025): `-o -` made rustc create a temporary metadata directory in the current working directory, which broke sandboxed builds (ChromeOS Portage), so the output moved to `std::env::temp_dir()`.
- #1526 (2025): writing to the global temp dir left stray files on the host, so the output moved back into `OUT_DIR`, which reintroduced the nondeterminism that #1200 fixed.

Writing the probe to `OUT_DIR` and deleting it before the script exits satisfies all three constraints: no nondeterministic content left in `OUT_DIR`, no writes to the working directory, and no clutter in the system temp dir.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -195,6 +195,11 @@ fn main() {
     println!("cargo:rerun-if-env-changed=CARGO_FEATURE_USE_LIBC");
     println!("cargo:rerun-if-env-changed=CARGO_FEATURE_RUSTC_DEP_OF_STD");
     println!("cargo:rerun-if-env-changed=CARGO_CFG_MIRI");
+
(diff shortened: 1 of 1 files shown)


sunfishcode:
Thanks!

sunfishcode:

PR #1633, opened 2026-06-15
Mangle the name of `rustix::runtime` to discourage users.

sunfishcode:
There have been a few bug reports from users appearing to use `rustix::runtime` as if it were a normal library feature. It is not. Since the comments in the file don't seem to be sufficient for discouraging users, add a random string to the module name.

This would be a semver break, except that `rustix::runtime` has always been `doc(hidden)` and commented as experimental.


sunfishcode:

PR #1640, opened 2026-07-01
Ensure reproducible builds when building with Bazel

mgeisler:
The build script would write a rustc `--emit=metadata` file into `OUT_DIR` purely to test whether a snippet compiles. Since the metadata is not stable, this file could vary from run to run.

When building with Bazel, which hashes the content of the files in `OUT_DIR`, the leftover file makes every build have a different cache key. This in turn defeats remote caching for the subtree of build graph below `rustic`.

This PR simply deletes the throwaway probe artifact before the build script exits so `OUT_DIR` is reproducible.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -251,7 +251,7 @@ fn can_compile<T: AsRef<str>>(test: T) -> bool {
         .arg("--target")
         .arg(target)
         .arg("-o")
-        .arg(out_file)
+        .arg(&out_file)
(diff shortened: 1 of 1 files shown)


mgeisler:
Hi there! I looked briefly at the build failures and I think they are unrelated?

xtqqczze:
@mgeisler please rebase for CI

PR #1644, opened 2026-07-22
Fix ioctl docs link and fd module wording

futurehua:
Fix ioctl docs link and fd module wording


xtqqczze:
@futurehua please rebase for CI

PR #1647, opened 2026-08-03
ci: fix syn pin

xtqqczze:
Fix the CI failure caused by an ambiguous `syn` package selection after multiple versions of `syn` are present in the dependency graph.

The failing command produces:

```log
error: There are multiple `syn` packages in your project, and the specification `syn` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  syn@2.0.119
  syn@3.0.3
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,6 +53,7 @@ jobs:
       run: |
         cargo update --package=half --precise=2.2.1
         cargo update --package=flate2 --precise=1.0.35
+        cargo update --package=serde --precise=1.0.228
         cargo update --package=syn --precise=2.0.106
         cargo update --package=textwrap --precise=0.16.1
         cargo update --package=parking_lot --precise=0.12.4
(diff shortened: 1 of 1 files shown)


a-shannon:
The minimal fix is to pin the Serde family before pinning Syn in both Rust 1.65 update blocks:

```sh
cargo update --package=serde --precise=1.0.228
cargo update --package=syn --precise=2.0.106
```

Pinning `serde_derive` alone fails because `serde 1.0.229` requires `serde_core =1.0.229`, whose lockstep constraint requires `serde_derive =1.0.229`. Pinning `serde` to 1.0.228 moves the family together and removes the Syn 3 node, making the bare Syn selector unambiguous.

Local Rust 1.65 validation passed: 34/34 commands in the `Check (1.65)` sequence, plus the native test run with 412 passed, 0 failed, and 5 ignored.

This is local evidence only; hosted CI and the cross-target test lanes are not claimed green.

sunfishcode:

PR #1467, opened 2025-05-23
Update statx to Linux 6.14

ginnyTheCat:
Fix #1385. Waiting on https://github.com/sunfishcode/linux-raw-sys/pull/155 + release and libc 0.2.190 for all constants.

PR #1656, opened 2026-08-09
Restore statx request-mask filtering

a-shannon:
## Summary

PR #379 added a forward-compatibility guard that masks unknown `statx`
request bits before invoking the kernel. After #787 enabled bitflags'
externally-defined-flags support, `StatxFlags::all()` became `u32::MAX`, so
`mask & StatxFlags::all()` became an identity operation and the guard stopped
filtering unknown bits.

This change:

- introduces a private `StatxRequestFlags` containing the 14 currently named
  elementary request bits;
- sanitizes requests before either backend or the compatibility path;
- continues rejecting `STATX__RESERVED` with `EINVAL`;
- keeps the public `StatxFlags` external-flags behavior unchanged; and
- adds a causal test proving that the sanitized set equals the named flags,
  not `StatxFlags::all()`.

This restores the future-kernel safety boundary intended by #379. It does not
claim a vulnerability on current kernels.

## Testing

Focused local validation recorded:

- linux_raw: causal unit 1/1 and `statx` integration 2/2;
- use-libc: causal unit 1/1 and `statx` integration 2/2;
- focused library and `fs` test-target Clippy with `-D warnings` for both
  backends;
- `git diff --check`; and
- independent source review of the exact diff with no findings.

This does not claim a repository-wide Clippy, MSRV, or CI pass.

## Relation to #1467

This is complementary to #1467, not a replacement for it. #1467 adds
`MNT_ID_UNIQUE`, `SUBVOL`, `WRITE_ATOMIC`, and the write-atomic attribute.
Whichever PR lands second should extend the private request allowlist with the
newly named request bits during rebase.


xtqqczze:
@a-shannon merge conflicts needs to be resolved

a-shannon:
Resolved the conflicts in [#1656](https://github.com/bytecodealliance/rustix/pull/1656) and rebased onto current `main` (`287214b8`), now at `f52c3147`. The resolution preserves upstream's `horizon` exclusion for `Timestamps` and this PR's separate 32-bit/MIPS Linux guard for the `StatxFlags` import. The functional patch is unchanged. Formatting and the FS-feature compile check passed locally; Linux validation is left to CI.

PR #1657, opened 2026-08-10
Add read-only FS_IOC_FSGETXATTR support

a-shannon:
## Summary

Add a read-only `ioctl_fsgetxattr` wrapper using the generated Linux
`struct fsxattr` definition.

This change:

- introduces a non-exhaustive `Fsxattr` result and named `FsxattrFlags`;
- returns all five non-padding fields from `FS_IOC_FSGETXATTR`;
- preserves unnamed returned `FS_XFLAG_*` bits through normal flag operations;
- zero-initializes the complete kernel object, including reserved padding;
- derives the ioctl opcode from the generated ABI type;
- adds no setter; and
- adds the missing `CASEFOLD`, `ENCRYPTED`, and `VERITY` names to `IFlags`.

## Testing

Focused local validation recorded for both the `linux_raw` and `use-libc`
configurations:

- four integration tests covering flag handling, ABI layout, wrapper-versus-raw
  result parity, and unsupported-descriptor errno parity;
- the field-conversion and unknown-bit operator unit test;
- the generated-constant opcode check; and
- focused Clippy with warnings denied.

The broader local Clippy run remained blocked only by unchanged diagnostics in
`timespec.rs`. This does not claim repository-wide Clippy, MSRV, CI, or
cross-architecture runtime validation.

## Related work

This is distinct from #1624, which adds `FICLONERANGE`. Both changes append
tests to `tests/fs/ioctl.rs`, so whichever lands second will need a small
test-file rebase.

### DIFF
--- a/src/fs/ioctl.rs
+++ b/src/fs/ioctl.rs
@@ -96,18 +96,27 @@ unsafe impl ioctl::Ioctl for Ficlone<'_> {
 
 #[cfg(linux_raw_dep)]
 bitflags! {
-    /// `FS_*` constants for use with [`ioctl_getflags`].
+    /// `FS_*` constants for use with [`ioctl_getflags`] and [`ioctl_setflags`].
     ///
     /// [`ioctl_getflags`]: crate::fs::ioctl::ioctl_getflags
+    /// [`ioctl_setflags`]: crate::fs::ioctl::ioctl_setflags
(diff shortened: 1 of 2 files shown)


xtqqczze:
@a-shannon needs a rebase for CI

PR #1630, opened 2026-06-03
Clean fsopen a bit & add support for `mount_setattr`

livingsilver94:
This is my attempt at supporting `mount_setattr`, taking into consideration the fact `struct mount_attr` is designed to be expandable, and that it currently borrows a user namespace's file descriptor. See discussions in  PR #1002 for more details.

The idea is to have a `MountAttr` struct that is a thin wrapper around `struct mount_attr`, that also manages lifetimes. The 
underlying type is not exposed: you can only set parameters via the constructor and setters.

While at it, I cleaned fsopen (and some related modules) a little. See the commit history.

Fixes #975.

## Test plan

```toml
[package]
name = "test"
version = "1.0.0"
rust-version = "1.63"
edition = "2021"

[dependencies]
rustix = { path = "/your/local/fork/of/rustix", features = [
    "mount",
    "linux_latest", # Or "use-libc" to test the other backend.
    "fs",
] }

```

```rust
use std::os::fd::AsFd;

use rustix::fs::{open, Mode, OFlags};
use rustix::io;
use rustix::mount::{
    mount_setattr, MountAttr, MountAttrFlags, MountPropagationFlags, MountSetAttrFlags,
};

// To create a test mount point, run:
// $ mkdir /tmp/testpoint
// # mount -t tmpfs tmpfs /tmp/testpoint

fn main() -> io::Result<()> {
    let dirfd = open("/tmp", OFlags::RDONLY | OFlags::DIRECTORY, Mode::empty())?;

    // Uncomment to test set_param_flag.
    // mount_setattr will return EINVAL, because this won't be a user namespace fd.
    // let fd = open("/dev/null", OFlags::RDONLY, Mode::empty())?;

    let attr = MountAttr::new(
        MountAttrFlags::MOUNT_ATTR_RDONLY,
        MountAttrFlags::empty(),
        MountPropagationFlags::empty(),
    );

    // Uncomment to test set_param_flag.
    // let attr = attr.set_param_flag(rustix::mount::MountAttrParamFlags::IdMap(fd.as_fd()));

    // Uncommenting this line gives a compilation error when set_param_flag is called.
    // drop(fd);

    mount_setattr(dirfd, "testpoint", MountSetAttrFlags::AT_RECURSIVE, attr)?;

    println!("/tmp/testpoint is now mounted read-only");
    Ok(())
}

```

Issue #1445, opened 2025-04-11
`listmount` and `statmount`


rusty-snake:
Issue for addition of `listmount` and `statmount`.

- https://www.man7.org/linux/man-pages/man2/listmount.2.html
- https://www.man7.org/linux/man-pages/man2/statmount.2.html

Note that these syscalls see heavy development and the manpages are likely to be already out of date.

cc @yujincheng08

PR #1659, opened 2026-08-10
Add listmount support

a-shannon:
## Summary

Add Linux `listmount` support to both the `linux_raw` and `use-libc`
backends.

This change:

- exposes `LISTMOUNT_ROOT` and `ListMountFlags::REVERSE`;
- uses the generated `mnt_id_req` v0 ABI for the caller's current mount
  namespace;
- accepts initialized or uninitialized caller-provided `Buffer<u64>` storage;
- returns only the mount-ID prefix reported initialized by the kernel;
- supports explicit cursor-based pagination;
- checks the returned count against the supplied capacity;
- marks only the initialized prefix for MemorySanitizer in the raw backend; and
- performs one syscall without allocating or retrying.

The API documents that separate pages do not form an atomic mount-tree
snapshot.

## Testing

Focused local validation on Rust/Cargo 1.89 and Linux 6.18 passed all four
integration tests with each backend, plus strict Clippy checks for both test
configurations.

The tests use an independent raw-syscall oracle. They cover exact-fit and
one-element-slack uninitialized buffers, cursor progression, invalid roots and
flags, and verify that reverse enumeration is the exact inverse of a stable
forward enumeration.

## Related work

This addresses the `listmount` portion of #1445. It does not add `statmount`.
PRs #1002 and #1630 implement `mount_setattr`; they are separate changes that
touch some of the same mount files.


xtqqczze:
@a-shannon please rebase now that CI is clean

a-shannon:
Rebased [#1659](https://github.com/bytecodealliance/rustix/pull/1659) onto current `main` (`287214b8`), now at `299ca7a0`. The rebase was clean and `range-diff` confirms the patch is unchanged. Formatting passed locally; Linux listmount validation is left to the new CI run.

PR #1663, opened 2026-08-18
Completely remove `rustc_attrs`.

sunfishcode:
rustix's use of `rustc_*` attributes was already behind a `--cfg` flag, but that flag is enabled in solme CI rules, and it appears rustc is no longer accepting user code using those attributes, so just comment them out entirely.


sunfishcode:

PR #1665, opened 2026-08-25
docs(linux_raw): fix vDSO linkage comments

daixiheguu:
Remove duplicated words from the LTO and non-LTO comments explaining weak vDSO symbol linkage in the Linux raw backend.


sunfishcode:
Please don't do this.

PR #1671, opened 2026-08-30
Fix build errors for the Horizon target

xtqqczze:
Horizon does not provide several libc APIs and types that are assumed by the generic Unix implementation. The affected code was being compiled unconditionally, resulting in missing symbols, incompatible types, and imports that are unavailable on Horizon.

- Account for Horizon's `poll` flag type, which since https://github.com/rust-lang/libc/pull/5045 uses `c_int` rather than `c_short`.

### DIFF
--- a/src/backend/libc/conv.rs
+++ b/src/backend/libc/conv.rs
@@ -3,7 +3,10 @@
 //! for converting between rustix's types and libc types.
 
 use super::c;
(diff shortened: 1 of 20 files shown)


sunfishcode:
Thanks!

sunfishcode:

PR #1673, opened 2026-09-05
avoid unnecessary type annotations for `.cast()`

xtqqczze:
Remove unnecessary type annotations from `.cast()` calls and let type inference do its thing.


xtqqczze:
Depends on https://github.com/bytecodealliance/rustix/pull/1674.

PR #1677, opened 2026-09-07
Add Padding type for uninitialized padding representation

xtqqczze:


src/fs/statx.rs
```diff
@@ -46,8 +47,8 @@ pub struct Statx {
     pub stx_atomic_write_segments_max: u32,
     pub stx_dio_read_offset_align: u32,
     pub stx_atomic_write_unit_max_opt: u32,
-    pub __spare2: [u32; 1usize],
-    pub __spare3: [u64; 8usize],
+    pub(crate) __spare2: Padding<[u32; 1usize]>,
```
xtqqczze, line 50: I guess making this field non-public could be considered a breaking change, even though it was probably an error to expose padding fields.

PR #1655, opened 2026-08-07
fix(bug): ns->us overflow handling

HalFrgrd:
I ran into an odd bug on macos. If my code polls for just under 5s (say 4.999_999_123 seconds), I immediately receive an `EINVAL` error.

This is due to the rounding logic. With the current logic, we set `tv_usec` to `1_000_000` when the nanosecond remainder is >= 999_999_001. This creates an invalid `timeval` on macOS and the OS returns `EINVAL`.

`tv_usec=1_000_000` is considered invalid and instead we should increase `tv_sec` by 1 and set `tv_usec=0`. 

# Solution

Checking if `tv_usec >= 1_000_000` https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/kern/kern_time.c#L670

XNU handling this issue in the same way that I propose: https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/time.h#L175-L177

Linux kernel is more forgiving if you set usec to 1_000_000: https://github.com/torvalds/linux/blob/master/fs/select.c#L723-L725. So I could revert the changes here for linux code and only keep the macos fix. Please let me know what is best.

# Reproducing

This is a small example that I have tested on macOS. `select` with a timeout of `999_999_123` fails using rustix master but is fixed when using my branch.

```rust
use rustix::event::{fd_set_insert, fd_set_num_elements, select, FdSetElement, Timespec};
use rustix::fd::AsRawFd;
use rustix::pipe::pipe;

fn main() {
    // Create a pipe so we have a valid file descriptor to pass to select()
    let (reader, _writer) = pipe().unwrap();
    let raw_fd = reader.as_raw_fd();
    let nfds = raw_fd + 1;
    let num_elems = fd_set_num_elements(1, nfds);

    // 1. 999_999_000 ns -> (999_999_000 + 999)/1000 = 999_999 μs (< 1,000,000 μs)
    let mut readfds1 = vec![FdSetElement::default(); num_elems];
    fd_set_insert(&mut readfds1, raw_fd);
    let res1 = unsafe {
        select(
            nfds,
            Some(&mut readfds1),
            None,
            None,
            Some(&Timespec {
                tv_sec: 0,
                tv_nsec: 999_999_000,
            }),
        )
    };
    println!("999_999_000 ns result: {:?}", res1);

[...]

PR #1681, opened 2026-09-17
Compile timespec for `net` feature

HalFrgrd:
Fixes the problem found here https://github.com/bytecodealliance/rustix/pull/1655#issuecomment-5717491906


ogoffart:
Thanks for the PR

@sunfishcode would be nice to have a release, because rustix 1.1.5 is failling to compile on our android CI

darkhonor:
This is the right fix, and it predates the duplicate I opened in #1690, which I have closed in favour of this one.

Offering one thing from it in case it is useful here rather than as a separate PR. The `check_no_default_features` job already checks `time` and `time,use-libc` individually, and the same pair for `net` would have caught this:

```yaml
    - run: cargo check --workspace --release --no-default-features --features net -vv
    - run: cargo check --workspace --release --no-default-features --features net,use-libc -vv
```

Verified against v1.1.5 on rustc 1.98.1 with the job's own `RUSTFLAGS: -D warnings -D elided-lifetimes-in-paths`: `net` fails with six `E0433`s through the linux_raw backend and `net,use-libc` with three through libc, and both pass with the one-line change in this PR. Entirely your call whether it belongs here or not at all.

Context on the impact, since a release is already being asked for above: this also breaks consumers that set `default-features = false` to opt into a minimal API surface. Full repro and verification matrix are in #1689, which I am closing as a duplicate of this.

